PR #4, opened 2021-04-06
Do not use self-chosen node ids for identifying peers.

abizjak:
## Purpose

This prevents a vulnerability where a malicious node could force another node to
drop an arbitrary connection by choosing the same node id.

The changes make the following logic for peer handling
- upon connection a peer P is assigned an id I by the node N (this is done by a 64-bit counter)
- during the handshake the peer P transfers to the node N their own "self" identifier. This is recorded by N but it is not used internally for anything other than reporting its peers. I left this in so that no other components of the system are affected, e.g., network dashboard, client.
- soft bans can be done either by IP address, socket address, or the identifier I.
- persistent bans can only be done by IP address
- we detect duplicate peers by either the socket address they are connected to us, the IP + advertised port. If an additional flag `--disallow-multiple-peers-on-ip` is set then we also disallow multiple peers on the same IP address.

## Changes

1. Add a `local_id` to each of the remote peers which is used to identify
them. This local id is simply the connection token we already have, which is
guaranteed to be unique.
2. Remove the capnproto dependency and serialization. This was woefully out of date and unusable, it only incurred a maintenance burden.
3. Add a flag `--disallow-multiple-peers-on-ip` to disallow multiple peers on the same IP address. This is disabled by default.
4. Add a flag `--no-clear-bans` to not clear persisted bans on startup. This is disabled by default (i.e., IP bans will be cleared on startup).

## Checklist

      hard-to-understand areas.


concordium-node/src/connection/message_handlers.rs
```diff
@@ -27,7 +29,7 @@ impl Connection {
             NetworkPayload::NetworkRequest(NetworkRequest::Handshake(handshake), ..) => {
                 return self.handle_handshake_req(handshake, conn_stats);
             }
-            _ => self.remote_id().ok_or_else(|| format_err!("handshake not concluded yet"))?,
+            _ => self.remote_peer.local_id,
```
abizjak, line 32: This is incorrect, we need to check that the peer is after handshake.

concordium-node/src/connection/message_handlers.rs
```diff
@@ -74,9 +76,9 @@ impl Connection {
     ) -> Fallible<()> {
         debug!("Got a Handshake request from peer {}", handshake.remote_id);
 
-        if self.handler.is_banned(BanId::NodeId(handshake.remote_id))? {
-            bail!("Rejecting handshake: banned node.");
-        }
+        // if self.handler.is_banned(BanId::NodeId(handshake.remote_id))? {
+        //     bail!("Rejecting handshake: banned node.");
+        // }
```
abizjak, line 81: This needs to be uncommented.

concordium-node/src/common/p2p_peer.rs
```diff
@@ -27,66 +34,146 @@ impl fmt::Display for PeerType {
...
+impl Hash for RemotePeer {
+    fn hash<H: Hasher>(&self, state: &mut H) { self.local_id.hash(state); }
 }
 
 impl RemotePeer {
-    /// Converts a remote peer to a regular peer object, as long as its id is
-    /// known.
+    /// Converts a remote peer to a peer object suitable for transmitting over
+    /// the network, as long as its id is known, i.e., as long as it is
+    /// post-handshake post-handshake.
```
tschudid, line 171: Too many post-handshakes?

concordium-node/src/network/buckets.rs
```diff
@@ -70,15 +69,11 @@ impl Buckets {
     }
 
     /// Returns all the nodes in buckets.
```
tschudid, line 71: with the exception of sender

abizjak: Indeed, fixed

concordium-node/src/common/p2p_peer.rs
abizjak: Indeed, fixed.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -241,7 +246,7 @@ type LogCallback = extern "C" fn(c_char, c_char, *const u8);
 type BroadcastCallback = extern "C" fn(i64, *const u8, i64);
 type CatchUpStatusCallback = extern "C" fn(*const u8, i64);
 type DirectMessageCallback =
-    extern "C" fn(peer_id: PeerId, message_type: i64, msg: *const c_char, msg_len: i64);
+    extern "C" fn(peer_id: PeerIdFFI, message_type: i64, msg: *const c_char, msg_len: i64);
```
tschudid, line 249: Unrelated to changes: Why is msg_len not u64?

tschudid (approved):
Found no major issue (joint review with @chrmatt)

chrmatt (approved):
Looks good.

concordium-node/src/consensus_ffi/ffi.rs
abizjak: No really interesting reason. In Haskell lengths/sizes are (for historical reasons) always Int. I imagine that is the source.

PR #7, opened 2021-04-08
Timeout for block construction.

eb-concordium:
Introducing a timeout for block construction.

### DIFF
--- a/concordium-node/src/consensus_ffi/ffi.rs
+++ b/concordium-node/src/consensus_ffi/ffi.rs
@@ -409,11 +409,11 @@ extern "C" {
 // TODO : Simplify arguments to function, or group with struct
 #[allow(clippy::too_many_arguments)]
 pub fn get_consensus_ptr(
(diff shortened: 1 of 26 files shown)


concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -249,6 +249,7 @@ type RegenesisFreeCallback = unsafe extern "C" fn(*const RwLock<Vec<BlockHash>>)
 extern "C" {
     pub fn startConsensus(
         max_block_size: u64,
+        block_construction_timeout: u64,
```
abizjak, line 252: We'll also want to add this parameter to the config structure.

This is in `concordium-node/concordium-node/src/configuration.rs`, specifically the BakerConfig structure (next to `maximum_block_size`.

abizjak (changes requested):
It's a good start, but some things should be improved.

abizjak (approved):

PR #9, opened 2021-04-11
Trusted connections

abizjak:
## Purpose

Enable the node to always keep some connections to nodes it wants.

## Changes

The main change is in the way connections are handled. There is an additonal class of "given" connections.
So now there are two classes of connections to nodes, the "given" ones and the "discovered" ones. We try to keep being connected to the given ones all the time, but we periodically drop and acquire new "discovered" connections. This latter behaviour is unchanged from what existed before.

The following are also changed as a result
- `--connect-to` is now taken to define a given connection.
- `PeerConnect` GRPC endpoint adds a given connection
- there is a new GRPC endpoint `PeerDisconnect` which removes a connection from the list of given connections, and drops it, if it exists at the time.

## Checklist

- [x] I have documented my code, in particular the intent of the hard-to-understand areas.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ hie.yaml
 **/**/TAGS
 /scripts/genesis/genesis-tmp.json
 .DS_Store
+/concordium-node/*.log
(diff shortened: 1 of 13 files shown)


tschudid (approved):
Looks good.

PR #16, opened 2021-04-19
Restrict when transactions are transmitted.

abizjak:
## Purpose

Transactions which are from accounts that do not exist, or that are not in order
are not retransmitted to peers.

## Changes

# TODO

- [ ] Do the same for update messages
- [ ] Add optional signature verification for transactions and updates before retransmission.
- [ ] Start rejecting transactions when there are too many pending ones.
- [ ] Check that the minimal amount of energy is available for the transaction, otherwise just reject it.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -292,17 +335,37 @@ reversePTT trs ptt0 = foldr reverse1 ptt0 trs
...
+                        let newKnown =
+                              case known of
+                                Seq.Empty -> Seq.singleton sn
+                                (h Seq.:<| _) | h > sn -> sn Seq.<| known
+                                              | otherwise -> known
+                        in Just (low - 1, newKnown)
+
+-- Helper function for updating an entry in the pending transaction table
+-- that is used by addPendingTransaction and addPendingAccount
+updatePttEntry :: (Ord n, Num n) => n -> n -> Maybe (n, Seq.Seq n) -> (Bool, Maybe (n, Seq.Seq n))
```
abizjak, line 365: Please document the meaning of arguments and the result type.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -461,39 +461,70 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
...
 doReceiveTransaction :: (TreeStateMonad pv m, TimeMonad m, SkovQueryMonad pv m) => BlockItem -> Slot -> m UpdateResult
 doReceiveTransaction tr slot = unlessShutDown $ do
-  (_, ur) <- doReceiveTransactionInternal tr slot
-  when (ur == ResultSuccess) $ purgeTransactionTable False =<< currentTime
-  return ur
+  -- Don't accept the transaction if there are too many pending transactions.
+  pendingNum <- numPendingTransactions <$> getPendingTransactions
+  maxPendingNum <- rpMaxPendingTransactionNum <$> getRuntimeParameters
+  if pendingNum > maxPendingNum
+  then return ResultUnverifiable
```
abizjak, line 468: I think this is abusing this return value. We should add a new return code for it.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -461,39 +461,70 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
...
-          Duplicate tx -> return (Just tx, ResultDuplicate)
-          ObsoleteNonce -> return (Nothing, ResultStale)
+doReceiveTransactionInternal :: (TreeStateMonad pv m, SkovQueryMonad pv m) => BlockItem -> Slot -> m (Maybe BlockItem, UpdateResult)
+doReceiveTransactionInternal tr slot = do
+        -- let maxPendingTransactionsPerAccount = 100
+        -- let maxPendingCredentials = 100
+        maxTimeToExpiry <- durationMillis . rpMaxTimeToExpiry <$> getRuntimeParameters
+        slotTime <- getSlotTimestamp slot
+        let expiry = transactionTimeToTimestamp $ msgExpiry tr
+        if tsMillis (expiry - slotTime) > maxTimeToExpiry
```
abizjak, line 486: I think you can just get rid of `tsMillis` here and in line 483 above.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -461,39 +461,70 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
...
+        then return (Just tr, ResultExpiryTooLate)
```
abizjak, line 487: This is problematic since if a transaction is a part of a block we must insert it, and there is no protocol level rule that says that its expiry will be reasonable if it is a part of a block.

So this check should not be here, but only in the `doReceiveTransaction`.

concordium-node/src/configuration.rs
```diff
@@ -44,6 +44,8 @@ pub(crate) fn is_compatible_wire_version(
 
 /// The maximum size of objects accepted from the network.
 pub const PROTOCOL_MAX_MESSAGE_SIZE: u32 = 20_971_520; // 20 MIB
+/// The maximum allowed transaction size in bytes.
+pub const MAX_TRANSACTION_MESSAGE_SIZE: u32 = 100_1000;
```
abizjak, line 48: I don't think this is 100kB as we discussed.

Also, this should not be a constant, but rather a part of this "Configuration" structure so it can be set as a runtime parameter. The default value there should be 100kB.

concordium-node/src/plugins/consensus.rs
```diff
@@ -36,6 +36,8 @@ use std::{
 };
 
 const FILE_NAME_GENESIS_DATA: &str = "genesis.dat";
+// The maximum allowed transaction size in bytes.
+pub const MAX_TRANSACTION_MESSAGE_SIZE: usize = 100_1000;
```
abizjak, line 40: Why is this constant duplicated?

concordium-consensus/src/Concordium/External.hs
```diff
@@ -527,6 +535,7 @@ toReceiveResult ResultContinueCatchUp = 10
 toReceiveResult ResultEarlyBlock = 11
 toReceiveResult ResultMissingImportFile = 12
 toReceiveResult ResultConsensusShutDown = 13
+toReceiveResult ResultExpiryTooLate = 14
```
abizjak, line 538: This datatype must be kept in sync with (1) the comment above, and (2) the type ConsensusFfiResponse in concordium-node/src/consensus_ffi/helpers.rs

concordium-consensus/src/Concordium/External.hs
```diff
@@ -399,7 +405,7 @@ startConsensusPassive ::
            -> CString -> Int64 -- ^Connection string to access the database. If length is 0 don't do logging.
            -> Ptr (StablePtr ConsensusRunner)
            -> IO Int64
-startConsensusPassive timeout maxBlock insertionsBeforePurge transactionsPurgingDelay transactionsKeepAlive gdataC gdataLenC cucbk regenesisArcPtr freeRegenesisArc regenesisCB maxLogLevel lcbk appDataC appDataLenC connStringPtr connStringLen runnerPtrPtr = handleStartExceptions logM $ do
+startConsensusPassive timeout maxTimeToExpiry maxPendingTransactionNum maxBlock insertionsBeforePurge transactionsPurgingDelay transactionsKeepAlive gdataC gdataLenC cucbk regenesisArcPtr freeRegenesisArc regenesisCB maxLogLevel lcbk appDataC appDataLenC connStringPtr connStringLen runnerPtrPtr = handleStartExceptions logM $ do
```
abizjak, line 408: This function and `startConsensus` are called from the rust code in `concordium-node/sr/consensus_ffi/ffi.rs` and the corresponding change must be made there.

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -193,30 +194,77 @@ emptyTransactionTable = TransactionTable {
...
 emptyPendingTransactionTable :: PendingTransactionTable
 emptyPendingTransactionTable = PTT HM.empty HS.empty Map.empty
 
+numPendingCredentials :: PendingTransactionTable -> Int
+numPendingCredentials = HS.size . _pttDeployCredential
+
+numPendingTransactions :: PendingTransactionTable -> Int
+numPendingTransactions table = numPendingCredentials table + countNonces (_pttWithSender table) + countNonces (_pttUpdates table)
+  where countNonces :: Foldable (m a) => m a (b, Seq.Seq c) -> Int 
+        countNonces = foldl' (\s (_, ns) -> s + length ns) 0
```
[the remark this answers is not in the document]
amaurremi: Removing this functionality, as discussed.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -517,6 +513,18 @@ stopBaker cptr = mask_ $
 +-------+------------------------------------+----------------------------------------------------------------------------------------+----------+
 |    13 | ResultConsensusShutDown            | Consensus has been shut down and the message was ignored                               | No       |
 +-------+------------------------------------+----------------------------------------------------------------------------------------+----------+
+|    14 | ResultExpiryTooLate                | The transaction expiry time is too far in the future                                   | No       |
++-------+------------------------------------+----------------------------------------------------------------------------------------+----------+
+|    15 | ResultVerificationFailed           | The transaction signature verification failed                                          | No       |
++-------+------------------------------------+----------------------------------------------------------------------------------------+----------+
+|    16 | ResultNonexistingSenderAccount     | The transaction's sender account does not exist according to the focus block           | No       |
```
amaurremi, line 520: Could it be the case that our focus block is wrong, and we couldn't look up the account by mistake? Should we still retransmit transactions in that case?

concordium-node/src/configuration.rs
```diff
@@ -225,6 +225,13 @@ pub struct BakerConfig {
         long = "import-blocks-from",
         help = "Path to a file exported by the database exporter"
     )]
+    #[structopt(
+        long = "max-expiry-duration",
+        help = "Maximum allowed time difference between now and a transaction's expiry time in \
+                milliseconds",
```
amaurremi, line 231: Nobody will have a max expiry granularity in milliseconds so this should be probably changed.

abizjak:
This should be done more systematically and with more thought now that we have more time.

PR #19, opened 2021-04-20
Accept multiple connections per event.

abizjak:
## Purpose

Process all incoming connections that are available for any readiness event that is triggered.
In the previous handing some incoming connections could be dropped, depending on the setup and timing.
This could lead to the node no longer accepting any new connections since no events were triggered.

## Changes

Iterate attempting to accept connections for as long as there are any available.
We stop after a fixed amount of time to prevent abuse, and we also stop if we already have enough connections.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/p2p/connectivity.rs
+++ b/concordium-node/src/p2p/connectivity.rs
@@ -330,7 +330,9 @@ impl P2PNode {
 
 #[derive(Debug)]
 pub enum AcceptFailureReason {
-    TooManyConnections,
+    TooManyConnections {
+        addr: SocketAddr,
+    },
     AlreadyConnectedToIP {
         ip: IpAddr,
     },
@@ -347,9 +349,13 @@ pub enum AcceptFailureReason {
 impl std::fmt::Display for AcceptFailureReason {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            AcceptFailureReason::TooManyConnections => {
-                f.write_str("Too many existing connections. Not accepting an additional one.")
-            }
+            AcceptFailureReason::TooManyConnections {
+                addr,
+            } => write!(
+                f,
+                "Too many existing connections. Not accepting an additional one from {}.",
+                addr
+            ),
             AcceptFailureReason::AlreadyConnectedToIP {
                 ip,
             } => write!(f, "Already connected to IP {}.", ip),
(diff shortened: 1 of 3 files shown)


concordium-node/src/p2p/maintenance.rs
```diff
@@ -614,13 +626,34 @@ pub fn spawn(node_ref: &Arc<P2PNode>, mut poll: Poll, consensus: Option<Consensu
...
+                        }
+                    }
+                    attempt_number += 1;
                 }
-                if i == 9 {
-                    warn!("too many connection attempts received at once; dropping the rest");
-                    break;
+                if attempt_number >= MAX_NUM_ACCEPT_ATTEMPTS {
+                    warn!(
+                        "Received 9 or more connections at once. Delaying accepting the remaining \
```
td202, line 654:
```suggestion
                        "Received too many connection requests at once. Delaying accepting the remaining \
```

concordium-node/src/p2p/connectivity.rs
```diff
@@ -362,24 +411,28 @@ pub fn accept(node: &Arc<P2PNode>) -> Fallible<Option<Token>> {
             && candidates_lock.len() + conn_read_lock.len()
                 >= node.config.hard_connection_limit as usize
         {
-            bail!("Too many connections, rejecting attempt from {}", addr);
+            return Err(AcceptFailureReason::TooManyConnections);
```
td202, line 414: It could be helpful to log the source address, since we're dropping this connection.

concordium-node/src/p2p/maintenance.rs
```diff
@@ -54,6 +55,10 @@ use std::{
     time::{Duration, Instant},
 };
 
+/// Maximum number of incoming connections to attempt to accept per iteration
+/// of the main connection maintenance loop.
+pub const MAX_NUM_ACCEPT_ATTEMPTS: u32 = 9;
```
td202, line 60: It would probably be better to make this part of `NodeConfig` (and correspondingly `ConnectionConfig`).

td202 (changes requested):

concordium-node/src/p2p/connectivity.rs
abizjak: Done.

concordium-node/src/p2p/maintenance.rs
abizjak: Done.

concordium-node/src/p2p/maintenance.rs
abizjak: Done.

td202 (approved):

PR #22, opened 2021-04-20
Reduce logging spam in case of many bad events.

abizjak:
## Purpose

In cases where there are a lot of invalid messages or just too many messages we have seen the logs get very large which causes further problems, and the node ends up using a lot of resources to simply emit the logs.

The purpose of this PR is to reduce the amount of logging and instead summarize the logs.

## Changes

- Summarize the bad events that happen on the network layer and emit them only at each housekeeping interval.
- In the scheduler only emit the first few invalid transactions. This makes logging still useful for debugging but also makes it so that blocks are produced in a reasonable amount of time. Before this change for example, it happened regularly that it took 10 or more seconds just to log all the invalid transactions after filterTransactions ran.

## Checklist

      hard-to-understand areas.


td202 (approved):
This seems fine, but I'm a bit concerned that having a lock and map for each of these counters seem quite high overhead. Ideally, I think these would be atomic counters in `ConnectionStats`.

PR #26, opened 2021-04-21
Add new ARs and IPs through chain updates

Bargsteen:
## Purpose

Allows us to add new anonymity revokers and identity providers through a chain update.

## Changes

- Handle the two new update types (added in https://github.com/Concordium/concordium-base/pull/18).
- Change signature of `processUpdateQueues` (both in Persistent and Basic) to include the ARs and IPs.
- Process AR and IP updates in a custom manner for two primary reasons:
   - We need to handle _all_ updates and not just the newest, since we are not simply replacing the existing value.
   - We do not allow existing ARs/IPs to be overwritten, so we ignore updates with duplicates ids.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/Updates.hs
```diff
@@ -324,14 +343,40 @@ processProtocolUpdates t pu0 uq = (pu', uq {_uqQueue = qr}, Map.fromAscList ql)
...
+-- Instead, it is handled here by ignoring updates with duplicate IPs/ARs.
+processARsAndIPsUpdates :: Ord k
+                        => Map.Map k v -- ^ The existing IPs / ARs.
+                        -> (v -> k) -- ^ Getter for the key field.
+                        -> Timestamp
+                        -> UpdateQueue v
+                        -> (Map.Map k v, UpdateQueue v, Map.Map TransactionTime v)
+processARsAndIPsUpdates oldValMap getKey t uq = (updatedValMap, uq {_uqQueue = qr}, changes)
+  where (ql, qr) = span ((<= t) . transactionTimeToTimestamp . fst) (uq ^. uqQueue)
+        (changes, updatedValMap) = foldr go (Map.empty, oldValMap) ql
```
abizjak, line 357: The use of foldr implicitly makes it so that updates with the later timestamp will be preferred if there is a clash of ids.

Is that the intention? If so, please document the intent.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
```diff
@@ -605,6 +634,65 @@ processBakerStakeThresholdUpdates t bu = do
...
+  => Timestamp
+  -> BufferedRef Updates
+  -> HashedBufferedRef ARS.AnonymityRevokers
+  -> m (Map.Map TransactionTime UpdateValue, BufferedRef Updates, HashedBufferedRef ARS.AnonymityRevokers)
+processAddAnonymityRevokerUpdates t bu hbar = do
+    u@Updates{..} <- refLoad bu
+    oldARs <- ARS.arRevokers <$> refLoad hbar
+    oldQ <- refLoad (pAddAnonymityRevokerQueue pendingUpdates)
+    let (ql, qr) = Seq.spanl ((<= t) . transactionTimeToTimestamp . fst) (uqQueue oldQ)
+    (changes, updatedARs) <- addAndAccumNonduplicateUpdates oldARs ARS.arIdentity UVAddAnonymityRevoker ql
```
abizjak, line 649: Why does this not return the `AnonymityRevokers` structure directly (instead of the Map it returns now)?

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
```diff
@@ -605,6 +634,65 @@ processBakerStakeThresholdUpdates t bu = do
...
+-- Ensuring that new IPs/ARs have unique ids is difficult when enqueueing.
+-- Instead, it is handled here by ignoring updates with duplicate IPs/ARs.
+-- It also accumulates the actual changes that occured.
+addAndAccumNonduplicateUpdates :: (MonadBlobStore m, Foldable f, Reference m ref (StoreSerialized v), Ord k)
+                      => StrictMap.Map k v -- ^ The existing IPs / ARs.
+                      -> (v -> k) -- ^ Getter for the key field.
+                      -> (v -> UpdateValue) -- ^ Data constructor for UpdateValue.
+                      -> f (TransactionTime, ref (StoreSerialized v)) -- ^ The updates.
+                      -> m (Map.Map TransactionTime UpdateValue, StrictMap.Map k v)
+addAndAccumNonduplicateUpdates oldMap getKey toUV = foldM go (Map.empty, oldMap)
```
abizjak, line 685: The use of foldM here is inconsistent with the use of `foldr` in the basic block state case. They traverse the list differently so they will disagree in cases where there are conflicting updates.

abizjak (changes requested):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
Bargsteen: The reason is that it is used for both `AnonymityRevokers` and `IdentityProviders`.
Perhaps there is some nice way to unwrap and wrap these newtypes without adding more parameters to the function?

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
```diff
@@ -5,6 +5,7 @@ module Concordium.GlobalState.Persistent.BlockState.Updates where
 
 import Data.Foldable (toList)
 import qualified Data.Map as Map
+import qualified Data.Map.Strict as StrictMap
```
td202, line 8: I'd suggest changing to:
```suggestion
import qualified Data.Map.Strict as Map
```
All the existing uses of Map here may as well use the strict functions. (Note that `Data.Map.Strict` and `Data.Map` use the same underlying data structure. The only difference is that the functions in `Data.Map.Strict` force head-normal-form evaluation of values that are stored in the map.)

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
```diff
@@ -605,6 +634,65 @@ processBakerStakeThresholdUpdates t bu = do
...
+    u@Updates{..} <- refLoad bu
+    oldIPs <- IPS.idProviders <$> refLoad hbip
+    oldQ <- refLoad (pAddIdentityProviderQueue pendingUpdates)
+    let (ql, qr) = Seq.spanl ((<= t) . transactionTimeToTimestamp . fst) (uqQueue oldQ)
+    (changes, updatedIPs) <- addAndAccumNonduplicateUpdates oldIPs IPS.ipIdentity UVAddIdentityProvider ql
+
+    newQ <- refMake oldQ { uqQueue = qr }
+    newU <- refMake u { pendingUpdates = pendingUpdates {pAddIdentityProviderQueue = newQ} }
+    updatedIPsRef <- refMake . IPS.IdentityProviders $ updatedIPs
+    return (changes, newU, updatedIPsRef)
```
td202, line 673: If no changes are made, then we should not make new references, but return the old ones. This will be extremely wasteful as we will end up storing a new copy of the identity providers map for every (finalized) block. (This also applies to the anonymity revokers.)

td202 (changes requested):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
Bargsteen: Ah, great catch! Should be fixed now 😊

abizjak (approved):

td202 (approved):

PR #30, opened 2021-04-22
Publish stagenet build only if the tag has the expected format

bisgardo:


abizjak:
Do you want to merge this soon @bisgardo @Thahara ?

PR #31, opened 2021-04-23
Fix issue with bootstrapping when name resolution is slow

td202:
## Purpose

This pull request addresses an issue where bootstrapping fails in some circumstances. The failure occurs when name resolution for the bootstrapper nodes takes a long time. Connection housekeeping ends up running on each iteration of the connection polling loop and attempting to bootstrap. As a result, the new connections to the bootstrapper nodes do not get the opportunity to complete the connection handshake before they are considered "old" and are closed.

## Changes

This PR implements two mitigations:

1. Record the time of the last housekeeping after the housekeeping has occurred. This means that the time that is spent in housekeeping (e.g. resolving bootstrapper nodes) is not included in the time measured between housekeeping events. This should prevent slow housekeeping (and in particular, slow name resolution for bootstrappers) from causing housekeeping to happen at an unduly high frequency, and thus prematurely closing handshaking connections.
2. Use a counter to ensure that housekeeping can only occur at most once every 11 iterations. (The number of iterations is fairly arbitrary, but should be enough to complete the handshake, but not so high as to unduly increase the housekeeping interval.) This double check should ensure that bootstrapping connections have sufficient opportunity to complete the handshake even if something other than housekeeping causes (significant) delays in the polling loop.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/p2p/maintenance.rs
+++ b/concordium-node/src/p2p/maintenance.rs
@@ -641,6 +641,8 @@ pub fn spawn(node_ref: &Arc<P2PNode>, mut poll: Poll, consensus: Option<Consensu
         let mut log_time = Instant::now();
         let mut last_buckets_cleaned = Instant::now();
         let mut last_peer_list_update = 0;
+        // The number of polling loop iterations since the last housekeeping.
+        let mut iterations_since_housekeeping = 0;
 
         let num_socket_threads = match node.self_peer.peer_type {
             PeerType::Bootstrapper => 1,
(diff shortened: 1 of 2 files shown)


abizjak (approved):
I'm not sure about the counter since it might cause problems with delays in new places we have not anticipated but in normal circumstances it should work fine.

An idea might also be to only do DNS resolution in the beginning and not while the node is running. But we can do that separately if it would make sense.

PR #34, opened 2021-04-25
Do not accept transactions that are never (or not very likely) going to make it into a block.

abizjak:
## Purpose

Limit what kinds of transactions are accepted by the node before they are added to the input pool.

## Changes

Check the following when accepting transactions
- transaction's energy is too low to ever be included in a block
- transaction's expiry time is too far in the future. The amount is set as a runtime parameter.
- transactions's size is too large
- introduce a new constant PROTOCOL_MAX_TRANSACTION_SIZE. This is checked at the earliest feasible point so that we do not store large objects in memory
- account does not exist in the last finalized block

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -304,6 +304,7 @@ handleStartExceptions logM c =
 startConsensus ::
            Word64 -- ^Maximum block size.
            -> Word64 -- ^Block construction timeout in milliseconds
+           -> Word64 -- ^The amount of time in the future a transaction's expiry can be. In seconds.
            -> Word64 -- ^Insertions before purging of transactions
            -> Word64 -- ^Time in seconds during which a transaction can't be purged
            -> Word64 -- ^Number of seconds between transaction table purging runs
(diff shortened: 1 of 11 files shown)


concordium-consensus/src/Concordium/GlobalState/Parameters.hs
```diff
@@ -101,7 +101,9 @@ data RuntimeParameters = RuntimeParameters {
   -- transaction table if a purge is executed.
   rpTransactionsKeepAliveTime :: !TransactionTime,
   -- |Number of seconds between automatic transaction table purging  runs.
-  rpTransactionsPurgingDelay :: !Int
+  rpTransactionsPurgingDelay :: !Int,
+  -- |The maximum allowed time difference between slot time and a transaction's expiry time.
+  rpMaxTimeToExpiry :: !TransactionTime
```
abizjak, line 106: This would ideally be `Duration`, but `rpTransactionKeepAliveTime` should also be that, so I kept it consistent with that so that I don't introduce accidental bugs.

amaurremi (approved):

concordium-consensus/src/Concordium/External.hs
```diff
@@ -399,7 +402,7 @@ startConsensusPassive ::
            -> CString -> Int64 -- ^Connection string to access the database. If length is 0 don't do logging.
            -> Ptr (StablePtr ConsensusRunner)
            -> IO Int64
-startConsensusPassive timeout maxBlock insertionsBeforePurge transactionsPurgingDelay transactionsKeepAlive gdataC gdataLenC cucbk regenesisArcPtr freeRegenesisArc regenesisCB maxLogLevel lcbk appDataC appDataLenC connStringPtr connStringLen runnerPtrPtr = handleStartExceptions logM $ do
+startConsensusPassive timeout maxTimeToExpiry maxBlock insertionsBeforePurge transactionsPurgingDelay transactionsKeepAlive gdataC gdataLenC cucbk regenesisArcPtr freeRegenesisArc regenesisCB maxLogLevel lcbk appDataC appDataLenC connStringPtr connStringLen runnerPtrPtr = handleStartExceptions logM $ do
```
td202, line 405: The order of arguments does not match the documentation. `maxBlock` should be first.

abizjak: Fixed in #35

Issue #28, opened 2021-04-22
Replace failure with anyhow 


abizjak:
**Task description**

The failure crate is deprecated. Its functionality is largely replaced by anyhow, but there are some minor differences in error handling

**Sub-tasks**
- [ ] replace failure with anyhow + thiserror in concordium-base/rust-src
- [ ] replace failure with anyhow in the node

abizjak:
Fixed by #49

PR #49, opened 2021-05-07
Replace failure with anyhow

EmilLa1:
## Purpose

This PR is in response to the **second** item mentioned in this [issue](https://github.com/Concordium/concordium-node/issues/28). 

The purpose is to replace the [deprecated](https://github.com/rust-lang-nursery/failure) '[failure](https://docs.rs/failure/0.1.8/failure/)' crate with the '[anyhow](https://docs.rs/anyhow/1.0.40/anyhow/)' and '[thiserror](https://docs.rs/thiserror/1.0.24/thiserror/)'.

## Changes

Removing all usages of the failure crate and replace it with appropriate usages of the 'anyhow' and 'thiserror'. 

Note: 
Prior to being merged, the submodule concordium-base/ has to be pinned to the correct commit. 
(This cannot be done before https://github.com/Concordium/concordium-base/pull/37 is merged.)

## Checklist

      hard-to-understand areas.


concordium-node/src/bin/bootstrap_checker.rs
```diff
@@ -1,26 +1,25 @@
...
     p2p::maintenance::{attempt_bootstrap, spawn, P2PNode},
     stats_export_service::instantiate_stats_export_engine,
     utils::get_config_and_logging_setup,
 };
-use failure::{ensure, Error};
 
 use std::{env, process::Command, thread, time::Duration};
 
-fn main() -> Result<(), Error> {
+fn main() -> Result<(), anyhow::Error> {
```
abizjak, line 22: This could just be `anyhow::Result<()>` instead.

concordium-node/src/bin/bootstrapper.rs
```diff
@@ -8,19 +8,19 @@ use std::{
...
     stats_export_service::instantiate_stats_export_engine,
     utils::get_config_and_logging_setup,
 };
-use failure::{ensure, Error, ResultExt};
 
 #[cfg(feature = "instrumentation")]
 use concordium_node::stats_export_service::start_push_gateway;
 
-fn main() -> Result<(), Error> {
+fn main() -> anyhow::Result<(), anyhow::Error> {
```
abizjak, line 23: And this could just be `anyhow::Result<()>`

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -66,8 +66,8 @@ pub struct HashBytes([u8; SHA256 as usize]);
 impl HashBytes {
     /// Construct HashBytes from a slice.
     /// This will succeed if and only if the length of the slice is 32.
-    pub fn new(bytes: &[u8]) -> Fallible<Self> {
-        let buf: Result<[u8; 32], _> = bytes.try_into();
+    pub fn new(bytes: &[u8]) -> anyhow::Result<Self> {
+        let buf: anyhow::Result<[u8; 32], _> = bytes.try_into();
         let buf = buf.context("HashBytes::new passed slice of incorrect length.")?;
```
abizjak, line 71: I don't think this context is necessary. That will already be the message from the `try_into`.

Wouldn't just `let buf = bytes.try_into()?` or `let buf = bytes.try_into().context("HashBytes::new. ...)?` work without ugly type annotations.

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -141,14 +141,14 @@ static PACKET_TYPE_FROM_INT: &[PacketType] = &[
...
+    type Error = anyhow::Error;
 
     #[inline]
-    fn try_from(value: u8) -> Fallible<PacketType> {
+    fn try_from(value: u8) -> anyhow::Result<PacketType> {
         PACKET_TYPE_FROM_INT
             .get(value as usize)
             .copied()
-            .ok_or_else(|| format_err!("Unsupported packet type ({})", value))
+            .ok_or_else(|| anyhow!("Unsupported packet type ({})", value))
```
abizjak, line 151: With `anyhow` this can be simplified to `.get(value as usize).copied().context(format!("Unsupported packed type ({})", value))

Not sure if that's better or not.

concordium-node/src/network/serialization/fbs.rs
```diff
@@ -198,7 +196,7 @@ fn deserialize_request(root: &network::NetworkMessage) -> Fallible<NetworkPayloa
                             wv.genesis_block()
                                 .map_or_else(|| bail!("Missing block hash"), |w| BlockHash::new(w))
                         })
-                        .collect::<Result<Vec<BlockHash>, failure::Error>>()?
+                        .collect::<Result<Vec<BlockHash>, anyhow::Error>>()?
```
abizjak, line 199: Use `anyhow::Result<Vec<BlockHash>>`

concordium-node/src/p2p/maintenance.rs
```diff
@@ -294,7 +294,7 @@ impl P2PNode {
             IpAddr::from_str(addy).context("Could not parse the provided listen address.")?
         } else {
             P2PNode::get_ip().ok_or_else(|| {
-                format_err!("Could not compute my own ip. Use `--listen-address` to specify it.")
+                anyhow!("Could not compute my own ip. Use `--listen-address` to specify it.")
```
abizjak, line 297: Similar to above, you can use 
```rust
P2PNode::get_ip().context("Could not compute my own ip. Use `--listen-address` to specify it.")?
```
instead.

concordium-node/src/consensus_ffi/helpers.rs
EmilLa1: We should probably do it anyway for consistency (wrt. the other comments you have).

Issue #43, opened 2021-05-03
Update flatbuffers compiler and dependency


abizjak:
**Task description**

The node currently needs a very specific old version of the flatbuffers compiler to build.
We should investigate whether we can remove that restriction and update the flatbuffers dependency to above 0.7

**Sub-tasks**
- [x] Make code compatible with recent versions of flatc
- [ ] Update the flatbuffers dependency

PR #50, opened 2021-05-10
Update flatbuffers and flatc dependencies.

MilkywayPirate:
## Purpose

Updating the build requirements information with the correct information on which version of [flatc](https://github.com/google/flatbuffers) is required to build a `concordium-node`. 

Moreover added a memo, that the `flatbuffers` dependency should be updated when we upgrade the Rust version.

This "fixes" the first item in https://github.com/Concordium/concordium-node/issues/43 Hence the "fixes" - it did work before as well - only difference is now, that it is documented in the README.md. 

The second item in https://github.com/Concordium/concordium-node/issues/43 must wait until we update Rust version. (This is not also noted in the Cargo.toml.)

## Changes

Updated the README.md and Cargo.toml with appropriate comments.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -63,7 +63,7 @@ jobs:
         #token: ${{ secrets.CONCORDIUM_CI }}
         submodules: recursive
     - name: Install system packages
-      run: sudo apt-get -y install liblmdb-dev flatbuffers-compiler libunbound-dev
+      run: sudo apt-get update && sudo apt-get -y install liblmdb-dev flatbuffers-compiler libunbound-dev
 
     # Set up Rust and restore dependencies and targets from cache.
     # This must be done before checking the Rust sources (obviously)


concordium-node/README.md
```diff
@@ -3,9 +3,7 @@
 * Rust (stable 1.45.2 for using static libraries)
 * binutils >= 2.22
 * cmake >= 3.8.0
-* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html)
-  commit fec58aa129818ed0c0613a7ec36b55135bf81278, but others around it are
-  likely to work as well (build using CMake and copy to `~/.local/bin`)
+* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html) >= 1.12.0 (build using CMake and copy to `~/.local/bin`)
```
abizjak, line 6: I think we should say that 1.11 and 1.12 are known to work, we don't know if `1.13` will work or not, so `>= 1.12` is probably wrong.

abizjak (changes requested):

abizjak (approved):

Issue #44, opened 2021-05-03
Remove serde serialization of network messages


abizjak:
**Task description**

We currently have serde_cbor and serde_msgpack variants of network message serialization. They are behind features and have never been used other than for benchmarks a long time ago.

They incur a maintenance burden so we should remove them. They are almost certainly in disrepair. Flatbuffers works well for us.

abizjak:
Fixed by #52

PR #52, opened 2021-05-10
Removed serde_cbor and serde_msgpack from network serialization.

EmilLa1:
## Purpose

To remove usages of serde_msgpack and serde_cbor in the network serialization.
This is according to the issue: https://github.com/Concordium/concordium-node/issues/44

## Changes

Removed the concrete implementation using serde_msgpack and serde_cbor together with the dependencies. 

## Checklist

      hard-to-understand areas.


abizjak:
There are still quite a few leftovers
- in many places there is 
```
#[cfg_attr(feature = "s11n_serde", derive(Serialize, Deserialize))]
```
or similar network related types.

- there is a macro only_fbs which should be removed since we now only support flatbuffers so there is no need to go through the indirection of that macro.

- I think the `s11n_serde` feature can be entirely removed since it is only used for network messages. All occurrences of conditional compilation (e.g., cfg(feature = "s11n_serde") or `cfg(not(feature = "s11n_serde")`) can be removed.

abizjak (approved):

PR #54, opened 2021-05-11
Fix testnet disto image

bisgardo:
- Fetch genesis-data from git.
- Build static libraries instead of downloading.
- Rename dockerfile.


Thahara (approved):

scripts/distribution/testnet.Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+ARG base_image_tag
+ARG static_libraries_image_tag
+
+# Fetch genesis-data.
+FROM alpine/git:latest as genesis-data
+ARG genesis_ref
+ARG genesis_path
+RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+RUN --mount=type=ssh git clone --depth 1 --branch "${genesis_ref}" git@gitlab.com:Concordium/genesis-data.git /data
+RUN mv "/data/${genesis_path}/genesis_data" /genesis-data
```
abizjak, line 13: The only thing that should be needed from genesis data is `genesis.dat` file.

I think its a bad idea to copy in private keys in this stage and then only taking out the `genesis.dat` in the stage at the end.

abizjak (approved):

Issue #46, opened 2021-05-04
Average throughput is not maintained correctly


td202:
**Bug Description**
The send/receive throughput reported via the GRPC APIs seem to be inaccurate, and generally report 1bps sent/received.

**Steps to Reproduce**
Run a node. Observe the average sent/received values in the node dashboard.

(Note, as far as I can tell, these stats are not currently available through a query from `concordium-client` query.)

**Expected Result**
Reasonable values > 0.001 kB/s.

**Actual Result**
0.001 kB/s.

**Versions**
- Software Version: 0.7.2
- OS: Windows

abizjak:
Fixed by #56

PR #56, opened 2021-05-12
Fixed bug in average throughput calculation.

EmilLa1:
## Purpose

To fix the bug mentioned here: https://github.com/Concordium/concordium-node/issues/46. 

## Changes

Instead of relying on the housekeeping interval for calculating the average of bps sent/received for the node, the function
`measure_throughput` now uses its own internal timestamp: `throughput_timestamp` as a delta time for calculating the average bps.

## Checklist

      hard-to-understand areas.


concordium-node/src/p2p/peers.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
     p2p::{maintenance::attempt_bootstrap, P2PNode},
     read_or_die,
 };
-
+use chrono::prelude::*;
```
abizjak, line 12: Could we be a bit more precise in importing. I think only a few definitions are needed, and this imports a lot more.

concordium-node/src/p2p/peers.rs
```diff
@@ -65,14 +66,22 @@ impl P2PNode {
...
+        let (avg_bpms_in, avg_bpms_out) = calculate_average_throughput(
+            self.stats.get_last_throughput_measurement(),
+            now,
+            prev_bytes_received * 1000,
+            bytes_received * 1000,
+            prev_bytes_sent * 1000,
+            bytes_sent * 1000,
+        );
+
+        let (avg_bps_in, avg_bps_out) = (avg_bpms_in, avg_bpms_out);
```
abizjak, line 81: I don't understand what this renaming is for. Shouldn't there be some rescaling?

concordium-node/src/p2p/peers.rs
```diff
@@ -145,3 +154,47 @@ pub fn check_peers(node: &Arc<P2PNode>, peer_stats: &[PeerStats], attempted_boot
         }
     }
 }
+
+/// Calculate the `checked` average of the received and send bytes.
+/// If any of the arguments would imply in a non-defined or otherwise
+/// non-meaningful result, then the function returns (0, 0).
+fn calculate_average_throughput(
```
abizjak, line 161: I think the "correct" thing to do would be to return a result or option. 

These defaulting values are not in the spirit of the rest of the project.

concordium-node/src/stats_export_service.rs
```diff
@@ -352,6 +360,26 @@ impl StatsExportService {
         self.outbound_low_priority_consensus_size.store(value as usize, Ordering::Relaxed);
     }
 
+    /// Gets the timestamp for the last throughput check.
+    pub fn get_last_throughput_measurement(&self) -> i64 {
```
abizjak, line 364: Shouldn't this be named with `timestamp` somewher in the string so it is clear that is what it returning.

concordium-node/src/stats_export_service.rs
```diff
@@ -352,6 +360,26 @@ impl StatsExportService {
...
+            self.last_throughput_measurement.get()
+        }
+        #[cfg(not(feature = "instrumentation"))]
+        {
+            self.last_throughput_measurement.load(std::sync::atomic::Ordering::Relaxed)
+        }
+    }
+
+    /// Sets the value of throughput timestamp.
+    pub fn set_last_throughput_measurement(&self, value: i64) {
```
abizjak, line 376: And likewise this.

concordium-node/src/p2p/peers.rs
```diff
@@ -51,7 +51,8 @@ impl P2PNode {
         self.get_peer_stats(Some(PeerType::Node)).into_iter().map(|stats| stats.local_id).collect()
     }
 
-    /// Measures the node's average byte throughput.
+    /// Measures the node's average byte throughput as bps i.e., bytes per
+    /// second.
     pub fn measure_throughput(&self, peer_stats: &[PeerStats]) -> (u64, u64) {
```
abizjak, line 56: Since this function can in principle fail I would recommend doing the following.

- have the function return `anyhow::Result<()>` (the output is never used, this function just modifies the state).
- in `maintanance` where the function is used do something like 
```if let Err(e) = measure_throuput(...) {
       error!("Could not measure througput: {}", e)
```
which will log the fact that this failed (and do nothing else).

concordium-node/src/p2p/peers.rs
```diff
@@ -65,14 +66,22 @@ impl P2PNode {
...
+        let now = Utc::now().timestamp_millis(); // Timestamps are stored in millis in order to allow measurements within the
+                                                 // second. Hence the number of bytes received/send must be multiplied by a
+                                                 // factor 1000 in order to obtain bps.
+        let (avg_bpms_in, avg_bpms_out) = calculate_average_throughput(
+            self.stats.get_last_throughput_measurement(),
+            now,
+            prev_bytes_received * 1000,
+            bytes_received * 1000,
+            prev_bytes_sent * 1000,
+            bytes_sent * 1000,
```
abizjak, line 78: It would be cleaner to do these multiplications in the `calculate_average_througput` function.

abizjak (approved):

abizjak (approved):

PR #62, opened 2021-05-17
Added a rustup toolchain file in order to pin the Rust version.

EmilLa1:
## Purpose

The concordium-node uses a pinned version of Rust.
The version can be pinned in a `rustup-toolchain` file to make it easier for developers i.e., developers no longer manually has to set the toolchain version for the repository. 

## Changeso

Added a rustup-toolchain file which pins the Rust version (Currently v1.45.2).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/rust-toolchain
+++ b/concordium-node/rust-toolchain
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.45.2"


abizjak (commented):
I don't particularly like this change since it is hiding configuration in an obscure file.

Ideally our project would work with more than a single version. It is currently an unfortunate reality that we don't support more recent versions.

I would prefer not to have this, but if you insist then this needs to be documented otherwise it is going to lead to obscure failures.

EmilLa1:
Actually I agree it could result in unexpected behaviour, which could be hard to figure out. Also even if it was documented. 
Maybe it's best as it is now.

PR #64, opened 2021-05-18
Do not allow rebroadcasting of catch-up status messages.

td202:
## Purpose

Prevent a node from relaying a catch-up status message to its peers even if it is sent as a broadcast message.

## Changes

When deciding whether a message can be rebroadcast, check that its type supports it.
Define `is_rebroadcastable` for `PacketType`, disallowing rebroadcasts for catch-up status messages.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/consensus_ffi/helpers.rs
+++ b/concordium-node/src/consensus_ffi/helpers.rs
@@ -170,7 +170,7 @@ impl PacketType {
     /// Determine whether a packet type can be relayed.
     /// Those that are must be subject to appropriate de-duplication
     /// checks to ensure they are not relayed endlessly.
-    fn is_rebroadcastable(&self) -> bool {
+    pub fn is_rebroadcastable(&self) -> bool {
         match self {
             PacketType::Block => true,
             PacketType::Transaction => true,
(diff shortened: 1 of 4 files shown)


abizjak (approved):
Could you please add this bugfix to the changelog before merging?

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -166,6 +166,21 @@ impl fmt::Display for PacketType {
...
+    /// Those that are must be subject to appropriate de-duplication
+    /// checks to ensure they are not relayed endlessly.
+    pub fn is_rebroadcastable(&self) -> bool {
+        match self {
+            PacketType::Block => true,
+            PacketType::Transaction => true,
+            PacketType::FinalizationRecord => true,
+            PacketType::FinalizationMessage => true,
+            PacketType::CatchUpStatus => false,
+        }
```
tschudid, line 180: Maybe add
`_ => false,`?

abizjak: > Maybe add
> `_ => false,`?

The list already contains all the different package types, so this does not add anything. The compiler checks completeness of the patterns.

td202: I think it's better not to have a default case here, so that if we add a new message type then the compiler will require us to be explicit about whether it should be rebroadcastable.

PR #78, opened 2021-06-04
getenvoy fix

EmilLa1:
## Purpose

Fixing the docker distribution builds. 
It used a no longer supported `getenvoy` version, which resulted in failed builds. 

## Changes

Updated the usages of `getenvoy` such that it accommodates the new API. 

- [x] Confirm that the newly generated docker distribution image actually works. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/distribution/builder.Dockerfile
+++ b/scripts/distribution/builder.Dockerfile
@@ -78,7 +78,7 @@ ENV ENABLE_TERM_HANDLER=true
 
 RUN apt-get update && apt-get install -y unbound curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq
 RUN curl -L https://getenvoy.io/cli | bash -s -- -b /usr/local/bin
-RUN getenvoy fetch standard:1.17.0
+RUN getenvoy use 1.18.3
 
 COPY --from=node-dashboard /static /node-dashboard/static
 COPY --from=node-dashboard /envoy.yaml /node-dashboard/envoy.yaml
--- a/scripts/distribution/builder.Dockerfile
+++ b/scripts/distribution/builder.Dockerfile
@@ -77,7 +77,7 @@ ENV DISTRIBUTION_CLIENT=true
 ENV ENABLE_TERM_HANDLER=true
 
 RUN apt-get update && apt-get install -y unbound curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq
-RUN curl -L https://getenvoy.io/cli | bash -s -- -b /usr/local/bin
+RUN curl -L https://getenvoy.io/install.sh | bash -s -- -b /usr/local/bin
 RUN getenvoy use 1.18.3
 
 COPY --from=node-dashboard /static /node-dashboard/static
--- a/scripts/distribution/concordium.conf
+++ b/scripts/distribution/concordium.conf
@@ -16,7 +16,7 @@ user=docker
 startretries=50
 
 [program:grpc-web-proxy]
-command=getenvoy run standard:1.17.0 -- --config-path /node-dashboard/envoy.yaml
+command=getenvoy run -c /node-dashboard/envoy.yaml
 autostart=true
 autorestart=true
 startretries=50


scripts/distribution/builder.Dockerfile
```diff
@@ -78,7 +78,7 @@ ENV ENABLE_TERM_HANDLER=true
 
 RUN apt-get update && apt-get install -y unbound curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq
 RUN curl -L https://getenvoy.io/cli | bash -s -- -b /usr/local/bin
```
td202, line 80: Possibly we should change this to:
```suggestion
RUN curl -L https://getenvoy.io/install.sh | bash -s -- -b /usr/local/bin
```
in line with the current getenvoy documentation. In any case, both currently redirect to the same URL, but who knows if they'll decide to change that.

EmilLa1: Yes I agree. I think we should look in other ways of acquiring the getenvoy.

scripts/distribution/builder.Dockerfile
```diff
@@ -77,8 +77,8 @@ ENV DISTRIBUTION_CLIENT=true
 ENV ENABLE_TERM_HANDLER=true
 
 RUN apt-get update && apt-get install -y unbound curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq
-RUN curl -L https://getenvoy.io/cli | bash -s -- -b /usr/local/bin
-RUN getenvoy fetch standard:1.17.0
+RUN curl -L https://getenvoy.io/install.sh | bash -s -- -b /usr/local/bin
+RUN getenvoy use 1.18.3
```
abizjak, line 81: I don't think the versions should just be changed without explicitly planning to do so, and planning to test the resulting version bump.

abizjak (changes requested):

scripts/distribution/builder.Dockerfile
td202: The new release of `getenvoy` doesn't work with older versions of `envoy`. If the new version works (which it seems to) then I think it would be better just to use it. Otherwise, we need a better way of getting a fixed version of envoy.

EmilLa1: Yes, as Thomas says it's unfortunate. 

We could perhaps use a docker image e.g., `envoyproxy/envoy:v1.17.0 ` if we really want to stay on the same version. But that would also be quite a change to the distribution itself.

abizjak: We can't use that directly because we make one image with all necessities in it. Changing this setup would require substantial changes in how this image is used.

Additionally, copying selected binaries from one docker image to another, with unclear and undocumented dependencies is a horrible, hacky, and fragile practice we should not do.

EmilLa1: Yep, that is what I meant with the "But that would also be quite a change to the distribution itself." comment. :-) 

And yes I agree, that we should find a proper solution to this.

abizjak: So I suppose the least shitty solution is to update the version now. I'll create an issue to fix this as well.

Could you please test this locally to make sure the entire node dashboard works as it should? Then we can merge.

EmilLa1: I did that already.

td202: > Additionally, copying selected binaries from one docker image to another, with unclear and undocumented dependencies is a horrible, hacky, and fragile practice we should not do.

I disagree with this. It's a bit hacky, but this might be the least fragile way of doing it (that doesn't involve building it from source) since it is very unlikely that a particular docker image will change or disappear.

abizjak (approved):

scripts/distribution/builder.Dockerfile
abizjak: It's not just hacky, it only works "by accident" while the shared library dependencies in those images are close enough.

If some program is built in an image with libc version Y and then copied somewhere where libc version X < Y is installed it likely won't work. 

Unless these are explicitly pinned and care is taken to either build fully static binaries (that means no libc dependencies) or use the same base image then it will break.

PR #80, opened 2021-06-08
Fix and clean up dockerized builds of node-related components

bisgardo:
Complete overhaul of the configuration, build, and documentation of the dockerized components.

Probably the biggest change is that genesis data no longer is baked into the node image. Instead, genesis has become it's own first-class component: The data is kept in a separate image that may copy the data into the expected location on startup. The updated documentation includes a full example docker-compose script on how to wire this up.


scripts/testnet-deployments/universal.Dockerfile
```diff
@@ -25,19 +25,21 @@ RUN tar -C /tmp -xf /tmp/static-consensus.tar.gz && \
...
-       concordium-node/target/debug/p2p_bootstrapper-cli \
-       concordium-node/target/debug/node-collector \
-       concordium-node/target/debug/node-collector-backend /out/debug/ && \
-    cp scripts/start.sh /out/start.sh
+    cp /build/concordium-node/target/debug/concordium-node \
+       /build/concordium-node/target/debug/p2p_bootstrapper-cli \
+       /build/concordium-node/target/debug/node-collector \
+       /build/concordium-node/target/debug/node-collector-backend \
+       /out/debug/ && \
+    cp /build/scripts/start.sh /out/start.sh
```
abizjak, line 42: And the same here.

scripts/testnet-deployments/node.Dockerfile
```diff
@@ -1,25 +1,15 @@
...
 ARG universal_image_name
 
-# Fetch genesis-data.
-FROM alpine/git:latest as genesis-data
-ARG genesis_ref
-ARG genesis_path
-RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
-RUN --mount=type=ssh git clone --depth 1 --branch "${genesis_ref}" git@gitlab.com:Concordium/genesis-data.git
-RUN mv "genesis-data/${genesis_path}" /data
-
```
abizjak, line 12: It's unclear to me what the purpose of this image is. It has an entrypoint `start.sh`, but it doesn't have any documentation on how to run it, what to set. Without genesis data it must assume that this data is mounted in some way in some place inside the image. This needs to be documented.

jenkinsfiles/master.Jenkinsfile
```diff
@@ -63,17 +67,11 @@ pipeline {
             steps {
                 sshagent (credentials: ['jenkins-gitlab-ssh']) {
                     sh '''\
-                        # Using '--no-cache' because we're cloning genesis data
-                        # and BuildKit (and '--ssh default') because the repo is on GitLab.
-                        DOCKER_BUILDKIT=1 docker build \
```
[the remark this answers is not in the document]
abizjak: Also, is buildkit and `--ssh default` still needed in light of removal of the `git clone`?

abizjak (changes requested):
There needs to be a lot more documentation on what the different images are for. It is not possible to figure out what the purpose of the slightly different images is.

I understand it is not documented as it is, but making substantial changes as you are it is a good opportunity to document it.

jenkinsfiles/master.Jenkinsfile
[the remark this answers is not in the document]
bisgardo: Resolving comment as it's outdated. I believe these flags are documented where they are used in the current revision.

scripts/testnet-deployments/universal.Dockerfile
```diff
@@ -25,19 +25,21 @@ RUN tar -C /tmp -xf /tmp/static-consensus.tar.gz && \
...
 ARG consensus_profiling=false
 ENV CONSENSUS_PROFILING="${consensus_profiling}"
-RUN ./scripts/build-binaries.sh "instrumentation,collector" "release" && \
-    ./scripts/build-binaries.sh "instrumentation,collector" && \
+RUN /build/scripts/build-binaries.sh "instrumentation,collector" "release" && \
     mkdir -p /out/release && \
+    cp /build/concordium-node/target/release/concordium-node \
+       /build/concordium-node/target/release/p2p_bootstrapper-cli \
+       /build/concordium-node/target/release/node-collector \
+       /build/concordium-node/target/release/node-collector-backend \
```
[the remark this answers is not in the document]
bisgardo: Done - note that this requires using bash explicitly.

scripts/testnet-deployments/node.Dockerfile
bisgardo: I think this comment is outdated. The current image uses the binary directly for its entrypoint.

scripts/docker/genesis.Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+#      -f genesis.Dockerfile \
+#      --ssh=default \
+#      .
+
+FROM alpine/git:latest as data
+RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+ARG tag
+ARG dir_path
+RUN --mount=type=ssh git clone --depth=1 --branch="${tag}" git@gitlab.com:Concordium/genesis-data.git /tmp/genesis-data && \
+    mv "/tmp/genesis-data/${dir_path}" /genesis-data && rm -rf /tmp/genesis-data
```
abizjak, line 17: I think this is a really bad idea. genesis data contains sensitive information and doing this makes it very easy to use the wrong path and leak testnet keys.

For testnet and mainnet we have public genesis blocks which can be used.

abizjak: I also think it is a misuse of a docker container to use it as a zip archive.

abizjak (changes requested):

scripts/docker/genesis.Dockerfile
bisgardo: This is a temporary container that exists only during build time. If you look below, only `genesis.dat` is copied into the final image. The other files never leave the build process. Even with caching, they are deleted in the same command as they are cloned and therefore don't get into the layer.

Also these images don't get published publicly, only in the internal ECR repo.

What makes it a misuse? I think it's a pretty common use case for init containers to inject data into the "main" ones.

bisgardo: Can we proceed on this?

scripts/docker/README.md
```diff
@@ -0,0 +1,86 @@
...
+networks:
+  concordium:
+```
+
+Run the script using `docker-compose`; for example:
+
+```shell
+export GENESIS_IMAGE="192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/genesis:<genesis-tag>"
+export NODE_IMAGE="192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/node:<tag>"
+export NODE_COLLECTOR_IMAGE="192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/node-collector:<tag>"
```
abizjak, line 82: I know this is an example, but I think the node and collector images should be on dockerhub, there is no reason they should be private.

scripts/docker/README.md
```diff
@@ -0,0 +1,86 @@
...
+    depends_on:
+      - genesis
+    networks:
+      - concordium
+    environment:
+      - CONCORDIUM_NODE_CONNECTION_BOOTSTRAP_NODES=bootstrap.${DOMAIN}:8888
+      - CONCORDIUM_NODE_DATA_DIR=/var/lib/concordium/data
+      - CONCORDIUM_NODE_CONFIG_DIR=/var/lib/concordium/config
+      - CONCORDIUM_NODE_PROMETHEUS_SERVER=1
+      - CONCORDIUM_NODE_RPC_SERVER_ADDR=0.0.0.0
```
abizjak, line 52: This does not explain how to configure a node as a baker.

scripts/docker/README.md
```diff
@@ -0,0 +1,86 @@
...
+  concordium:
+```
+
+Run the script using `docker-compose`; for example:
+
+```shell
+export GENESIS_IMAGE="192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/genesis:<genesis-tag>"
+export NODE_IMAGE="192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/node:<tag>"
+export NODE_COLLECTOR_IMAGE="192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/node-collector:<tag>"
+export DOMAIN=mainnet.concordium.software
```
abizjak, line 83: Seeing that mainnet is hardcoded here it should also be hardcoded for the genesis.

scripts/docker/genesis.Dockerfile
abizjak: In any case, my main objection is to the cloning of the repository to get a file out. Even though it is only during the build it is copying keys all over which is not good practice.

I'm OK if you want to have `genesis:mainnet` and `genesis:testnet` images with publicly accessible testnet and mainnet genesis blocks.

scripts/docker/README.md
abizjak: And they should be tagged with either the released version or something like dev-???

scripts/docker/genesis.Dockerfile
bisgardo: I disagree that it's error prone - there is literally no parameters you can give this job to make it leak the keys.

But what I wrote about caching is not entirely correct. In the current form, the keys could get into the cache, although only on a machine that has the credentials to get them anyway.

Anyway, after the genesis location has been parameterized, this has become more of a convenience thing that we can live without. So I will remove it. If we need it after all, the genesis-data repo may also be a better place to put it.

scripts/docker/README.md
bisgardo: Genesis image was removed.

scripts/docker/README.md
bisgardo: Can we do this separately? I would find it annoying that some of the images were public and others private. For now I'm just intending this to be fixed for internal use. We already have the public dev-images for people who want that.

scripts/docker/README.md
bisgardo: Added.

collector-backend/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+FROM concordium/base:${base_image_tag} as build
+EXPOSE 8080
+WORKDIR /build
+COPY . .
+ARG build_profile
+RUN cargo build --"${build_profile}"
+
+FROM ubuntu:20.04
+ARG build_profile
+COPY --from=build /build/target/"${build_profile}"/node-collector-backend /node-collector-backend
```
abizjak, line 11: Isn't the name of the binary just `collector-backend`? That's what's in the version I see.

collector-backend/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ENTRYPOINT ["/node-collector-backend"]
```
abizjak, line 12: Don't forget to update it here as well.

scripts/docker/README.md
```diff
@@ -0,0 +1,76 @@
...
+    environment:
+    - CONCORDIUM_NODE_CONNECTION_BOOTSTRAP_NODES=bootstrap.${DOMAIN}:8888
+    - CONCORDIUM_NODE_DATA_DIR=/mnt/data
+    - CONCORDIUM_NODE_CONFIG_DIR=/mnt/config
+    - CONCORDIUM_NODE_CONSENSUS_GENESIS_DATA_FILE=/mnt/genesis.dat
+    - CONCORDIUM_NODE_BAKER_CREDENTIALS_FILE=/mnt/baker-credentials.json
+    - CONCORDIUM_NODE_PROMETHEUS_SERVER=1
+    - CONCORDIUM_NODE_RPC_SERVER_ADDR=0.0.0.0
+    ports:
+    - "8888:8888"
```
abizjak, line 41: Does this mean that 8888 would be exposed to the host?
If so, why does it need to be? Aren't the nodes communicating on the internal concordium network?

scripts/docker/README.md
abizjak: Ok.

I think this setup needs to be rethought in any case. It is somewhat complex and without clear purpose. I'm fine if this is merged so that it at least works and is better documented than it was.

abizjak (approved):
The name of the collector-backend binary should be changed I believe, otherwise let's get this merged.

collector-backend/Dockerfile
bisgardo: No this is the name I see.

collector-backend/Dockerfile
bisgardo: Same as above. I suppose it picks the name from the filename of the source?

scripts/docker/README.md
bisgardo: This example is for a node that joins an existing network, not a cluster running on an internal docker network. The port needs to be exposed to allow incoming connections.

PR #87, opened 2021-06-17
Database Versioning

td202:
These changes are extracted from #1.

## Purpose

Add version metadata to the tree state database. The purpose of this is to support the database exporter in the context of future protocol updates. Since the database exporter does not have the context to determine the protocol version (and therefore serialization format) for a tree state database without this metadata. (Previously, it could always assume a particular version, but in future it will not be able to.)

## Changes

Global state database now includes version metadata. The treestate directory and blockstate file names are suffixed with "-0" to indicate genesis index 0, for compatibility with protocol updates.  A legacy database will automatically be migrated by renaming and adding version metadata.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/External.hs
```diff
@@ -298,6 +302,22 @@ handleStartExceptions logM c =
...
+    unless (blockStateExists || treeStateExists) $ do
+        oldBlockStateExists <- doesFileExist $ dbPath </> "blockstate" <.> "dat"
+        oldTreeStateExists <- doesDirectoryExist $ dbPath </> "treestate"
+        when (oldBlockStateExists && oldTreeStateExists) $ do
+            logM GlobalState LLInfo "Migrating global state from legacy version."
+            renameFile (dbPath </> "blockstate" <.> "dat") (dbPath </> "blockstate-0" <.> "dat")
+            renameDirectory (dbPath </> "treestate") (dbPath </> "treestate-0")
+            runLoggerT (addDatabaseVersion (dbPath </> "treestate-0")) logM
+            logM GlobalState LLInfo "Migration complete."
+
```
abizjak, line 320: Wouldn't it be cleaner if this also logged something if there is a partial database.

Additionally, we already have database versioning at the "outer layer", the database directory is passed in when the node is started, and it is currently `database-v4`. It seems confusing to have this versioning at two different layers in different naming schemes.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -405,11 +427,13 @@ startConsensusPassive ::
...
         appData <- peekCStringLen (appDataC, fromIntegral appDataLenC)
+        -- Do globalstate migration if necessary
+        migrateGlobalState appData logM
         let runtimeParams = RuntimeParameters {
               rpBlockSize = fromIntegral maxBlock,
               rpBlockTimeout = fromIntegral timeout,
-              rpTreeStateDir = appData </> "treestate",
-              rpBlockStateFile = appData </> "blockstate",
+              rpTreeStateDir = appData </> "treestate-0",
+              rpBlockStateFile = appData </> "blockstate-0",
```
abizjak, line 436: Document the choice of naming please.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -172,25 +245,28 @@ dbStepSize = 2^(26 :: Int) -- 64MB
...
   -- ^Open read only
+  -> Int
+  -- ^Initial database size
   -> IO (DatabaseHandlers pv st)
-makeDatabaseHandlers treeStateDir readOnly = do
+makeDatabaseHandlers treeStateDir readOnly initSize = do
   _storeEnv <- mdb_env_create
-  mdb_env_set_mapsize _storeEnv dbInitSize
-  mdb_env_set_maxdbs _storeEnv 4
+  mdb_env_set_mapsize _storeEnv (initSize + dbStepSize - initSize `mod` dbStepSize)
```
abizjak, line 268: Is the idea of this that the initial size is a multiple of step size?

It would be useful to be clear about why that is important, so that this is not changed in future updates.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -199,30 +275,124 @@ makeDatabaseHandlers treeStateDir readOnly = do
...
+--
+-- This is required for functionality such as the block exporter, which reads the database but does
+-- not have sufficient context to infer the protocol version.
+openReadOnlyDatabase
+  :: FilePath
+  -- ^Path of database
+  -> IO (Maybe VersionDatabaseHandlers)
+openReadOnlyDatabase treeStateDir = do
+  _storeEnv <- mdb_env_create
+  mdb_env_set_mapsize _storeEnv dbInitSize
```
abizjak, line 298: Why is the initial env size set differently for the read-only database?

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -199,30 +275,124 @@ makeDatabaseHandlers treeStateDir readOnly = do
...
+  mdb_env_set_maxdbs _storeEnv databaseCount
+  mdb_env_set_maxreaders _storeEnv 126
+  -- TODO: Consider MDB_NOLOCK
+  mdb_env_open _storeEnv treeStateDir [MDB_RDONLY]
+  (_metadataStore, mversion) <- transaction _storeEnv True $ \txn -> do
+    _metadataStore <- MetadataStore <$> mdb_dbi_open' txn (Just metadataStoreName) []
+    mversion <- loadRecord txn _metadataStore versionMetadata
+    return (_metadataStore, mversion)
```
abizjak, line 306: Why is it not necessary to handle the resizing problem in this transaction?

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -199,30 +275,124 @@ makeDatabaseHandlers treeStateDir readOnly = do
...
+            -- protocol version itself, since the database handlers are parametric in the protocol
+            -- version.
+            case promoteProtocolVersion vmProtocolVersion of
+                SomeProtocolVersion (_ :: SProtocolVersion pv) ->
+                    transaction _storeEnv True $ \txn -> do
+                        _blockStore <- BlockStore <$> mdb_dbi_open' txn (Just blockStoreName) []
+                        _finalizationRecordStore <- FinalizationRecordStore <$> mdb_dbi_open' txn (Just finalizationRecordStoreName) []
+                        _finalizedByHeightStore <- FinalizedByHeightStore <$> mdb_dbi_open' txn (Just finalizedByHeightStoreName) []
+                        _transactionStatusStore <- TransactionStatusStore <$> mdb_dbi_open' txn (Just transactionStatusStoreName) []
+                        return (Just (VersionDatabaseHandlers @pv DatabaseHandlers{..}))
```
abizjak, line 322: And here as well?

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -199,30 +275,124 @@ makeDatabaseHandlers treeStateDir readOnly = do
...
   return handlers
 
+-- |Add a database version record to a database (if one is not already present).
+-- The record indicates database version 0 and protocol version 'P1', which is appropriate when
+-- migrating a database from an earlier version.
+addDatabaseVersion :: (MonadLogger m, MonadIO m) => FilePath -> m ()
+addDatabaseVersion treeStateDir = do
+  handlers :: DatabaseHandlers 'P1 () <- liftIO $ makeDatabaseHandlers treeStateDir False dbInitSize
+  handlers' <- execStateT
+    (resizeOnFull 4096 $ \h -> transaction (_storeEnv h) False $ \txn ->
```
abizjak, line 370: Why was 4096 chosen? Just because it is the default page size? Could you clarify for future reference.

concordium-consensus/src/Concordium/External.hs
abizjak: Well it was doing versioning, but only one version was ever supported, and no migrations were done.

I think we should rename it to `database` only. I think the rust driver should control where the database location is, since some other data is also stored in the data directory, and we don't want accidental conflicts.

abizjak (approved):
I think this is OK now.

But before this and the corresponding base branch are merged we need to update the stack version wherever it is hardcoded to 2.5.1 at the moment. And announce that people should use 2.7.1 from now on, otherwise the .cabal file in base will be constantly regenerated.

concordium-consensus/src/Concordium/External.hs
abizjak: Actually forget about that. It is easier if it just stays database-v4, otherwise migration is going to be problematic.

I would prefer that versioning would be done on that level, but with the split between R and H it is too much effort.

PR #91, opened 2021-06-22
WIP: Blockstate flushing

td202:
## Purpose

Ensure that the operating system buffers are flushed when the block state is written.
Previously, the file handle was flushed with `hFlush`, but this only ensures the data is sent to the operating system; it does not ensure that the operating system has written it to disk.
This meant that the blockstate was not guaranteed to be written before the tree state that references it, leading to database corruption if the node is terminated unexpectedly.

## Changes

Introduce `hFlushOS`, which calls `FlushFileBuffers` (on Windows) or `fsync` (on Unix) to enforce flushing of the file to disk.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
@@ -53,6 +53,7 @@ import qualified Concordium.Types.IdentityProviders as IPS
 import qualified Concordium.Types.AnonymityRevokers as ARS
 import qualified Concordium.GlobalState.Parameters as Parameters
 import Concordium.GlobalState.Basic.BlockState.AccountReleaseSchedule
+import Concordium.GlobalState.Persistent.BlobStore.Flush
 import Concordium.Types
 import Concordium.Types.Updates
 import Concordium.Wasm
@@ -114,7 +115,7 @@ loadBlobStore blobStoreFilePath = do
 -- ensuring all the contents is written out.
 flushBlobStore :: BlobStore -> IO ()
 flushBlobStore BlobStore{..} =
-    withMVar blobStoreFile (hFlush . bhHandle)
+    withMVar blobStoreFile (hFlushOS . bhHandle)
 
 -- |Close all references to the blob store, flushing it
 -- in the process.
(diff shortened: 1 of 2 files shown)


EmilLa1:
@td202 It seems that the expected behaviour is not the same for Linux as it is on MacOS. 

As per documentation: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html 

I think on MacOS we should use `F_FULLFSYNC fcntl`.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore/Flush.hsc
```diff
@@ -0,0 +1,83 @@
...
+    res <- fcntl (fdFD fd) (#const F_FULLFSYNC)
+    when (res == -1) $ throwErrno "hFlushOS"
+#else
+    -- Linux/other
+    res <- fdatasync (fdFD fd)
+    when (res /= 0) $ throwErrno "hFlushOS"
+#endif
+
+-- |Flush a file handle, including the operating system buffers.
+hFlushOS :: Handle -> IO ()
```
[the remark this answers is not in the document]
td202: Renamed to `hFlushToDisk`.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -312,6 +312,8 @@ startConsensus ::
     Word64 ->
     -- |Number of seconds between transaction table purging runs
     Word64 ->
+    -- |Block state flush mode
+    Word64 ->
```
abizjak, line 316: It would be useful to document what the meaning of the number is, e.g., 0 for *, 1 for *.

concordium-consensus/src/Concordium/GlobalState/Parameters.hs
```diff
@@ -69,6 +68,19 @@ getExactVersionedCryptographicParameters bs = do
...
+-- |Mode to use for flushing the blob store buffer.
+-- Flushing to disk is more resilient, as it ensures that the data is written when a flush occurs.
+-- However, it does increase the overhead of the flush operation.
+data FlushMode
+    = FlushToOS
+    -- ^Flush the buffer to the operating system, but don't force a disk flush.
+    | FlushToDisk
+    -- ^Flush the buffer to the operating system and subsequently to disk.
+
+-- |Decode a 'FlushMode' from a 'Word64'.
```
abizjak, line 80: Document intended semantics please.

PR #94, opened 2021-06-23
Refactor types to base

abhconcordium:
## Purpose
Updated `concordium-node` to parity with `concordium-base` https://github.com/Concordium/concordium-base/pull/47

Many type definitions which are used across multiple repos have been transferred to `concordium-base` from `concordium-node`. This branch removes those definitions from node, and includes the imports/refactoring required to compile using the new version of `concordium-base`.

## Changes

Multiple type definitions moved to `concordium-base`.

The JSON serialization used in the response of the `GetRewardStatus` GRPC call has been changed.
Amounts that were previously represented as integers are now represented as strings in the JSON serialization.
This is in line with how amounts are serialized elsewhere.
This is a breaking change for the tokenomics reporter.

## Checklist

- [ ] Manual testing of affected endpoints


abhconcordium:
@td202 
Only change I made from your original types file was reverting the 
`data BakerStatus` back to one where they didn't take arguments, other than that types should be compatible with your CB-1082-protocol-update branch.

This compiles with the correct concordium-base without warnings or errors. 
Any thoughts on other types which would also make sense to bring over to base?

td202:
> 
> 
> @td202
> Only change I made from your original types file was reverting the
> `data BakerStatus` back to one where they didn't take arguments, other than that types should be compatible with your CB-1082-protocol-update branch.
> 
> This compiles with the correct concordium-base without warnings or errors.
> Any thoughts on other types which would also make sense to bring over to base?

I am moving `BakerStatus` back out of concordium-base, because this type is not part of the GRPC interface. It is only used between `Getters` and `External`, where it is used to construct an `Int64` that's passed to Rust.

concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/AccountReleaseSchedule.hs
```diff
@@ -34,113 +35,130 @@ module Concordium.GlobalState.Basic.BlockState.AccountReleaseSchedule (
...
+                    )
+                )
+        let (_pendingReleases, _totalLockedUpBalance) =
+                Vector.ifoldl'
+                    ( \acc idx -> \case
+                        Nothing -> acc
+                        Just (rel, _) ->
+                            let f (pending, am) Release{..} =
+                                    let pending' = Map.alter (maybe (Just [idx]) (Just . (idx :))) timestamp pending
+                                        am' = am + amount
```
abizjak, line 144: I'm guessing only indendation/line breaks changed, but this format style seems substantially worse than the previous one.

abizjak (approved):

concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/AccountReleaseSchedule.hs
td202: I don't love everything about the fourmolu formatting style, but I don't agree that it's substantially worse. I think it actually more clearly delineates the arguments of multi-line function calls such as `replicateM` and `ifoldl'` here.

td202 (approved):

PR #95, opened 2021-07-05
Update the rust version to 1.53.

### LINKED ISSUE #43: Update flatbuffers compiler and dependency

abizjak:
Also update dependencies as appropriate, and handle new clippy hints.

## Purpose

Update the minimum supported rust version and dependencies.

## Changes

- Update dependencies enabled by the update of the rust version.
- Update the rust version in CI
- Address clippy suggestions.
- Update the required flatc version to 2.0.

Closes #43 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/README.md
+++ b/concordium-node/README.md
@@ -1,9 +1,9 @@
 # Concordium node implementation
 ## Dependencies to build the project
-* Rust (stable 1.45.2 for using static libraries)
+* Rust (stable 1.53 for using static libraries)
 * binutils >= 2.22
 * cmake >= 3.8.0
-* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html) v1.11 and v1.12 are known to work. (build using CMake and copy to `~/.local/bin`)
+* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html) v2.0 is known to work. (build using CMake and copy to `~/.local/bin`)
 * protobuf >= 3.7.1
 * LLVM and Clang >= 3.9
 * [Unbound](https://www.nlnetlabs.nl/documentation/unbound/howto-setup/) >= 1.9.2 (the dependency `openssl-devel` is named `libssl-dev` on Ubuntu 19.10)
--- a/concordium-node/README.md
+++ b/concordium-node/README.md
@@ -3,7 +3,8 @@
 * Rust (stable 1.53 for using static libraries)
 * binutils >= 2.22
 * cmake >= 3.8.0
-* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html) v2.0 is known to work. (build using CMake and copy to `~/.local/bin`)
+* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html)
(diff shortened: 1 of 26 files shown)


concordium-node/src/utils.rs
```diff
@@ -2,26 +2,20 @@
```
EmilLa1, line 2: Maybe we should move `generate_bootstrap_dns` (and its helper functions) to `concordium-node/src/bin/dnsrecordgen.rs` as that is the only place it's being used.

concordium-node/src/utils.rs
```diff
@@ -296,194 +278,6 @@ pub fn generate_bootstrap_dns(
...
-                }
-            }
-            Err(_) => Err("Invalid data"),
-        }
-    } else {
-        Err("Missing bytes")
-    }
-}
-
 pub fn generate_ed25519_key() -> SecretKey { SecretKey::generate(&mut OsRng::default()) }
```
EmilLa1, line 281: This could also be moved to `concordium-node/src/bin/genkey.rs` as it is the only place being used.

EmilLa1 (commented):
Good job. 

I think we should take care not to have _too many_ files like `concordium-node/src/utils.rs` where it's not clear from the context what the purpose of the file and the logic (functions, structs etc.) inside are. 

Also having these files where the containing logic is used from different modules can lead to _unclean_ module apis.

EmilLa1 (approved):

concordium-node/src/utils.rs
abizjak: I am strongly against "moving things around on the fly" as part of PRs whose purpose is different.

Or in general moving things around unless there is a clear reason for it. It just destroys everybody's mental image of where things are.

There should at some point be some analysis and revision/cleanup of the bootstrapping code, DNS resolution. Things can be moved around then.

concordium-node/README.md
```diff
@@ -1,9 +1,9 @@
 # Concordium node implementation
 ## Dependencies to build the project
-* Rust (stable 1.45.2 for using static libraries)
+* Rust (stable 1.53 for using static libraries)
 * binutils >= 2.22
 * cmake >= 3.8.0
-* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html) v1.11 and v1.12 are known to work. (build using CMake and copy to `~/.local/bin`)
+* [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html) v2.0 is known to work. (build using CMake and copy to `~/.local/bin`)
```
td202, line 6: It might be useful to link to the release page for 2.0.0, where they have binaries available: https://github.com/google/flatbuffers/releases/tag/v2.0.0
(I don't know about this situation on other platforms, but msys2 and chocolatey don't currently have this version available.)

td202 (approved):
A couple of suggested changes, but otherwise seems fine.

concordium-node/README.md
abizjak: Added.

concordium-node/src/utils.rs
abizjak: I do not wish to move around things that are not related to the PR.

PR #98, opened 2021-07-09
Documentation of the network layer.

EmilLa1:
## Purpose

Documenting the network layer. 

## Changes

Added a document describing the network layer.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/docs/network-layer.md
+++ b/docs/network-layer.md
@@ -1,0 +1,71 @@
+# Network Layer
+This document describes the network layer of the Concordium blockchain.
+
+## Overview 
+The network layer consists of a peer-to-peer (P2P) network. Each node in the network is connected to a *set of peers*. The node uses the established connections for various tasks including: *retrieving more peers*, _consensus_, _catching up_ on the chain etc. There exist two *types* of peers in the network, namely the *Node* and the *Bootstrapper* types. The *Node* carries out the tasks mentioned above while the purpose of the latter's task is to solely serve peers.
+
+The peers are orchestrated in a random graph and communicates using a [*gossip-like*](https://en.wikipedia.org/wiki/Gossip_protocol) protocol.
+
+Peers can take part in more than one network at the same time. A *network* groups peers together, thus the node can request to *join* a network, leave a *network* and also request more peers of a certain *network*. Currently only a single *network* is being used.
+
+## Joining a Network
+When a node wishes to join a network it starts out by contacting a *bootstrapper*. The *bootstrapper* is a special variant of the *concordium-node* which relays peers (only the ones matching the requested `NetworkId`) but disregards all other types of messages.
+
+The *bootstrapper* sends out a *randomized* list of peers to the requesting node. The node who wishes to join a network will in turn try to connect to these *candidates* received from the *bootstrapper*. 
+The *bootstrapper* will only keep peers on its list if the *bootstrapper* and the peer have successfully established a connection (see below).
+
+The node will also send a *GetPeers* message to its connected peers if it still *lacks* peers. In return a peer will send a list of *candidates* to the node, to which the node can try establishing connections. 
+
+It is also possible for a node to be *pinned* to certain other peers. This is a task that is carried out by the node operator when configuring the node. The node will always try to keep a connection to each of the *pinned* peers.
+
+The node operator can via the RPC interface make the node *join* or *leave* a given *network* (specified by the `NetworkId`).
+
+## Connection Establishment
+When a node and a peer try to establish a connection, the node and the other peer will perform a 2-part handshake.
+
+The purpose of the *first part* of the handshake is to create a *secure channel* between the two peers. In practice a [*Noise XX*](https://noiseexplorer.com/patterns/XX/) handshake is being carried out by the two peers. This ensures *authenticity* and *confidentiality* for the rest of the communication. Note that currently none of the data sent between the peers is in fact secret and thus it is not a requirement from a security point-of-view that the *secure channel* is in fact being encrypted.
+
+After completing the first part of the handshake, then the two peers will carry out the second part of the handshake. By checking various metadata, the two peers confirm that they are in fact compatible. Specifically the peers check that:
+
+1. The client versions are compatible
+2. The wire versions are compatible
+3. The peers have *compatible* genesis blocks. *Compatible* in this context means that one peer's list of *genesis blocks* must be a prefix of the other's.
+
+When the *handshake* has been successfully completed then the peers will *promote* each other into their corresponding lists of peers, i.e. the peers are now connected and can communicate with each other, hence the peers can now relay messages between them, carry out *consensus* and *catchup* etc.
+
+## Control Flow
+When the node is running, it continously checks for incoming connections and polls for events related to the already established connections. Furthermore the node performs several housekeeping tasks wrt. networking related matters. More specifically the node checks that its current connections are *healthy* i.e., verifying reasonable latency, removing connections of which a handshake is missing etc. If any connections are deemed *unhealthy* then the node will close the connection and remove the peer from its set of peers. Note that depending on the level of *unhealthy* the node will perhaps *soft ban* a peer (more on this below).
+
+
+If the node has exceeded its maximum amount of peers it will remove a random subset of its peers. Note that the node will never remove any of the *pinned peers*. 
+The node then proceeds to check if it should unban any of the *soft banned* peers and does so if that is the case. Note that *soft bans* have a duration of 5 minutes.
+
+Afterwards the node may try to *re-bootstrap* again (if the node did not recently *bootstrap*). Finally if the node still lacks peers then it will send out *GetPeers* messages to its peers.
+
+
+The node listens on its established connections and processes the received messages accordingly. If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. Before processing the message the node checks that it has not yet seen the message before or otherwise the node will not process the *duplicated* message. 
+
+The control flow of the *network layer* can be summarized as follows:
+
+1. Listen for incoming connections
+2. Process network messages
+3. (Optionally) Carry out housekeeping tasks if a suffcient time period has passed since last time a housekeeping was done.
+4. (Optionally) Establish new outgoing connections if required
+5. Repeat
+
+## Processing Network Message
+When the node receives a network message on one of its connections it will either dispatch the network message to the *consensus layer* or immediately handle the network message if it was only *network related*. An example of the latter could be that the node was asked for peers by the *GetPeers* message, responding to a *Ping*, *Join*-/*Leave* network messages etc. 
+However, when the message is for the *consensus layer*, it is first checked that the message was not a *duplicate message* i.e., a message that the node has received before (more on this below). If the message was a *duplicate*, then node will not process the *duplicated* message. 
+The node dispatches *new* messages to a *low-priority queue* or a *high-priority queue* depending on the content of the network message. Messages containing *blocks*, *finalization records*, *finalization messages*, *catch-up status'* are treated as high priority, while *transactions* are treated as low priority.
+If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. 
+
+The node holds *de-duplication* buffers for each of the *rebroadcastable* network packet types; finalization-, transaction-, block-, and finalization record packets. These *de-duplication* buffers contain hashes of previously received messages. Before the node *handles* the actual message (i.e. passing the message through to the *consensus* layer) it is checked that the message is not a duplicate (by computing the hash of the message and checking whether it was contained within the corresponding de-duplication buffer). If the message was a duplicate then the node will simply disregard the duplicated message. 
+
+## Banning
+A node holds information if another peer is *permanently banned* or *soft banned* (e.g. a temporary ban could have a duration of 24 hours). 
+Permanent banning is carried out by a manual process, thus the operator of the node can put other peers on the permanent ban list by calling the RPC API. Banned peers are identified by their IP address and so the node will not establish connections to such an address.
+
+*Soft banning* is on the other hand an automatic procedure carried out by the node. The node *soft bans* a peer based on the outcome from the *consensus* layer. Reasons include: the peer sends malformed messages; the peer sends messages with *bad* data e.g. an inconsistent *catch-up* message; etc. *Soft bans* are also given by the IP address of the peer and as such the node will immediately disconnect from the peer and avoid establishing new connections to the peer within the *soft ban duration*.
+
+## Connecting to the *Mainnet*
+In order to connect to the Concordium *Mainnet* the node should be configurged with the Concordium *bootstrappers*. Currently there are two *bootstrappers* for the Concordium *Mainnet* and they can be reached at `bootstrap.mainnet.concordium.software`. Likewise there are two *bootstrapppers* for the Concordium *Testnet* and they are located at `bootstrap.testnet.concordium.com`.


docs/network-layer.md
```diff
@@ -0,0 +1,61 @@
...
+
+If the node has exceeded its maximum amount of peers it will remove a random subset of its peers. Note that the node will never remove any of the *pinned peers*. 
+The node then proceeds to check if it should unban any of the *soft banned* peers and does so if that is the case. 
+
+Afterwards the node may try to *re-bootstrap* again (if the node did not recently *bootstrap*). Finally if the node still lacks peers then it will send out *GetPeers* messages to its peers.
+
+The node listens on its established connections and processes the received messages accordingly. If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. Beforehand processing the message the node asserts that it has not yet seen the message before or otherwise the node will not process the *duplicated* message.
```
abizjak, line 44:
```suggestion
The node listens on its established connections and processes the received messages accordingly. If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. Before processing the message the node asserts that it has not yet seen the message before or otherwise the node will not process the *duplicated* message. 
```

abizjak: Perhaps a paragraph about how deduplication works would be useful here.

docs/network-layer.md
```diff
@@ -0,0 +1,61 @@
...
+The control flow of the *network layer* can be summarized as follows:
+
+1. Listen for incoming connections
+2. Process network messages
+3. Housekeeping
+4. (Optionally) Establish new connections
+5. Repeat
+
+## Banning
+A node holds information if another peer is *permanent banned* or *soft banned* (e.g. a temporary ban could have a duration of 24 hours).
```
abizjak, line 55:
```suggestion
A node holds information if another peer is *permanently banned* or *soft banned* (e.g. a temporary ban could have a duration of 24 hours). 
```

docs/network-layer.md
```diff
@@ -0,0 +1,61 @@
...
+Permanent banning is carried out by a manual process, thus the operator of the node can put other peers on the permanent ban list by calling the RPC API. Banned peers are identified by their IP address and so the node will not establish connections to such an address.
+
+*Soft banning* is on the other hand an automatic procedure carried out by the node. The node *soft bans* a peer based on the outcome from the *consensus* layer, reasons include; the peer sends malformed messages, the peer sends messages with *bad* data e.g. an inconsistent *catch-up* message etc. *Soft bans* are also given by the IP address of the peer and as such the node will immediately disconnect from the peer and avoid establishing new connections to the peer within the *soft ban duration*.
+
+## Connecting to the *Mainnet*
+In order to connect to the Concordium *Mainnet* the node should be configurged with the Concordium *bootstrappers*. Currently there exists 2 *bootstrappers* for the Concordium *Mainnet* and they can be reached at `bootstrap.mainnet.concordium.software`.
```
abizjak, line 61: If the mainnet ones are here, the testnet ones should also be `bootstrap.testnet.concordium.com`.

abizjak (changes requested):
I think this is OK, but I am missing the flow of messages.

Specifically it would be good to describe the (conceptual) steps a protocol message (e.g., block, transaction, ...) goes through when received.

docs/network-layer.md
EmilLa1: Will elaborate on this.

EmilLa1:
Added the flow of messages and other changes wrt. review comments.

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
+# Network Layer
+This document describes the network layer of the Concordium blockchain.
+
+## Overview 
+The network layer consists of a peer-to-peer (P2P) network. Each node in the network is connected to a *set of peers*. The node uses the established connections for various tasks including: *retrieving more peers*, _consensus_, _catching up_ on the chain etc.
```
abizjak, line 5: Maybe also add that there are two kinds of peers here, the bootstrappers and "node" (or however they are called in the code). 
And bootstrappers don't run consensus, they just serve peers. Bootstrappers are mentioned below, but they could be in this overview as well.

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Control Flow
+When the node is running, it continously checks for incoming connections and polls for events related to the already established connections. Furthermore the node performs several housekeeping tasks wrt. networking related matters. More specifically the node checks that its current connections are *healthy* i.e., verifying reasonable latency, removing connections of which a handshake is missing etc. If any connections are deemed *unhealthy* then the node will close the connection and remove the peer from its set of peers. Note that depending on the level of *unhealthy* the node will perhaps *soft ban* a peer (more on this below).
+
+
+If the node has exceeded its maximum amount of peers it will remove a random subset of its peers. Note that the node will never remove any of the *pinned peers*. 
+The node then proceeds to check if it should unban any of the *soft banned* peers and does so if that is the case.
```
abizjak, line 41: Perhaps rephrase this to clarify when this unbanning happens (the 5min window).

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. Before processing the message the node asserts that it has not yet seen the message before or otherwise the node will not process the *duplicated* message. 
+The node holds *de-duplication* buffers each of the *rebroadcastable* network packet types; finalization-, transaction-, block-, and finalization records packets. These *de-duplication* buffers contains hashes of previous received messages. Before the node *handles* the actual message (I.e. passing the message through to the *consensus* layer) then it's asserted that the message is  not a duplicate (by computing the hash of the message and checks whether it was contained within the corresponding de-duplication buffer), hence if the message was a duplicate then the node will simply disregard the duplicated message.
```
abizjak, line 61:
```suggestion
The node holds *de-duplication* buffers each of the *rebroadcastable* network packet types; finalization-, transaction-, block-, and finalization records packets. These *de-duplication* buffers contains hashes of previous received messages. Before the node *handles* the actual message (i.e. passing the message through to the *consensus* layer) then it's asserted that the message is  not a duplicate (by computing the hash of the message and checks whether it was contained within the corresponding de-duplication buffer), hence if the message was a duplicate then the node will simply disregard the duplicated message. 
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+The control flow of the *network layer* can be summarized as follows:
+
+1. Listen for incoming connections
+2. Process network messages
+3. (Optionally) Carry out housekeeping tasks if a suffcient time period has passed since last time a housekeeping was done.
+4. (Optionally) Establish new outgoing connections if required
+5. Repeat
+
+## Processing Network Messages
+At the time of the connection establishment (*pre-handshake*) the connection is being registered for emitting *events*, hence a network message has been written to the underlying TCP socket. When such an event occurs, then the node will either dispatch the networking message to the *consensus layer* or immediately handle the message if it was not. And example of the former could be that the node was requested for peers by the *GetPeers* message, responding to a *Ping* etc.
```
abizjak, line 57: I think the first sentence is confusing.  It has both too many, and too little detail to be understandable.

The listening for events part is a low-level detail so that we only read from connections that have something to read from.

I think that detail should be separated from the more high-level processing for messages which happens after a message is read. This part would be the same regardless of how the low-level socket reads happen.

abizjak (approved):

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+The peers are orchestrated in a random graph and communicates using a [*gossip-like*](https://en.wikipedia.org/wiki/Gossip_protocol) protocol.
+
+Peers can take part in more than one network at the same time. A *network* groups peers together, thus the node can request to *join* a network, leave a *network* and also request more peers of a certain *network*. Currently only a single *network* is being used.
+
+## Joining a Network
+When a node wishes to join a network it starts out by contacting a *bootstrapper*. The *bootstrapper* is a special variant of the *concordium-node* which relays peers (only the ones matching the requested `NetworkId`) but disregards all other types of messages.
+
+The *bootstrapper* sends out a *randomized* list of peers to the requesting node. The node who wishes to join a network will in turn try to connect to these *candidates* received from the *bootstrapper*. 
+The *bootstrapper* will only keep peers on its list if the *bootstrapper* and the peer has sucesfully established a connection (see below).
```
td202, line 15:
```suggestion
The *bootstrapper* will only keep peers on its list if the *bootstrapper* and the peer have successfully established a connection (see below).
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
+# Network Layer
+This document describes the network layer of the Concordium blockchain.
+
+## Overview 
+The network layer consists of a peer-to-peer (P2P) network. Each node in the network is connected to a *set of peers*. The node uses the established connections for various tasks including: *retrieving more peers*, _consensus_, _catching up_ on the chain etc. There exists two *types* of peers in the network, namely the *Node* and the *Bootstrapper* types. The *Node* carries out the tasks mentioned above while the purpose of the latter's task is to solely serve peers.
```
td202, line 5:
```suggestion
The network layer consists of a peer-to-peer (P2P) network. Each node in the network is connected to a *set of peers*. The node uses the established connections for various tasks including: *retrieving more peers*, _consensus_, _catching up_ on the chain etc. There exist two *types* of peers in the network, namely the *Node* and the *Bootstrapper* types. The *Node* carries out the tasks mentioned above while the purpose of the latter's task is to solely serve peers.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+Peers can take part in more than one network at the same time. A *network* groups peers together, thus the node can request to *join* a network, leave a *network* and also request more peers of a certain *network*. Currently only a single *network* is being used.
+
+## Joining a Network
+When a node wishes to join a network it starts out by contacting a *bootstrapper*. The *bootstrapper* is a special variant of the *concordium-node* which relays peers (only the ones matching the requested `NetworkId`) but disregards all other types of messages.
+
+The *bootstrapper* sends out a *randomized* list of peers to the requesting node. The node who wishes to join a network will in turn try to connect to these *candidates* received from the *bootstrapper*. 
+The *bootstrapper* will only keep peers on its list if the *bootstrapper* and the peer has sucesfully established a connection (see below).
+
+The node will also send a *GetPeers* message to its connected peers if it still *lacks* peers. In return a peer will send a list of *candidates* to the node, of which the node can try establishing a connection to.
```
td202, line 17:
```suggestion
The node will also send a *GetPeers* message to its connected peers if it still *lacks* peers. In return a peer will send a list of *candidates* to the node, to which the node can try establishing connections. 
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Joining a Network
+When a node wishes to join a network it starts out by contacting a *bootstrapper*. The *bootstrapper* is a special variant of the *concordium-node* which relays peers (only the ones matching the requested `NetworkId`) but disregards all other types of messages.
+
+The *bootstrapper* sends out a *randomized* list of peers to the requesting node. The node who wishes to join a network will in turn try to connect to these *candidates* received from the *bootstrapper*. 
+The *bootstrapper* will only keep peers on its list if the *bootstrapper* and the peer has sucesfully established a connection (see below).
+
+The node will also send a *GetPeers* message to its connected peers if it still *lacks* peers. In return a peer will send a list of *candidates* to the node, of which the node can try establishing a connection to. 
+
+It is also possible for a node to be *pinned* to certain other peers. This is a task that is carried out by the node operator when configuring the node. The node will always try to keep a connection the the *pinned* peers.
```
td202, line 19:
```suggestion
It is also possible for a node to be *pinned* to certain other peers. This is a task that is carried out by the node operator when configuring the node. The node will always try to keep a connection to each of the *pinned* peers.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+The node operator can via the RPC interface make the node *join* or *leave* a given *network* (specified by the `NetworkId`).
+
+## Connection Establishment
+When a node and a peer try to establish a connection, the node and the other peer will perform a 2-part handshake.
+
+The purpose of the *first part* of the handshake is to create a *secure channel* between the two peers. In practice a [*Noise XX*](https://noiseexplorer.com/patterns/XX/) handshake is being carried out by the two peers. This ensures *authenticity* and *confidentiality* for the rest of the communication. Note that currently none of the data sent between the peers is in fact secret and thus it is not a requirement from a security point-of-view that the *secure channel* is in fact being encrypted.
+
+After completing the first part of the handshake, then the two peers will carry out the second part of the handshake. By checking various meta data, the two peers confirm that they are in fact compatible. Specifically the peers asserts that:
```
td202, line 28:
```suggestion
After completing the first part of the handshake, then the two peers will carry out the second part of the handshake. By checking various metadata, the two peers confirm that they are in fact compatible. Specifically the peers assert that:
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. The client versions are compatible
+2. The wire versions are compatible
+3. The peers have *compatible* genesis blocks. *Compatible* in this context means that one of the other must provide a set of *genesis blocks* which should be a subset of others.
```
td202, line 32:
```suggestion
3. The peers have *compatible* genesis blocks. *Compatible* in this context means that one peer's list of *genesis blocks* must be a prefix of the other's.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+When the *handshakes* has been successfully completed then the peers will *promote* each other into their corresponding lists of peers, i.e. the peers are now connected and can communicate with each other, hence the peers can now relay messages between them, carry out *consensus* and *catchup* etc.
```
td202, line 34:
```suggestion
When the *handshake* has been successfully completed then the peers will *promote* each other into their corresponding lists of peers, i.e. the peers are now connected and can communicate with each other, hence the peers can now relay messages between them, carry out *consensus* and *catchup* etc.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Control Flow
+When the node is running, it continously checks for incoming connections and polls for events related to the already established connections. Furthermore the node performs several housekeeping tasks wrt. networking related matters. More specifically the node checks that its current connections are *healthy* i.e., verifying reasonable latency, removing connections of which a handshake is missing etc. If any connections are deemed *unhealthy* then the node will close the connection and remove the peer from its set of peers. Note that depending on the level of *unhealthy* the node will perhaps *soft ban* a peer (more on this below).
+
+
+If the node has exceeded its maximum amount of peers it will remove a random subset of its peers. Note that the node will never remove any of the *pinned peers*. 
+The node then proceeds to check if it should unban any of the *soft banned* peers and does so if that is the case. Note that *soft bans* has a duration of 5 minutes.
```
td202, line 41:
```suggestion
The node then proceeds to check if it should unban any of the *soft banned* peers and does so if that is the case. Note that *soft bans* have a duration of 5 minutes.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+The control flow of the *network layer* can be summarized as follows:
+
+1. Listen for incoming connections
+2. Process network messages
+3. (Optionally) Carry out housekeeping tasks if a suffcient time period has passed since last time a housekeeping was done.
+4. (Optionally) Establish new outgoing connections if required
+5. Repeat
+
+## Processing Network Message
+When the node receives a network message on one of its connections it will either dispatch the network message to the *consensus layer* or immediately handle the network message if it was only *network related*. An example of the former could be that the node was requested for peers by the *GetPeers* message, responding to a *Ping*, *Join*-/*Leave* network messages etc.
```
td202, line 57:
```suggestion
When the node receives a network message on one of its connections it will either dispatch the network message to the *consensus layer* or immediately handle the network message if it was only *network related*. An example of the latter could be that the node was asked for peers by the *GetPeers* message, responding to a *Ping*, *Join*-/*Leave* network messages etc. 
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. Listen for incoming connections
+2. Process network messages
+3. (Optionally) Carry out housekeeping tasks if a suffcient time period has passed since last time a housekeeping was done.
+4. (Optionally) Establish new outgoing connections if required
+5. Repeat
+
+## Processing Network Message
+When the node receives a network message on one of its connections it will either dispatch the network message to the *consensus layer* or immediately handle the network message if it was only *network related*. An example of the former could be that the node was requested for peers by the *GetPeers* message, responding to a *Ping*, *Join*-/*Leave* network messages etc. 
+However in the first case, the network message is either dispatched to a *low priority queue* or a *high priority queue* depending on the content of the network message. Messages containing *blocks*, *finalization records*, *finalization messages*, *catch-up status'* will be prioritized in favor of simple *transactions*.
```
td202, line 58:
```suggestion
However, when the message is for the consensus layer, it is either dispatched to a *low-priority queue* or a *high-priority queue* depending on the content of the network message. Messages containing *blocks*, *finalization records*, *finalization messages*, *catch-up status'* are treated as high priority, while *transactions* are treated as low priority.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. Before processing the message the node asserts that it has not yet seen the message before or otherwise the node will not process the *duplicated* message.
```
td202, line 60: Perhaps reorganise this a bit to mention deduplication before the message processing is discussed so that the order of operations is more apparent.

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+If the message was *re-broadcastable*, then the node will relay the message to its own peers such that the message will propagate throughout the network. Before processing the message the node asserts that it has not yet seen the message before or otherwise the node will not process the *duplicated* message. 
+The node holds *de-duplication* buffers each of the *rebroadcastable* network packet types; finalization-, transaction-, block-, and finalization records packets. These *de-duplication* buffers contains hashes of previous received messages. Before the node *handles* the actual message (i.e. passing the message through to the *consensus* layer) then it's asserted that the message is  not a duplicate (by computing the hash of the message and checks whether it was contained within the corresponding de-duplication buffer), hence if the message was a duplicate then the node will simply disregard the duplicated message.
```
td202, line 61:
```suggestion
The node holds *de-duplication* buffers for each of the *rebroadcastable* network packet types; finalization-, transaction-, block-, and finalization record packets. These *de-duplication* buffers contain hashes of previously received messages. Before the node *handles* the actual message (i.e. passing the message through to the *consensus* layer) it is checked that the message is not a duplicate (by computing the hash of the message and checking whether it was contained within the corresponding de-duplication buffer). If the message was a duplicate then the node will simply disregard the duplicated message. 
```
In some other places where you use "assert", I think "check" would be more appropriate.

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Banning
+A node holds information if another peer is *permanently banned* or *soft banned* (e.g. a temporary ban could have a duration of 24 hours). 
+Permanent banning is carried out by a manual process, thus the operator of the node can put other peers on the permanent ban list by calling the RPC API. Banned peers are identified by their IP address and so the node will not establish connections to such an address.
+
+*Soft banning* is on the other hand an automatic procedure carried out by the node. The node *soft bans* a peer based on the outcome from the *consensus* layer, reasons include; the peer sends malformed messages, the peer sends messages with *bad* data e.g. an inconsistent *catch-up* message etc. *Soft bans* are also given by the IP address of the peer and as such the node will immediately disconnect from the peer and avoid establishing new connections to the peer within the *soft ban duration*.
```
td202, line 67:
```suggestion
*Soft banning* is on the other hand an automatic procedure carried out by the node. The node *soft bans* a peer based on the outcome from the *consensus* layer. Reasons include: the peer sends malformed messages; the peer sends messages with *bad* data e.g. an inconsistent *catch-up* message; etc. *Soft bans* are also given by the IP address of the peer and as such the node will immediately disconnect from the peer and avoid establishing new connections to the peer within the *soft ban duration*.
```

docs/network-layer.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Connecting to the *Mainnet*
+In order to connect to the Concordium *Mainnet* the node should be configurged with the Concordium *bootstrappers*. Currently there exists 2 *bootstrappers* for the Concordium *Mainnet* and they can be reached at `bootstrap.mainnet.concordium.software`. Likewise there exists 2 *bootstrapppers* for the Concordium *Testnet* and they are located at `bootstrap.testnet.concordium.com`.
```
td202, line 70:
```suggestion
In order to connect to the Concordium *Mainnet* the node should be configurged with the Concordium *bootstrappers*. Currently there are two *bootstrappers* for the Concordium *Mainnet* and they can be reached at `bootstrap.mainnet.concordium.software`. Likewise there are two *bootstrapppers* for the Concordium *Testnet* and they are located at `bootstrap.testnet.concordium.com`.
```

td202 (approved):
Some minor fixes, but otherwise fine.

PR #112, opened 2021-07-23
Reduce lock contention for queries.

abizjak:
In the current setup there is a global lock on database operations because all the operations are through a single file handle. This is somewhat of a necessity since Haskell mandates that a file can be opened either in read mode by many, or as a single instance in write (or readwrite) mode. However all modern operating systems allow opening files by multiple readers and writers by one or multiple processes.

On linux there is the call [pread](https://man7.org/linux/man-pages/man2/pwrite.2.html) which allows to read from a file descriptor without modifying it. This allows concurrent file access, the synchronization being handled by the operating system. This avoids constant seeks necessary in the current setup.

These changes (see timings below) make the queries now CPU bound. The reason for high CPU usage on the node side is that a lot of values are quite expensive to deserialize. The main culprits are various keys since we check during deserialization that they are valid. This happens always, even if we already know they are valid because we store them in the node's database. Changing this behaviour so that we could have "trusted deserialization" methods which would check less is non-trivial though and might require changing the serialization format to explicitly record length. But it will speed up queries a lot if we do not have to go through the step fully deserializing and promptly JSON encoding the value we retrieve.

Times to query all accounts (there are 418) in block
ebef3627979fc2ec1d5e182265378639988e93d3fb309df41e2e223cc6fd39f0 on the testnet
with the old node running with --haskell-rts-flags=-N4.

The first column indicates how many parallel requests were made. As is clear
from the numbers, most of the time that the client has spent is waiting for a
response (the user time is a small percentage of execution time).
```
1  0.11s user 0.04s system 8% cpu 1.715 total
2  0.05s user 0.06s system 6% cpu 1.703 total
3  0.12s user 0.02s system 5% cpu 2.532 total
4  0.13s user 0.08s system 5% cpu 3.913 total
5  0.13s user 0.08s system 5% cpu 4.162 total
6  0.13s user 0.08s system 5% cpu 3.917 total
7  0.14s user 0.07s system 4% cpu 4.246 total
8  0.15s user 0.05s system 4% cpu 4.128 total
9  0.15s user 0.06s system 4% cpu 4.700 total
10 0.13s user 0.08s system 4% cpu 4.353 total
11 0.14s user 0.09s system 4% cpu 4.909 total
12 0.08s user 0.09s system 4% cpu 3.771 total
13 0.08s user 0.12s system 4% cpu 4.098 total
14 0.13s user 0.09s system 4% cpu 4.980 total
15 0.10s user 0.10s system 4% cpu 4.176 total
[...]

PR #113, opened 2021-07-25
WIP: Blobstore mmap

td202:
## Purpose

It has long been proposed to use a memory-mapped file for the block state.
This PR does this in a fairly simple, but hopefully effective, fashion.
While writes still acquire a lock and append to the file handle, reads use a read-only memory map of the file.
If the read is for some data that is beyond the size of the memory map, the file is re-mapped.
If the data is *still* beyond the end of the file, then reading from the handle is used as a fall-back.

This assumes that generally data is not partially written to the blob store, however, it should be robust even if it is by falling back to using the file handle.
Falling back to the file handle may also be necessary if the data is ever read before it is flushed (which *probably* doesn't happen).

This should also reduce lock contention similar to https://github.com/Concordium/concordium-node/pull/112.

Testing is necessary to ensure that performance doesn't degrade over many queries and remappings, and that behaviour is consistent across platforms.

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore/Flush.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore/Flush.hs
@@ -2,7 +2,7 @@
 
 -- |This module provides functionality for flushing the operating system buffer associated with
 -- a file handle.
-module Concordium.GlobalState.Persistent.BlobStore.Flush(hFlushOS) where
+module Concordium.GlobalState.Persistent.BlobStore.Flush (hFlushOS) where
 
 import Control.Concurrent
 import Control.Monad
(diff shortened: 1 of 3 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore/Flush.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE CPP #-}
+
+-- |This module provides functionality for flushing the operating system buffer associated with
+-- a file handle.
+module Concordium.GlobalState.Persistent.BlobStore.Flush (hFlushOS) where
+
+import Control.Concurrent
+import Control.Monad
+import Foreign.C
+import GHC.IO.Exception
```
abizjak, line 10: Unless I forget, this import is not needed on linux, and causes a compilation warning.

abizjak:
Recording here so we don't forget.

The use of fsync makes catchup (specifically out of band catchup) about 50% slower compared to the `main` version when most blocks are empty. This is on linux, other platforms may differ.

This needs to be addressed somehow since such a slowdown is not really acceptable for the gain.

EmilLa1 (approved):
Looks good

PR #114, opened 2021-07-26
Fix build of 'dev-master' images

concordium-github:
The 'dev-master' image (intended for use in Docker Compose) is build automatically on a daily basis. The base images are pushed with tag 'latest' by default and only retagged manually (there's currently no process for this).

This PR changes the build job to use the base images with tag 'latest'. Due to the automatic nature of these builds, using 'latest' seems both necessary and natural.

Part of https://concordium.atlassian.net/browse/DO-319.


bisgardo:
Replaced by https://github.com/Concordium/concordium-node/pull/121 (has proper author).

abizjak:
@bisgardo Can this branch be deleted as well?

PR #119, opened 2021-08-04
Windows service

### LINKED ISSUE #118: Windows node distribution
**Task description**

Provide a distribution of the node for Windows.
This includes a service wrapper that will run as a background service and start and monitor nodes.

**Sub-tasks**
- [ ] A service wrapper for running nodes based on a configuration file, running as a Windows service.
- [ ] Shortcuts for starting, stopping and configuring nodes.
- [ ] An installer/uninstaller.
- [ ] A pipeline for reproducible, automatic builds on Windows.

td202:
## Purpose

Support for installing and running a node on a Windows system in a convenient fashion.

Closes #118

## Changes

* Support advanced file-based logging with `log4rs` in the node by providing a `--log-config` argument.
* Jenkins file and build script for windows builds.
* Add a node runner service executable (under `service/windows`) that starts and monitors nodes.
* Add an installer for the node, collector and node runner service.

## Checklist

- [x] The code compiles without warnings. (Except for "String overflow" warnings when building the installer.)
      hard-to-understand areas.


service/windows/src/main.rs
```diff
@@ -0,0 +1,312 @@
...
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+use std::time::{Duration, Instant};
+use std::{env, thread};
+use winapi::shared::winerror::ERROR_FAILED_SERVICE_CONTROLLER_CONNECT;
+use windows_service::{service::*, service_control_handler::*, service_dispatcher};
+
+const SERVICE_NAME: &str = "ConcordiumNode";
```
EmilLa1, line 25: Why not just "Concordium Node"? Or is the UpperCamelCase the standard way of naming services?

service/windows/src/node.rs
```diff
@@ -0,0 +1,478 @@
...
+}
+
+pub enum LoggerConfig {
+    NoLog,
+    ConfigLog(PathBuf),
+    FileLog(LogFileConfig),
+}
+
+/// Runner for a node
+pub struct NodeConfig {
```
EmilLa1, line 32: I guess it could've been preferred if we could have been using the already existing configuration type(s) as they seem quite identical? Maybe the `NodeConfig` could just wrap the needed cfg types from `configuration.rs`?

stack.yaml
```diff
@@ -0,0 +1,15 @@
+# This should be kept in line with concordium-consensus/stack.yaml and
```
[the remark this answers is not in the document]
td202: This file is used for building the consensus on Windows. Also, it's useful for the Haskell language server, which is really my main reason for adding it.

service/windows/src/main.rs
td202: This seems to be the convention. There are exceptions, but very few services include spaces in their names. Note that the service also has a display name, which is how it's listed in the services panel. (Though the service name is used in Task Manager.)

service/windows/src/node.rs
td202: Perhaps, but I think there are enough meaningful differences that having a new configuration type makes sense. I think it would be a bit of an awkward refactor to completely share the configuration `struct`s.

concordium-node/src/configuration.rs
```diff
@@ -600,6 +600,12 @@ pub struct CommonConfig {
         env = "CONCORDIUM_NODE_NO_LOG_TIMESTAMP"
     )]
     pub no_log_timestamp: bool,
+    #[structopt(
+        long = "log-config",
+        help = "Configure logging with a log4rs configuration file",
```
abizjak, line 605: Please add that this disables logging to stdout/err.

I also don't think the name of the library we are using should be in the help text.

service/windows/installer/custom-actions/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "custom-actions"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+winapi = { version = "0.3.9", features = ["winnt", "winerror", "winsvc"] }
```
abizjak, line 9: Is the patch version important?

service/windows/installer/custom-actions/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+        hInstall: MsiHandle,
+        szName: LPCWSTR,
+        szValueBuf: LPWSTR,
+        pcchValueBuf: LPDWORD,
+    ) -> UINT;
+}
+
+/// A custom action to delete a directory and all of its contents.
+#[no_mangle]
+pub extern "system" fn DeleteDataFolder(install_handle: MsiHandle) -> UINT {
```
abizjak, line 24: Generally the recommendation and style we try to use is that `extern` functions are not public.

service/windows/installer/custom-actions/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+            if remove_dir_all(p).is_err() {
+                return ERROR_FUNCTION_NOT_CALLED;
+            }
+        }
+    }
+    ERROR_SUCCESS
+}
+
+const SERVICE_AUTO_START: DWORD = 0x00000002;
+const SERVICE_DEMAND_START: DWORD = 0x00000003;
```
abizjak, line 41: What is the meaning of these? Can there be documentation?

service/windows/src/config.rs
```diff
@@ -0,0 +1,327 @@
...
+    Ok(PathBuf::from(conf))
+}
+
+/// Configuration of nodes to run.
+pub struct Config {
+    pub nodes: Vec<NodeConfig>,
+}
+
+/// Convenience macro for looking up a value in a TOML table.
+macro_rules! toml_get_as {
```
abizjak, line 38: Is it really necessary to use a macro? Please try to write a normal function, if that is not possible at the very least explain the need for it.

Additionally,  why make a closure and immediately apply it?

service/windows/src/config.rs
```diff
@@ -0,0 +1,327 @@
...
+                v.as_str()
+                    .ok_or_else(|| {
+                        anyhow!(
+                            "Environment variables must be strings, but {}.{} is not",
+                            table_name,
+                            k
+                        )
+                    })?
+                    .to_string(),
+            );
```
abizjak, line 72: I'd recommend 

```rust
let value = v.as_str().with_context(|| format!("Environment variables must be strings ...)?;
env.insert(k.to_string(), value.to_string())
```
for readability.

service/windows/src/config.rs
```diff
@@ -0,0 +1,327 @@
...
+    array_name: &str,
+) -> anyhow::Result<Option<Vec<String>>> {
+    if let Some(vs) = args_array_opt {
+        let mut args = Vec::with_capacity(vs.len());
+        for arg in vs {
+            args.push(
+                arg.as_str()
+                    .ok_or_else(|| anyhow!("{} contains a non-string argument", array_name))?
+                    .to_string(),
+            );
```
abizjak, line 92: And the same here.

service/windows/src/main.rs
```diff
@@ -0,0 +1,312 @@
...
+            exit_code: ServiceExitCode::Win32(0),
+            checkpoint: 0,
+            wait_hint: Duration::from_secs(0),
+            process_id: None,
+        }
+    };
+    ($state: expr) => {
+        simple_status!($state, ServiceControlAccept::empty())
+    };
+}
```
abizjak, line 64: This macro can be replaced by two simple methods with no loss, so it should be.

Macros should only be used when needed. They are harder to understand, navigate, and maintain.

service/windows/src/main.rs
```diff
@@ -0,0 +1,312 @@
...
+            service_type: ServiceType::OWN_PROCESS,
+            current_state: $state,
+            controls_accepted: ServiceControlAccept::empty(),
+            exit_code: ServiceExitCode::Win32(0),
+            checkpoint: $checkpoint,
+            wait_hint: $wait_hint,
+            process_id: None,
+        }
+    };
+}
```
abizjak, line 80: And this can be replaced by a function directly.

service/windows/src/node.rs
```diff
@@ -0,0 +1,478 @@
...
+    node_process: Child,
+    /// The configuration settings used to start the node.
+    pub node_config: NodeConfig,
+    /// A temporary file used to redirect the node's stderr.
+    stderr_path: TempPath,
+    /// The child process for the collector.
+    collector_process: Option<Child>,
+    /// A temporary file used for configuring the node's logging.
+    /// (When this is dropped, the file will be deleted.)
+    #[allow(dead_code)]
```
abizjak, line 95: Why is this needed?

service/windows/installer/custom-actions/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
abizjak, line 1: And this create must also be formatted and added to clippy.

abizjak (changes requested):

td202:
> One thing to note however, as I'm not usually using Windows products, I'm a bit unsure on how the "expected behaviour" should be from a Windows installer. Basically, does our installer follow the "common guidelines (if such exists!?)" in order to guarantee the best possible UX.

I think generally, yes. There are guidelines such as the logo requirements: https://docs.microsoft.com/en-us/windows/win32/msi/windows-installer-and-logo-requirements. But these don't say very much about UX. I think we mostly conform with them (though I don't think I fully understand the stuff about advertising).

concordium-node/src/configuration.rs
td202: The configuration file determines whether logging is to stdout, stderr, a file or some combination. If we don't specify "log4rs", there's no clue for the user to figure out the configuration format.

service/windows/installer/custom-actions/src/lib.rs
td202: Is this documented somewhere? Generally, it seems pretty common (including in our code) to use `pub extern`, and this is what is used in the [rust documentation](https://doc.rust-lang.org/std/keyword.extern.html) also.

service/windows/src/config.rs
td202: A macro is more convenient here because it can be variadic. The closure and application allows us to use `?` in `Option`. The macro can be applied to both the `Table` and `Value` types (which have similar `get` functions). The macro can also be called with functions that return either a value or a borrowed pointer.

I've spent a while trying to rewrite this as a function, but I don't think it can be done with a single function that works for all use cases of the macro. I don't think there's enough to be gained to justify putting more effort into rewriting this.

service/windows/src/node.rs
td202: The field is never explicitly read, but is kept so that it will be dropped when the `Node` is dropped. Will rename to `_log_config_file` so that the annotation is not required.

service/windows/installer/Node.wxs
```diff
@@ -0,0 +1,349 @@
...
+        <Directory Id='TARGETDIR' Name='SourceDir'>
+            <Directory Id='ProgramFiles64Folder' Name='PFiles'>
+                <Directory Id='Concordium' Name='Concordium'>
+                    <Directory Id='INSTALLDIR' Name='Node 1.0'>
+                        <!-- The installation directory. Here, we install:
+                            * concordium-node.exe
+                            * HSdll.dll
+                            * node-collector.exe
+                            * node-runner-service.exe
+                            * third-party-licenses.exe
```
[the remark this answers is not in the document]
td202: Fixed.

service/windows/installer/custom-actions/Cargo.toml
td202: Shouldn't be. Removed.

service/windows/installer/custom-actions/src/lib.rs
td202: Added

service/windows/src/config.rs
td202: Done.

service/windows/src/config.rs
td202: Done.

service/windows/src/main.rs
td202: Done.

service/windows/src/main.rs
td202: Done.

EmilLa1 (approved):

service/windows/src/main.rs
```diff
@@ -0,0 +1,312 @@
...
+            break;
+        }
+        // Wait for a second. This can be woken early by the event handler.
+        thread::park_timeout(std::time::Duration::from_secs(1));
+    }
+
+    // Shut down the nodes.
+    // Nodes are first sent a CTRL+BREAK signal to trigger shutdown gracefully, and are allowed 61
+    // seconds to do so.  Any node that has not shut down in that time will be terminated.
+    let expected_shutdown_duration = Duration::from_secs(61);
```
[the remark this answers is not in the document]
td202: I wanted to pick a number that is long enough that the node should be able to shutdown in that time, and not longer than the usual limits on service shutdown times. The preshutdown time-out is 3 minutes. The Services control panel allows 125 seconds. If the operating system is rebooting, it may only have 5 seconds. If it is being shut down, it might have 20 seconds. But in these last cases, if the service runner has to be killed, then the child processes will also end up being killed, which is all the runner could do anyway at that point.

I picked 61 seconds because it's twice the default housekeeping interval, plus 1 second, and seems to be in a reasonable range.

service/windows/installer/custom-actions/Cargo.toml
td202: Done.

service/windows/src/main.rs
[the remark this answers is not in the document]
abizjak: I think it would be valuable if this rationale is included in the comment so it is not lost in the pull request comments only.

concordium-node/src/configuration.rs
abizjak: I think a common way for this is to include a default file, or a link to it, with documentation in the file, in the form of comments, of what the options do. It is more user friendly than a name of a crate which has fairly poor user documentation. We also don't say what features the crate is used with, so this help line is not really enough.

service/windows/src/config.rs
abizjak: It is more convenient to write perhaps, but it is harder to understand. 

I feel that the parsing logic is already hard to follow, and using this macro does not help. Having clearer named type-specific functions would in my view help.

abizjak (approved):
I would still prefer the unresolved comments to be addressed, but if they are not, I will just be quietly unhappy since they are not blocking.

If there is a will to address at least one issue, I would strongly encourage the one about the log file format.

PR #121, opened 2021-08-04
Fix docker-compose setup

bisgardo:
## Purpose

The `dev-master` image (intended for use in Docker Compose) is built automatically on a daily basis. The base images are pushed (also automatically) with tag `latest` and only retagged manually (there's currently no process for this).

The image currently doesn't build and the deployment script is out of date as well.

## Changes

The build job now use the base images with tag `latest`. Due to the automatic nature of the builds, this seems both necessary and natural.

The deployment script has been updated to configure the binaries directly with environment variables as much as possible. The entrypoint scripts are targeted to only include exactly what is needed.

Part of https://concordium.atlassian.net/browse/DO-319.

## Checklist

      hard-to-understand areas.


docker-compose/build.sh
```diff
@@ -12,11 +11,14 @@ VERSION="$1"
...
+# Flag '--pull' ensures that images are re-pulled.
 docker build \
 	-f docker-compose/Dockerfile \
 	--build-arg ghc_version="${GHC_VERSION}" \
 	--build-arg "consensus_profiling=$CONSENSUS_PROFILING" \
 	--label ghc_version="${GHC_VERSION}" \
 	--label "consensus_profiling=$CONSENSUS_PROFILING" \
 	-t "concordium/dev-node:$VERSION" \
+	--pull \
+	--no-cache \
```
bisgardo, line 23:
```suggestion
```

docker-compose/bakers.yaml
```diff
@@ -2,67 +2,56 @@ version: '3'
...
+    - BAKER_ID_GEN_DNS=baker_id_gen
+    - CONCORDIUM_NODE_CONNECTION_DESIRED_NODES=${DESIRED_PEERS}
+    - CONCORDIUM_NODE_CONNECTION_BOOTSTRAP_NODES=bootstrapper:8888
+    - CONCORDIUM_NODE_EXTERNAL_PORT=8888
+    - CONCORDIUM_NODE_DATA_DIR=/var/lib/concordium/data
+    - CONCORDIUM_NODE_CONFIG_DIR=/var/lib/concordium/config
+    - CONCORDIUM_NODE_RPC_SERVER_ADDR=0.0.0.0
     ports:
-    - 11100-11500:10000
+    - "11100-11500:10000"
```
td202, line 44: For some reason, I find on Windows this syntax doesn't work. The only way I've found to expose the ports is to bind them to ephemeral host ports, e.g.:
```suggestion
    - "127.0.0.1::10000"
```
`docker ps` can then show you how the bindings are set up.

td202 (changes requested):
Because of the way inflation is calculated, which is slow on non-GMP builds, it takes a very long time for nodes to get going when the genesis time is far in the past. Either `scripts/genesis/genesis.json` should be updated with a more recent genesis time, or `generate-test-genesis.py` could be updated to set the genesis time to the current time.

docker-compose/bakers.yaml
bisgardo: That's fine by me. But why bind explicitly to `127.0.0.1` instead of just `"10000"`?

docker-compose/bakers.yaml
```diff
@@ -2,67 +2,56 @@ version: '3'
...
-    - DESIRED_PEERS=${DESIRED_PEERS}
-    - BOOTSTRAP_FIRST_NODE=bootstrapper:8888
-    - EXTERNAL_PORT=8888
-    - NUM_BAKERS=${NUM_BAKERS}
-    - MODE=local_basic
-    - DATA_DIR=/var/lib/concordium/data
-    - CONFIG_DIR=/var/lib/concordium/config
-    - RPC_SERVER_ADDR=0.0.0.0
-    - EXTRA_ARGS=${EXTRA_ARGS}
+    - LD_LIBRARY_PATH=/usr/local/lib
```
abizjak, line 34: I don't think this is necessary since the node is linked statically.

docker-compose/bootstrapper-entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -exo pipefail
+
+# Ensure that expected dirs exist.
+mkdir -p "${CONCORDIUM_NODE_CONFIG_DIR}" "${CONCORDIUM_NODE_DATA_DIR}"
```
abizjak, line 6: Would it not be better to create this directory directly in the build and avoid this needless indirection via the start script, just setting the entrypoint to the bootstrapper.

docker-compose/node-entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+# Ensure that expected dirs exist.
+mkdir -p "${CONCORDIUM_NODE_CONFIG_DIR}" "${CONCORDIUM_NODE_DATA_DIR}"
+
+if [ -n "${GENESIS_DATA_PATH}" ]; then
+	# Copy 'genesis.dat' - better solution: support flag in concordium-node to set location.
+	cp "${GENESIS_DATA_PATH}/genesis.dat" "${CONCORDIUM_NODE_DATA_DIR}"
+	# Select unique baker credentials file.
+	id="$(curl "http://${BAKER_ID_GEN_DNS}:8000/next_id")"
+	export CONCORDIUM_NODE_BAKER_CREDENTIALS_FILE="${GENESIS_DATA_PATH}/bakers/baker-${id}-credentials.json"
+fi
```
abizjak, line 14: What is the intend of the `GENESIS_DATA_PATH` variable? What if it is not set, is the assumption that the data dir already contains genesis.dat? If not, I think the script should fail directly here.

Could you document the intended behaviour please.

docker-compose/README.md
```diff
@@ -5,49 +5,47 @@ spinning up a network using docker-compose for local development.
 
 ## Usage
 
-Two parameterized `docker-compose` files are available:
+A parameterized Docker Compose file `bakers.yaml` is available:
```
abizjak, line 8: There are a number of environment variables that can be set, can you please refer to them in this readme (and a pointer to the node environment variables as well).

docker-compose/bakers.yaml
bisgardo: done

docker-compose/bootstrapper-entrypoint.sh
bisgardo: good idea, done

docker-compose/README.md
bisgardo: done

abizjak (approved):

docker-compose/node-entrypoint.sh
bisgardo: Added some docs to explain the setup. I think that since the node will fail by itself if `genesis.dat` is shouldn't be necessary to do that same check here.

Note that I don't expect users to modify the Compose file. But if they decide to remove `GENESIS_DATA_PATH` it must be because they want to take control of the setup just like if they change some other setting that we also don't check. In that case I think it's best for the scripts to get out of the way and interfere as little as possible.

PR #122, opened 2021-08-05
Create SQL tables with a useful index.

abizjak:
## Purpose

Every time we query the database we filter by account or contract and possibly narrow down by ID. The new indices reflect this and make many queries substantially faster.

The downside is that they do take up more space than the previous ones, but I believe this is a good tradeoff since they make the wallet-proxy more performant and thus more usable.

## Changes

- Replace the automatically derived migration with a custom database creation script that allows more flexibility.
- Check for existence of databases and correct column types on startup manually.
- Log the state of transaction logging databases on startup.

Related PRs
- https://github.com/Concordium/concordium-base/pull/64

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState.hs
@@ -425,9 +425,14 @@ instance (IsProtocolVersion pv) => GlobalStateConfig (DiskTreeDiskBlockWithLogCo
     initialiseGlobalState (DTDBWLConfig rtparams gendata txLog) = do
         -- check if all the necessary database files exist
       (blockStateFile, existingDB) <- checkExistingDatabase rtparams
-      dbHandle <- liftIO $ do
-        dbHandle <- connectPostgres txLog
-        createTable dbHandle
+      dbHandle <- do
+        dbHandle <- liftIO $ connectPostgres txLog
+        liftIO (checkTablesExist dbHandle) >>= \case
+          Ok -> logEvent GlobalState LLInfo "Using existing PostgreSQL tables for transaction logging."
+          NoTables -> do
+            logEvent GlobalState LLInfo "No relevant tables found in transaction logging database. Creating them."
+            liftIO $ createTables dbHandle
+          IncorrectFormat -> logExceptionAndThrow GlobalState (DatabaseOpeningError (userError "The connected SQL database has some of the 'ati', 'cti', or 'summaries', but either not all or they have incorrect columns."))
         return dbHandle
       if existingDB then do
         pbscBlobStore <- liftIO $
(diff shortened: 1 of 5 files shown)


concordium-consensus/src/Concordium/GlobalState.hs
```diff
@@ -425,9 +425,14 @@ instance (IsProtocolVersion pv) => GlobalStateConfig (DiskTreeDiskBlockWithLogCo
...
-        dbHandle <- connectPostgres txLog
-        createTable dbHandle
+      dbHandle <- do
+        dbHandle <- liftIO $ connectPostgres txLog
+        liftIO (checkTablesExist dbHandle) >>= \case
+          Ok -> logEvent GlobalState LLInfo "Using existing PostgreSQL tables for transaction logging."
+          NoTables -> do
+            logEvent GlobalState LLInfo "No relevant tables found in transaction logging database. Creating them."
+            liftIO $ createTables dbHandle
+          IncorrectFormat -> logExceptionAndThrow GlobalState (DatabaseOpeningError (userError "The connected SQL database has some of the 'ati', 'cti', or 'summaries', but either not all or they have incorrect columns."))
```
SimonKamp, line 435:
```suggestion
          IncorrectFormat -> logExceptionAndThrow GlobalState (DatabaseOpeningError (userError "The connected SQL database has some of the 'ati', 'cti', or 'summaries', but either not all of them exist or at least one of them has incorrect columns."))
```

SimonKamp (approved):
Looks good. Maybe the `IncorrectFormat` message should be changed.

td202:
- Are there any cases where the cti or ati tables are looked up just by the id? Likely, this is not the case, but code generated by the Persist template Haskell will assume that the `id` column is the primary key.
- Would it be worth updating the Persist entity definition to reflect the changed primary key?
- Is it possible to generate (some of) the SQL to ensure type safety and reduce the maintenance burden?
- Has the tokenomics reporter been tested with these changes? I imagine that it should be faster as a result.

abizjak:
>     * Are there any cases where the cti or ati tables are looked up just by the id? Likely, this is not the case, but code generated by the Persist template Haskell will assume that the `id` column is the primary key.

Not in any of the tools we use.
 
>     * Would it be worth updating the Persist entity definition to reflect the changed primary key?

I tried, but it is annoying since we want the ID to be automatically generated on insertion. Additionally setting Primary key constraint in persistent requires all columns in the key to have FromJSON, ToJSON and Read instances which does not make sense for the key we have. We could make them, but they would be stupid.

>     * Is it possible to generate (some of) the SQL to ensure type safety and reduce the maintenance burden?

As above. While it would be possible, it would be equally or more hacky since Persistent has its own defaults and I am not in the mood to spend reading tea leaves of template haskell errors.

Changing this database schema is a big breaking change anyhow, so I think the maintenance burden is minor.

Additionally, having the schema explicit is also good because this database will be used outside of Haskell.

>     * Has the tokenomics reporter been tested with these changes? I imagine that it should be faster as a result.

No.

td202 (approved):

PR #127, opened 2021-08-10
Docker Compose: Various cleanup

bisgardo:
## Purpose

Various minor improvements and simplifications to the Docker Compose setup.

## Changes

- Replace 'concordium-tools-baker-id-gen' with trivial node.js script (allows us to delete the [repository](https://github.com/Concordium/concordium-tools-baker-id-gen).
- Remove obsolete "consensus profiling" parameter.
- Inline build script into Jenkinsfile.
- Remove unnecessary option for delaying collector startup.
- Make scripts read all parameters and set default values eagerly (our usual convention for scripts).
- Other minor script refactorings.

## Checklist

      hard-to-understand areas.


docker-compose/node-entrypoint.sh
```diff
@@ -2,19 +2,22 @@
 
 set -euxo pipefail
 
+genesis_data_path="${GENESIS_DATA_PATH-}"
+data_dir="${CONCORDIUM_NODE_DATA_DIR}"
```
abizjak, line 6: Why do this renaming? CONCORDIUM_NODE_DATA_DIR is the environment variable the node uses, it seems to introduce needless overhead to rename a variable.

docker-compose/node-counter.js
```diff
@@ -0,0 +1,4 @@
+var count = 0;
+require('http')
+	.createServer((req, res) => res.end(String(count++)))
```
abizjak, line 3: How atomic is this increment?

I also don't like we are introducing another language to this. Could we not use python instead since we are already using it for other things here.

docker-compose/node-entrypoint.sh
bisgardo: The style that we've found useful in bash in general is to extract all expected variables up front with the convention to only use them as lower-case variables afterwards (to separate internal from external variables and avoid leaking). This improves error handling by ensuring that the script doesn't crash mid-way because something is missing (the `set -u` flag makes it crash immediately) and that default values are processed in a single location. It also makes it easier to see from the source which variables are expected.

docker-compose/node-counter.js
bisgardo: Javascript is single-threaded so this is as atomic as it gets. I don't know how to do it in python but I highly doubt that it is as trivial as this.

I agree that new languages should be introduced carefully, but in this case where it's practically a one-liner, I think it's a perfect match for the problem at hand.

docker-compose/node-entrypoint.sh
abizjak: This is contradicted by the use of BAKER_ID_URL below.

I find this style annoying since it makes it harder to understand what is going on or trace configurations since there is always many different names for the thing you are looking for so searching is hard.

bisgardo: Yeah that's because `BAKER_ID_URL` is required only if `GENESIS_DATA_PATH` is set. I found that adding the logic to handle that up front was overkill. So no the convention is not perfect but I've found it to simplify debugging tremendously time and again. Not that I expect it to matter much with a script this simple, I just like to be consistent about it.

docker-compose/node-counter.js
abizjak: Ok. I'm still unhappy that we are doing this, but ok. For the record, the equivalent python, while not as trivial, is only slightly less trivial

```python
from http.server import *

count = 0

class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        global count
        self.send_response(200)
        self.send_header('Content-type','text/html')
        self.end_headers()
        message = str(count)
        count += 1
        self.wfile.write(bytes(message, "utf8"))

with HTTPServer(('0.0.0.0', 3000), Handler) as server:
    server.serve_forever()
```

abizjak (approved):

docker-compose/node-entrypoint.sh
abizjak: It would alleviate the annoyance I have greatly if the convention was to just lowercase the variables instead of giving them different names since searching with "ignore case" is well supported in tools.

td202 (approved):

docker-compose/node-entrypoint.sh
bisgardo: done

PR #129, opened 2021-08-11
Make the scheduler test utils parametric over the protocol version.

abizjak:
## Purpose

To enable tests for the upcoming protocol 2 transactions. Since there is lots of noise introduced by this change it is better if it is a trivial separate PR instead of mixing it with the upcoming changes.

## Changes

Internal changes in tests only.

## Checklist

      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG

### DIFF
--- a/concordium-consensus/tests/scheduler/SchedulerTests/TestUtils.hs
+++ b/concordium-consensus/tests/scheduler/SchedulerTests/TestUtils.hs
@@ -10,7 +10,7 @@ Also checks invariants on the block state after each processed transaction.
 NOTE: This processes each transaction individually - for testing grouped transactions, see
       'SchedulerTests.TransactionGroupingSpec' and 'SchedulerTests.TransactionGroupingSpec2'.
 -}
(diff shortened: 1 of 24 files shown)


eb-concordium (approved):

concordium-consensus/tests/scheduler/SchedulerTests/TestUtils.hs
```diff
@@ -85,16 +85,16 @@ defaultParams = TestParameters
 -- | A test case for executing a list of transactions, specifying 'ResultSpec's for the result
 -- of each transaction's execution. The transactions are run with 'Sch.filterTransactions'
 -- in sequenced and not grouped, with the given parameters.
-data TestCase = TestCase
+data TestCase pv = TestCase
```
td202, line 88: Perhaps a test case should existentially quantify the protocol version, subject to `IsProtocolVersion pv`. That way, `mkSpecs` can process lists of test cases of multiple versions.

td202 (commented):
I am not sure that the noise justifies removing this PR from its context. Many of the files will have to be changed again to have test cases for the new protocol version, in any case.

PR #134, opened 2021-08-14
MacOS node

Bargsteen:
## Purpose

Create a distribution of the node for macOS.

## Changes

- Add logging to the unified logging system on macOS.
  - This required adding a C wrapper for native calls, as they were macros.
  - Add a flag `--use-mac-log` to the node and the collector, which is used when logging to distinguish different nodes running at the same time, fx mainnet and testnet.
- Add a build script which creates an installer for macOS. This includes a number of extra files and small programs. More details exist in the readme at `scripts/distribution/macOS-package/README.md`

## Checklist

      hard-to-understand areas.


EmilLa1 (changes requested):
Fine job overall!

Only a few comments. 

Also, we need to address that the building of the installer currently does not seem to work with a non-beta xcode version. :-)

concordium-node/macos_log_wrapper.c
```diff
@@ -0,0 +1,5 @@
+#include <os/log.h>
+
+void wrapped_os_log_with_type(os_log_t log, os_log_type_t type, const char* message) {
+    os_log_with_type(log, type, "%{public}s", message);
+}
```
abizjak, line 5: Please document the reason for this (the fact that os_log_with_type is a macro, so cannot be used in FFI if I recall correctly).

concordium-node/src/utils.rs
```diff
@@ -85,6 +97,37 @@ pub fn setup_logger_env(env: Env, no_log_timestamp: bool) {
...
+    // in the message.
+    let (level_filter, log_lvl) = if trace {
+        (LevelFilter::Trace, "trace")
+    } else if debug {
+        (LevelFilter::Debug, "debug")
+    } else {
+        (LevelFilter::Info, "info")
+    };
+
+    let subsystem_name = format!("software.concordium.{}.node", net_name);
```
abizjak, line 119: I think the entire subsystem name should be given. This partial combination where some of it is hardcoded, and some is given seems unnecessarily complicated.

concordium-node/Cargo.toml
```diff
@@ -65,6 +65,7 @@ hex = "0.4"
 rpassword = "5.0"
 anyhow = "1.0"
 thiserror = "1.0"
+dashmap = "4"
```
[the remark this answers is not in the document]
abizjak: I agree with Emil, I would strongly encourage you to remove this dependency.

concordium-node/src/bin/collector.rs
```diff
@@ -120,24 +121,25 @@ struct ConfigCli {
...
-        Env::default().filter_or("LOG_LEVEL", "info")
-    } else {
-        Env::default().filter_or("LOG_LEVEL", "warn")
-    };
-
-    setup_logger_env(env, conf.no_log_timestamp);
+    #[cfg(target_os = "macos")]
+    setup_logger(conf.trace, conf.debug, &conf.net_name);
+    #[cfg(not(target_os = "macos"))]
+    setup_logger(conf.trace, conf.debug, conf.no_log_timestamp);
```
abizjak, line 142: I am not sure this is a good idea to only support the bespoke system logger on the mac.

I think a better solution would be to have configuration option `CONCORDIUM_NODE_COLLECTOR_USE_MAC_LOG` (or whatever name you deem good) which is an `Option<String>`. If given the system logger is used, and the provided string is the service name (`software.concordium.mainnet...`).

The same comment applies to the node configuration below.

abizjak (changes requested):

scripts/distribution/macOS-package/build.sh
```diff
@@ -0,0 +1,361 @@
...
+# Copy the compiled items (binaries and supporting data) to the build folder.
+function copyCompiledItemsToBuildDir() {
+    copyNodeBinaries
+    copyInstallerPluginData
+}
+
+# Get the tool dylibbundler, which is used to recursively find and
+# bundle the needed dylibs for a binary.
+# Checks whether the tool is already installed and then potentially skips the build step.
+function getDylibbundler() {
```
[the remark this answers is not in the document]
EmilLa1: Moreover as far as I see, the usage of `macdylibbundler` makes it hard to reproduce builds, which i'm not much in favor of. I would prefer a more rigorous way of bundling e.g., by a script which uses the above tools.

concordium-node/src/macos_log.rs
```diff
@@ -0,0 +1,220 @@
...
+/// Map log levels into appropriate macOS levels.
+/// Note that we coalesce `Trace` and `Debug` into `Debug`, since the log viewer
+/// on mac Console.app, only has The macOS log viewer, Console.app, shows Error
+/// and Fault by default. Info and Debug has to be enabled manually.
+impl From<log::Level> for Level {
+    fn from(other: log::Level) -> Self {
+        match other {
+            log::Level::Trace => Self::Debug,
+            log::Level::Debug => Self::Debug,
+            log::Level::Info => Self::Info,
```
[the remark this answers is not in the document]
Bargsteen: As discussed with Ales, the info and debug messages are of little value with respect to offering customer support, so they should not be stored on disk by default. However, it is possible to configure this using `sudo log config --subsystem <subsystem-name> --mode persist:debug`, which I've described in the readme.
Info and debug messages can, however, be useful for developers, who can stream them using `log stream` or `Console.app`, or manually toggle the persistence.
I also made the native macOS logging system optional, so developers might also opt for the standard logging method.

scripts/distribution/macOS-package/build.sh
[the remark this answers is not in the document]
Bargsteen: I've mentioned `otool` and `install_name_tool` as an alternative in the README.
The builds should be functionally reproducible, as we are using a fixed resolver for Stack. I've added a comment about this in the build script.

concordium-node/src/utils.rs
```diff
@@ -91,6 +131,8 @@ pub fn get_config_and_logging_setup() -> anyhow::Result<(config::Config, config:
             "enabled"
         }
     );
+
+    // FIXME: This is only the log level for consensus when log_config is used.
```
EmilLa1, line 135: Fixme?

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -0,0 +1,120 @@
+//
+//  MyInstallerPane.m
+//  NodeConfigurationInstallerPlugin
+//
+//  Created by Concordium on 30/07/2021.
```
EmilLa1, line 5: If there is comments, they should be descriptive. 
This block doesn't state more than the obvious.

scripts/distribution/macOS-package/README.md
```diff
@@ -0,0 +1,172 @@
+# Concordium Node for MacOS
+
+## Prerequities
+
+- Xcode 13
```
EmilLa1, line 5: There was no way of making this work with the currently released XCode?

scripts/distribution/macOS-package/build.sh
```diff
@@ -0,0 +1,373 @@
...
+                    && cd "$toolsDir" \
+                    && unzip "dylibbundler.zip" \
+                    && logInfo " -- Building..." \
+                    && cd "$macdylibbundlerDir" \
+                    && make
+        logInfo "Done"
+    fi
+}
+
+# Use dylibbundler to recursively find and bundle the dylibs for node and collector.
```
EmilLa1, line 212: I thought we agreed not to use dylibbundler, but to do the bundling ourselves?

EmilLa1 (commented):
A few questions and remarks. Otherwise fine.

concordium-node/src/utils.rs
Bargsteen: This is an issue introduced in another PR, so I simply added a note for now. It should be fixed in a separate PR and has nothing to do with the mac node :)

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
Bargsteen: Xcode adds these comments to all code files by default. I can remove them if you want, but they are common to see in iOS/macOS apps, including our own wallet: https://github.com/Concordium/concordium-reference-wallet-ios/tree/main/ConcordiumWallet ;)

scripts/distribution/macOS-package/README.md
Bargsteen: No, unfortunately not. I even tried Xcode versions from 2019 and 2018. However, Xcode 13 should be released later this month.

scripts/distribution/macOS-package/build.sh
Bargsteen: Ales and I arrived at the conclusion that it is ok for now, as we have a fixed resolver for Stack. This means that builds will use the exact same versions of the dependencies. Installers built on different machines might have different hashes, though, as the dependencies are built on different machines and GHC doesn't promise that level of reproducibility.

scripts/distribution/macOS-package/template/scripts/postinstall
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
+set -e
+
+function addBinariesToPath() {
+    # Ensure the local bin folder exists
+    echo "Ensuring '/usr/local/bin' exists";
+    if [ ! -d "/usr/local/bin" ]; then
+        mkdir "/usr/local/bin";
+    fi
```
abizjak, line 10: Use `mkdir -p` unless mac does not have that.

concordium-node/src/bin/collector.rs
```diff
@@ -120,24 +121,28 @@ struct ConfigCli {
         env = "CONCORDIUM_NODE_COLLECTOR_MAX_GRPC_FAILURES_ALLOWED"
     )]
     pub max_grpc_failures_allowed: u64,
+    #[cfg(target_os = "macos")]
+    #[structopt(
+        long = "use-mac-log",
+        help = "Enable native logging on macOS by providing a subsystem name, e.g. \
+                'software.concordium.mainnet.node'. Log messages can be found via Console.app or \
+                the log commandline tool by searching for the subsystem.",
```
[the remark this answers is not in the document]
Bargsteen: Ah, neat. Added :)

abizjak (approved):
I am OK with the changes to the node for this, however since I do not have access to a mac I am unable to test the installation and meaningfully review the mac specific parts.

EmilLa1 (approved):

PR #143, opened 2021-08-27
Genesis: Add dockerfile

bisgardo:
The image simplifies the process of generating new genesis test configurations.

### DIFF
--- a/scripts/genesis/Dockerfile
+++ b/scripts/genesis/Dockerfile
@@ -1,0 +1,16 @@
+FROM concordium/genesis-tools:latest
+RUN apt-get update && \
+    apt-get -y install python3 && \
+    rm -rf /var/lib/apt/lists/*
+
+# Use globally installed tools.
+ENV GENESIS_DAT_TOOL=genesis
+ENV GENERATE_UPDATE_KEYS=generate-update-keys
+ENV CLIENT_TOOL=client
+ENV GENESIS_ACCOUNTS_TOOL=genesis_tool
+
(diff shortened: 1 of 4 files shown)


scripts/genesis/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV GENESIS_DAT_TOOL=genesis
+ENV GENERATE_UPDATE_KEYS=generate-update-keys
+ENV CLIENT_TOOL=client
+ENV GENESIS_ACCOUNTS_TOOL=genesis_tool
+
+WORKDIR /work
+ENV GENESIS_DIR=/work/out
+
+COPY ./generate-test-genesis.py /
+ENTRYPOINT [ "/generate-test-genesis.py" ]
```
abizjak, line 16: I would like that I was consulted before this change was made.

The generate-test-genesis.py script already supports using the genesis-tools docker image and it is not clear from this why the additional dockerfile was needed. It incurs a maintenance burden.

Moreover the documentation does not explain why this complexity is preferable to using

```console
USE_DOCKER= ./generate-test-genesis.py
```
that is already documented in the README.

At the very least there must be an explanation as to why this is here in the same place that the use of the genesis script is explained already.

bisgardo: I didn't see this because it was in the parent folder's readme, not the one in this folder.

But now that I do, I still think that using a trivial dockerfile is less complex and easier to understand than the existing solution (which requires you to install an extra dependency).

PR #146, opened 2021-09-03
Update protocol update specification hash.

abizjak:
See https://github.com/Concordium/concordium-update-proposals for reference.

## Purpose

Use correct specification hash for the update.


abizjak:
Please double check that the hashes are correct.

eb-concordium (approved):

td202 (approved):

PR #148, opened 2021-09-03
Multiversion query changes

td202:
## Purpose

Revise queries handling for multi-genesis.

## Changes

* The transaction log should record the absolute height of blocks.
* The `GetBlocksAtHeight` query uses allows specifying the genesis index to use as a base. If it is unspecified or 0, it is treated as absolute. The query also allows restricting the query to a specific genesis index.
* The parent of a regenesis block is reported as the last finalized block of the previous chain.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Queries.hs
+++ b/concordium-consensus/src/Concordium/Queries.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE MonoLocalBinds #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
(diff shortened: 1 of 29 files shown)


concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -79,28 +83,49 @@ liftSkovQueryLatest a = MVR $ \mvr -> do
...
+      VersionedSkovM gsconf finconf pv a
     ) ->
-    MVR gsconf finconf (Maybe a)
-liftSkovQueryLatestResult a = MVR $ \mvr -> do
+    MVR gsconf finconf a
+liftSkovQueryAtGenesisIndex genIndex a = MVR $ \mvr -> do
+    versions <- readIORef (mvVersions mvr)
+    liftSkovQuery mvr (versions Vec.! fromIntegral genIndex) a
+
+atLatestSuccessfulVersion ::
```
abizjak, line 100: Please document.

concordium-node/src/rpc.rs
```diff
@@ -563,7 +563,11 @@ impl P2p for RpcServerImpl {
     ) -> Result<Response<JsonResponse>, Status> {
         authenticate!(req, self.access_token);
         call_consensus!(self, "GetBlocksAtHeight", JsonResponse, |cc: &ConsensusContainer| {
-            cc.get_blocks_at_height(req.get_ref().block_height)
+            cc.get_blocks_at_height(
+                req.get_ref().block_height,
+                req.get_ref().from_genesis_index,
+                req.get_ref().restrict_to_genesis_index,
```
abizjak, line 569: Is `false` the default value for the `bool` type?

It would be useful to document that we are using the semantics of default values in protobuf.

abizjak (approved):

eb-concordium (approved):

PR #149, opened 2021-09-06
Update debian packages build setup and instructions.

abizjak:
## Purpose

Update the build scripts and descriptions for the deb packages.

## Changes

- Provide simple scripts for building mainnet and testnet versions.
- Change directory location for databases in packages so that they do not conflict.
- Enable testnet and mainnet packages to be installed and run at the same time.
- The default ports are set the same as for the windows service.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/distribution/ubuntu-packages/README.md
+++ b/scripts/distribution/ubuntu-packages/README.md
@@ -71,7 +71,7 @@ The general strategy for building the package is as follows.
 
 1. Obtain genesis-data for the specific network. Copy `genesis.dat` to the
    [template/data](template/data) directory, naming it
-   `$build_env_name_lower-genesis.dat` (e.g., `testnet-genesis.dat`).
+   `${build_env_name_lower}-genesis.dat` (e.g., `testnet-genesis.dat`).
 2. Update [template/debian/changelog](template/debian/changelog) as appropriate.
 3. Update [template/debian/control](template/debian/control) as appropriate.
 4. Possibly update `CONCORDIUM_NODE_COLLECTOR_URL` in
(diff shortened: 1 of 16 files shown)


limemloh (approved):

scripts/distribution/ubuntu-packages/README.md
```diff
@@ -145,20 +165,20 @@ These files are __not__ removed by the uninstall scripts since they contain
 # Configuration of the node
 
 The node is configured via `systemd` unit files. When installed there is only
-the system unit file in `/lib/systemd/system/concordium-node.service` which
+the system unit file in `/lib/systemd/system/concordium-testnet-node.service` which
```
EmilLa1, line 168: This is not only related to this line, but the section "Configuration of the node" as a whole.
It might be a good idea, to "go back" to your previous used notation i.e., ${build_env_name_lower} instead of simply using 'testnet' for the whole section.
This would emphasize (tho obvious) that it would be the same for `mainnet` and other possible environments.

scripts/distribution/ubuntu-packages/template/debian/rules
```diff
@@ -2,17 +2,12 @@
...
-# 		docker run -v $(shell pwd)/binaries:/out static-node-binaries cp /build/bin/concordium-node /out
-# 		docker run -v $(shell pwd)/binaries:/out static-node-binaries cp /build/bin/node-collector /out
-
 # install two services
 # They will be enabled and started automatically when installed.
 # Enabled means they will be started on boot.
 override_dh_installsystemd:
-		dh_installsystemd --name=concordium-node # --no-enable --no-start
-		dh_installsystemd --name=concordium-node-collector # --no-enable --no-start
+		dh_installsystemd --name=concordium-${build_env_name_lower}-node # --no-enable --no-start
```
EmilLa1, line 9: Any particular reasons that the options (which are commented out) are included?

EmilLa1 (approved):
I think it's good to go. 
A few minor remarks which I will let be up to you.

scripts/distribution/ubuntu-packages/template/debian/rules
abizjak: The reason is historical, but I left them in so that they are easy to discover. I'll make it into a proper comment instead.

PR #150, opened 2021-09-06
Commit to version 1.1.0.

abizjak:
## Purpose

Bump the latest released version number since we are starting the test phase.

## Changes

No code changes, just reflect the release in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+## concordium-node 1.1.0
+
 - Update dependencies, code cleanup, and removal of the staging_net feature. No functional changes.
 - Fix a bug in average throughput calculation that was triggered in some cases
   of bad connectivity.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,60 +4,65 @@
 
 ## concordium-node 1.1.0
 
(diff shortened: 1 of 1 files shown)


EmilLa1 (approved):

td202:
We should take this opportunity to clean up the changes a bit.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+## concordium-node 1.1.0
+
 - Update dependencies, code cleanup, and removal of the staging_net feature. No functional changes.
```
td202, line 7: "No functional changes." is misleading.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
...
 - Fix a bug in average throughput calculation that was triggered in some cases
   of bad connectivity.
```
td202, line 9: Perhaps we should credit kryptomouse for this?

abizjak: We have not named individual contributions like this for other changes, so it would not be correct to single out this instance.

I also don't think this belongs to the changelog. It would make sense in release notes to bring attention to external contributions.

abizjak:
> We should take this opportunity to clean up the changes a bit.

I organized the changelog a bit and removed the intermediate states that no longer make sense.

PR #152, opened 2021-09-07
Fix windows build script

td202:
## Purpose

The windows build script should set the default rust version to ensure that the rust dependencies are built with the correct version.

## Changes

Set default rust version in windows build script.

## Checklist

      hard-to-understand areas.


abizjak (commented):
I'd recommend removing all `+$rustVersion-x86_64-pc-windows-gnu` from cargo invocations as well.

Only the `msvc` ones are necessary.

abizjak (approved):

PR #153, opened 2021-09-07
Distribution: Stagenet: Fix distribution image build

bisgardo:
## Purpose

The [install link](https://getenvoy.io/install.sh) for Envoy is no longer active and is now redirecting to the documentation for how to install the software using distro-specific packages instead.

## Changes

Replacing the old method of installing and running Envoy with the official instructions for installing a Debian package (on Ubuntu). The installed version of the package is pinned, but results in a downgrade from v. `1.18.3` to `1.18.2` as that is the newest available version in the repo.

Also updated the "environment" string used to construct the bootstrapping URL.

### DIFF
--- a/scripts/distribution/builder.Dockerfile
+++ b/scripts/distribution/builder.Dockerfile
@@ -52,7 +52,7 @@ FROM ubuntu:20.04
 
 # Which environment we are building the image for.
 # This affects URLs. Currently it should be either
-#   - eu.staging.concordium.com
+#   - staging.concordium.com
 #   - testnet.concordium.com
 #   - mainnet.concordium.software
 ARG environment
--- a/scripts/distribution/builder.Dockerfile
+++ b/scripts/distribution/builder.Dockerfile
@@ -75,9 +75,14 @@ ENV DISTRIBUTION_CLIENT=true
 ENV CONCORDIUM_NODE_COLLECTOR_URL=https://dashboard.${environment}/nodes/post
 ENV CONCORDIUM_NODE_COLLECTOR_GRPC_HOST=http://localhost:10000
 
-RUN apt-get update && apt-get install -y curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq
-RUN curl -L https://getenvoy.io/install.sh | bash -s -- -b /usr/local/bin
-RUN getenvoy use 1.18.3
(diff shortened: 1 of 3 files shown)


scripts/distribution/builder.Dockerfile
```diff
@@ -52,7 +52,7 @@ FROM ubuntu:20.04
 
 # Which environment we are building the image for.
 # This affects URLs. Currently it should be either
-#   - eu.staging.concordium.com
+#   - staging.concordium.com
```
bisgardo, line 55:
```suggestion
#   - stagenet.concordium.com
```

abizjak (approved):

scripts/distribution/builder.Dockerfile
```diff
@@ -75,9 +75,14 @@ ENV DISTRIBUTION_CLIENT=true
 ENV CONCORDIUM_NODE_COLLECTOR_URL=https://dashboard.${environment}/nodes/post
 ENV CONCORDIUM_NODE_COLLECTOR_GRPC_HOST=http://localhost:10000
 
-RUN apt-get update && apt-get install -y curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq
-RUN curl -L https://getenvoy.io/install.sh | bash -s -- -b /usr/local/bin
-RUN getenvoy use 1.18.3
+RUN apt-get update && \
+    apt-get install -y curl netbase ca-certificates supervisor nginx libnuma1 libtinfo6 libpq-dev liblmdb-dev jq apt-transport-https gnupg2 curl lsb-release && \
+    rm -rf /var/lib/apt/lists/*
```
td202, line 80: I'm not clear what the purpose of this line is, and why it's here specifically.

EmilLa1 (approved):

scripts/distribution/builder.Dockerfile
bisgardo: The `rm` line? It's [recommended best practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) to reduce the size of the image by removing the repo cache.

td202: Seems reasonable. Should it also be added at the end of the next run step?

bisgardo: Yes!

PR #156, opened 2021-09-08
Correct the version number used in the Windows installer

td202:
## Purpose

Correct the version number used in the Windows installer.

## Changes

- 1.0 install directory becomes 1.1.0.
- Installer 1.0.2 becomes 1.1.0.
- Use Windows line endings in the nodes.toml file.
- Add a note in the Readme about updating the version.
- Update product ID and allow upgrading from 1.0.2 to 1.1.0.

## Checklist


service/windows/installer/Node.wxs
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:Util="http://schemas.microsoft.com/wix/UtilExtension">
-    <Product Name="Concordium Node" Manufacturer="Concordium Software" Id="ae8c6e39-db6f-4538-9b8f-efefa3e195ad" UpgradeCode="297295b4-c716-4d33-8170-0f6136663bfd" Language="1033" Codepage="1252" Version="1.0.2">
-        <Package Id="*" Keywords="Concordium Installer" Description="Concordium Node 1.0.2 Installer" Manufacturer="Concordium Software" InstallerVersion="500" Languages="1033" Compressed="yes" InstallScope="perMachine" />
+    <Product Name="Concordium Node" Manufacturer="Concordium Software" Id="ae8c6e39-db6f-4538-9b8f-efefa3e195ad" UpgradeCode="297295b4-c716-4d33-8170-0f6136663bfd" Language="1033" Codepage="1252" Version="1.1.0">
+        <Package Id="*" Keywords="Concordium Installer" Description="Concordium Node 1.1.0 Installer" Manufacturer="Concordium Software" InstallerVersion="500" Languages="1033" Compressed="yes" InstallScope="perMachine" />
```
abizjak, line 4: This does not seem sustainable especially since it is not documented anywhere that this is a thing that needs to be kept in sync.

I think this either needs to be changed so that it is instantiated by the build job, or the version should be removed.

If it is kept, it should be documented in build instructions that this needs to be instantiated with a specific version.

service/windows/installer/Node.wxs
```diff
@@ -1,12 +1,12 @@
...
         <Media Id="1" Cabinet="Node.cab" EmbedCab="yes" />
         <Directory Id="TARGETDIR" Name="SourceDir">
             <Directory Id="ProgramFiles64Folder" Name="PFiles">
                 <Directory Id="Concordium" Name="Concordium">
-                    <Directory Id="INSTALLDIR" Name="Node 1.0">
+                    <Directory Id="INSTALLDIR" Name="Node 1.1.0">
```
abizjak, line 9: And the same here.

abizjak (changes requested):

service/windows/installer/Node.wxs
td202: I have added some documentation of this. Updating this version number and refreshing the GUID on new releases is something that should be done manually as part of the release process. Perhaps we should have a document that describes the necessary steps for making a new release, in which case this should be included.

I do not agree that this should be set by the build job, as setting this should be part of the release process. To make this more sustainable, I have reduced the instances to a single variable.

service/windows/installer/Node.wxs
td202: Simplified with a WiX variable.

service/windows/installer/Node.wxs
abizjak: Ok, that's fine. But the instructions for building the package should be such that somebody other than you can reliably make a release and build a version.

Can you please list the manual steps that need to be done to achieve this in the README, naming concretely what should be done (e.g., update versionNumber when conditions P and Q apply).

abizjak (approved):

Issue #166, opened 2021-09-16
Regression: Expired transactions are not rejected.


abizjak:
**Bug Description**

Already expired transactions are not rejected.

This is a regression introduced by #1.

**Steps to Reproduce**

Send an expired transaction to the node.

**Expected Result**

The node rejects and does not retransmit the transaction.

**Actual Result**

The transaction is accepted and later on deleted when the baker bakes a block (or by the purge mechanism) since it is expired.

**Versions**
- Software Version: 1.0.2+

abizjak:
Fixed by #167

PR #167, opened 2021-09-16
Fix regression and reject already expired transactions.

abizjak:
This regression was introduced by #1.

## Purpose

Address #166 

## Changes

Reintroduce the check for transaction expiry when receiving transactions.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -471,7 +471,7 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
 --   * 'ResultSuccess' if the transaction is freshly added.
 --   * 'ResultDuplicate', which indicates that either the transaction is a duplicate
 --   * 'ResultStale' which indicates that a transaction with the same sender
---     and nonce has already been finalized. In this case the transaction is not added to the table.
+--     and nonce has already been finalized, or the transaction has already expired. In this case the transaction is not added to the table.
 --   * 'ResultInvalid' which indicates that the transaction signature was invalid.
 --   * 'ResultShutDown' which indicates that consensus was shut down, and so the transaction was not added.
 --   * 'ResultExpiryTooLate' which indicates that transaction's expiry was too far in the future and so the transaction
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -502,11 +504,10 @@ doReceiveTransaction tr slot = unlessShutDown $ do
     return ur
 
     where
-          isExpiryTooLate = do
+          isExpiryTooLate now = do
             maxTimeToExpiry <- rpMaxTimeToExpiry <$> getRuntimeParameters
-            now <- utcTimeToTransactionTime <$> currentTime
             let expiry = msgExpiry tr
-            return $ expiry > maxTimeToExpiry + fromIntegral now
+            return $ expiry > maxTimeToExpiry + fromIntegral (utcTimeToTransactionTime now)
```
td202, line 510: `fromIntegral` should not be required.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -481,8 +481,10 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
 doReceiveTransaction :: (TreeStateMonad pv m, TimeMonad m, SkovQueryMonad pv m) => BlockItem -> Slot -> m UpdateResult
 doReceiveTransaction tr slot = unlessShutDown $ do
   -- Don't accept the transaction if its expiry time is too far in the future
-  expiryTooLate <- isExpiryTooLate
+  now <- currentTime
+  expiryTooLate <- isExpiryTooLate now
   if expiryTooLate then return ResultExpiryTooLate
```
td202, line 486:
```suggestion
  expiryTooLate <- isExpiryTooLate now
  maxTimeToExpiry <- rpMaxTimeToExpiry <$> getRuntimeParameters
  if msgExpiry tr > maxTimeToExpiry + utcTimeToTransactionTime now then return ResultExpiryTooLate
```
Removing the `where` clause.

abizjak: I like that the function is named so I do not agree it should be inlined.

concordium-consensus/src/Concordium/Skov/Update.hs
abizjak: Fixed.

td202 (approved):

PR #174, opened 2021-09-27
Fix mac node issues regarding special chars in node names

Bargsteen:
## Purpose

Fixes two related issues regarding special character handling in the mac node installer.
Previously:
- If a node name contained XML, the installation would succeed, but the node wouldn't run, as it would (likely) get a parse error on the service file.
- If a node name contained special characters such as `/`, which has a special meaning for `sed`, the installation could fail.

Also adds a limit to the node names, as the collector-backend, by default, won't accept names longer than 100 bytes.

## Changes

- Restrict the allowed characters to a-z, A-Z, -, _, and spaces (not allowed as pre/suffixes).
- Restrict the length of node name to 100.
- Update error message accordingly.
- Explain limitations for names in the service file to aid advanced users

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
+++ b/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
@@ -40,7 +40,7 @@
                 // Initialize the dialog.
                 [warning addButtonWithTitle:@"OK"];
                 [warning setMessageText:@"Invalid node names"];
-                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and can only contain a-z, A-Z, 0-9, '-', or '_'."];
+                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and can only contain a-z, A-Z, 0-9, spaces, '-', or '_'."];
                 [warning setAlertStyle:NSAlertStyleInformational];
                 
                 // Display the warning dialog.
@@ -102,12 +102,12 @@
 
 // Checks that both the mainnet and testnet node names:
 // - have length > 0 and <= 100
-// - only contain characters in [a-zA-Z0-9-_]
+// - only contain characters in [a-zA-Z0-9-_ ]
 - (BOOL) nodeNamesAreValid
 {
     // NB: This character set must be accepted by the service file XML parser.
     // This naturally excludes '<', '>', etc., but the service won't start with a sequence of ',.' in the node name either.
-    NSCharacterSet *allowedChars = [[NSCharacterSet characterSetWithCharactersInString:@"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-_"] invertedSet];
+    NSCharacterSet *allowedChars = [[NSCharacterSet characterSetWithCharactersInString:@"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-_ "] invertedSet];
 
     // Default maximum node name length that the collector-backend will accept. Measured in UTF-8 encoded bytes.
     int maxLen = 100;
--- a/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
+++ b/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
@@ -40,7 +40,7 @@
                 // Initialize the dialog.
                 [warning addButtonWithTitle:@"OK"];
                 [warning setMessageText:@"Invalid node names"];
-                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and can only contain a-z, A-Z, 0-9, spaces, '-', or '_'."];
+                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and can only contain a-z, A-Z, 0-9, spaces, '-', or '_'. Additionally, names cannot start or end with spaces."];
                 [warning setAlertStyle:NSAlertStyleInformational];
                 
                 // Display the warning dialog.
@@ -103,6 +103,7 @@
 // Checks that both the mainnet and testnet node names:
 // - have length > 0 and <= 100
 // - only contain characters in [a-zA-Z0-9-_ ]
+// - cannot start or end with spaces
 - (BOOL) nodeNamesAreValid
 {
     // NB: This character set must be accepted by the service file XML parser.
@@ -123,6 +124,8 @@
             &&  mainnetNameLen <= maxLen
             &&  testnetNameLen > 0
             &&  testnetNameLen <= maxLen
+            &&  ![mainnetName hasPrefix:@" "]
+            &&  ![mainnetName hasSuffix:@" "]
             && !([mainnetName rangeOfCharacterFromSet:allowedChars].location != NSNotFound)
             && !([testnetName rangeOfCharacterFromSet:allowedChars].location != NSNotFound);
 }
--- a/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
+++ b/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
@@ -40,7 +40,7 @@
                 // Initialize the dialog.
                 [warning addButtonWithTitle:@"OK"];
                 [warning setMessageText:@"Invalid node names"];
-                [warning setInformativeText:@"Node names cannot be empty or contain '\"'."];
+                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and can only contain a-z, A-Z, 0-9, '-', or '_'."];
                 [warning setAlertStyle:NSAlertStyleInformational];
                 
                 // Display the warning dialog.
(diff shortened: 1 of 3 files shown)


scripts/distribution/macOS-package/template/scripts/postinstall
```diff
@@ -17,6 +17,27 @@ function addBinariesToPath() {
...
+# Replaces all occurrences of 'placeholder' with 'replacement' in file 'fileName'.
+# The replacement can contain any characters except for special characters in XML, which must be escaped.
+function replaceInFile() {
+    local placeholder=$1
+    local replacement=$2
+    local fileName=$3
+
+    # Other tools, such as sed and awk, has limitations to the characters allowed, so '/' and '&' could not be used.
+    # This tiny script does not have that problem.
+    python3 - <<EOF
```
abizjak, line 29: Is python3 always installed globally on macs?

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -67,26 +68,31 @@ - (void) saveConfigurationToDisk {
     os_log(customLog, "Create file: %@", configFilePath);
     BOOL createdFile = [[NSFileManager defaultManager] createFileAtPath:configFilePath contents:nil attributes:nil];
     os_log(customLog, "File-creation result: %d", createdFile);
-    
+
+    // Escape characters with special meaning in XML.
+    // The method is only available on CFStringRef, so: convert to CFStringRef, escape XML, convert back to NSString.
+    NSString *escapedMainnetName = (__bridge NSString*)CFXMLCreateStringByEscapingEntities(CFAllocatorGetDefault(), (__bridge CFStringRef)[_oMainnetNodeName stringValue], NULL);
```
abizjak, line 74: The documentation of this method states
```
Specifies the entities to be replaced. Dictionary keys should be the entity names (for example, “para” for &para;), and the values should be CFString objects containing the expansion. Pass NULL to indicate no entities other than the standard five.
```

What are the "standard five"?

scripts/distribution/macOS-package/template/scripts/postinstall
EmilLa1: I believe python 2.7.x is installed by default. Python 3 should probably be fetched from homebrew etc.

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
EmilLa1: My guess would be something along the lines of: T = String | Number | Bool | T | [T]   ....I cannot find a source though.

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -102,16 +108,11 @@ - (void) saveConfigurationToDisk {
...
 - (BOOL) nodeNamesAreValid
 {
     NSString *mainnetName = [_oMainnetNodeName stringValue];
     NSString *testnetName = [_oTestnetNodeName stringValue];
-    
-    return      [mainnetName length] > 0
-            &&  [testnetName length] > 0
-            && ![mainnetName containsString:@"\""]
-            && ![testnetName containsString:@"\""];
+    return [mainnetName length] > 0 && [testnetName length] > 0;
```
EmilLa1, line 115: Should there be an upper bound?

EmilLa1: Can the _oTestnetNodeName and _oMainnetNodeName be NULL?

EmilLa1: Not directly related to this check, however should it be possible to `toggle` which type of nodes, that should be installed/enabled via the installer ?

scripts/distribution/macOS-package/template/scripts/postinstall
```diff
@@ -17,6 +17,27 @@ function addBinariesToPath() {
...
 }
 
+# Replaces all occurrences of 'placeholder' with 'replacement' in file 'fileName'.
+# The replacement can contain any characters except for special characters in XML, which must be escaped.
+function replaceInFile() {
+    local placeholder=$1
+    local replacement=$2
+    local fileName=$3
+
+    # Other tools, such as sed and awk, has limitations to the characters allowed, so '/' and '&' could not be used.
```
EmilLa1, line 27: Wouldn't sed work?
```
~ echo "//foo//" > foo.txt
~ sed -i'' -e 's/\//1/g' foo.txt
~ cat foo.txt                   
~ 11foo11
```

Bargsteen: No, because the `1` in this case is provided by the user and so it might contain a `/` or some other character which has special meaning in sed.

EmilLa1: Surely you can just escape them?

EmilLa1:
```
~ echo "xxxfooxxx" > foo.txt     
~ sed -i'' -e 's/x/\//g' foo.txt
~ cat foo.txt                    
~///foo///
```
Or what am I missing? My main objection we're dragging a python dependency into the build for something that seems to be solvable with just sed? But I might be missing your point... :-)

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
Bargsteen: According to the [six-year-old source](https://github.com/opensource-apple/CF/blob/3cc41a76b1491f50813e28a4ec09954ffa359e6f/CFXMLParser.c#L1902), the standard five are: `&` `<` `>` `\` `"`
However, I can see that `'` is also escaped, so perhaps they've expanded the list of standard characters to be escaped in a more recent version.

scripts/distribution/macOS-package/template/scripts/postinstall
abizjak: Then we should use the version that is definitely installed.

scripts/distribution/macOS-package/template/scripts/postinstall
abizjak: My recommendation is to just whitelist some characters in the name, say [a-zA-Z0-9.-, ] and check this in the installer. when the user is entering them. Then we don't have to worry about escaping anything. If they really want a fancy name with smileys they can edit the configuration file manually.

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
Bargsteen: Added upper bound 👍

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -40,7 +40,7 @@ - (BOOL)shouldExitPane:(InstallerSectionDirection)aDir
                 // Initialize the dialog.
                 [warning addButtonWithTitle:@"OK"];
                 [warning setMessageText:@"Invalid node names"];
-                [warning setInformativeText:@"Node names cannot be empty or contain '\"'."];
+                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and can only contain a-z, A-Z, 0-9, '-', or '_'."];
```
EmilLa1, line 43:
```suggestion
                [warning setInformativeText:@"Node names must be between 1 and 100 characters in length and must be a sequence of the following characters: a-z, A-Z, 0-9, '-', or '_'."];
```

EmilLa1 (approved):

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
Bargsteen: That was also how I first wrote it, but Lotte suggested the `can only contain` version ;)

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -101,17 +101,30 @@ - (void) saveConfigurationToDisk {
...
 // Checks that both the mainnet and testnet node names:
-// - have length > 0
-// - contain no double quotes
+// - have length > 0 and <= 100
+// - only contain characters in [a-zA-Z0-9-_]
 - (BOOL) nodeNamesAreValid
 {
+    // NB: This character set must be accepted by the service file XML parser.
+    // This naturally excludes '<', '>', etc., but the service won't start with a sequence of ',.' in the node name either.
+    NSCharacterSet *allowedChars = [[NSCharacterSet characterSetWithCharactersInString:@"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-_"] invertedSet];
```
abizjak, line 110: I would recommend adding space to this list of allowed characters since people are likely to want this.

Bargsteen: I allowed spaces.
However, the XML parser will ignore pre- and suffixes of spaces, so I disallowed that.

abizjak (approved):
Great. I like the fact you added documentation to the XML files for anybody that will update them manually.

EmilLa1 (approved):
Good job

Issue #179, opened 2021-10-03
Block parent is sometimes deemed missing


abizjak:
**Bug Description**

```
2021-10-03T12:23:07.861279649Z: INFO: Skov: Finalization statistics: finalizationCount=26585 lastFinalizedTime=1.6332637878612645e9 finalizationPeriodEMA=Just 1.4626803686410766e-2 finalizationPeriodEMSD=Just 4.948750310804371e-2
2021-10-03T12:23:07.861645200Z: DEBUG: Skov: Block 5533cb484c6ffa4a09060dad06aeac8a50fa30cb04fd4ae901fba8b30684764b marked finalized
2021-10-03T12:23:07.861660720Z: ERROR: TreeState: Failed to find block d3b55ae6d0fd21e889fa3d6798da704324d355dfa54d64c3e168b7f6a3a1f5ee
2021-10-03T12:23:07.861667390Z: ERROR: TreeState: Block status: Just (BlockAlive d3b55ae6d0fd21e889fa3d6798da704324d355dfa54d64c3e168b7f6a3a1f5ee)
2021-10-03T12:23:07.861672100Z: ERROR: TreeState: Couldn't find parent block in disk
concordium-node: user error (Couldn't find parent block in disk)
concordium-node: warning: too many hs_exit()s
```
(this is on testnet during normal "start node and catch up")

**Steps to Reproduce**

This bug is not reliably reproducible. I have not seen it on my workstation at work, but trying to catch up on this machine (at home) triggers it somewhat consistently (after a few minutes of catchup, depending on what peers I get).

The bug seems to depend on 
1. the vagaries of the Haskell runtime and when it decides to run finalizers for weak pointers. We seem to be using them incorrectly since we have weak pointers to compound types, and documentation explicitly states that
```
WARNING: weak pointers to ordinary non-primitive Haskell types are particularly fragile, because the compiler is free to optimise away or duplicate the underlying data structure. Therefore attempting to place a finalizer on an ordinary Haskell type may well result in the finalizer running earlier than you expected. This is not a problem for caches and memo tables where early finalization is benign.
```
[...]

PR #180, opened 2021-10-03
Retrieve live blocks from the block table as well when trying to get block parents.

abizjak:
## Purpose

Partially addresses #179 

This addresses a sporadic node crash in a case when an unfortunate runtime
behaviour combined with the timing of finalization.

Since finalization marks blocks as finalized from most recent backwards it
follows block parent pointers backwards. As a result if the weak pointer is
finalized for some reason before all blocks are finalized it used to crash.

We have not seem many instances of this, but I have just experienced it
consistently during catchup on testnet.

## Changes

When following parent pointers, if the weak pointer cannot be de-referenced attempt to look up the block in the block table as well.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
@@ -430,11 +430,21 @@ getWeakPointer weakPtr ptrHash name = do
             lf <- use lastFinalized
             if ptrHash == getHash lf then
               return lf
-            else do
-              nb <- readBlock ptrHash
-              case nb of
-                Just sb -> constructBlock sb
-                Nothing -> logErrorAndThrowTS ("Couldn't find " ++ name ++ " block in disk")
+            else
+              -- Weak pointers are used for parent pointers. A block that is alive should always have
+              -- a parent that is also alive, which means either actually in memory `BlockAlive` or already
+              -- finalized. If we fail to dereference the weak pointer we should thus be able to directly look
+              -- up the block from the block table.
+              use (blockTable . at' ptrHash) >>=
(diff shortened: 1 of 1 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -430,11 +430,21 @@ getWeakPointer weakPtr ptrHash name = do
...
-              case nb of
-                Just sb -> constructBlock sb
-                Nothing -> logErrorAndThrowTS ("Couldn't find " ++ name ++ " block in disk")
+            else
+              -- Weak pointers are used for parent pointers. A block that is alive should always have
+              -- a parent that is also alive, which means either actually in memory `BlockAlive` or already
+              -- finalized. If we fail to dereference the weak pointer we should thus be able to directly look
+              -- up the block from the block table.
+              use (blockTable . at' ptrHash) >>=
+                 \case Just (BlockAlive bp) -> return bp
```
td202, line 439: Does this case occur sufficiently often that it is more efficient to check the block table first than to call `readBlock` first?

td202 (approved):
Seems fine. I only have a minor concern about performance, but that can be deferred.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
abizjak: I do not have the numbers unfortunately and when logging the cases it does vary how often it happens based on how it is run so I'm not really sure which option is best.

I did it this way since I assumed that lookup in the hashmap will still be  dominated by a lookup in the LMDB database.

Issue #135, opened 2021-08-16
Optimize and improve the use of LMDB database


abizjak:
**Description**

Upon each finalization the node (in persistent state configuration) currently performs at least two LMDB transactions, the first to store the finalization record via `addFinalization` and then for each block that is finalized via `markFinalized`. If there are transactions in a block an additional DB transaction is performed to mark all the contained transactions as finalized.

This has a significant effect on performance of catchup when there is low load on the chain. Block import is dominated by database operations, in particular fsync(or fdatasync) which is called on each transaction seems to dominate runtime.

With a bit of change we could reduce the number of database transactions we perform per finalization to 1, which would improve initial catchup by about 30% when most blocks are empty (this is based on benchmarks on a single machine, so the exact number may vary).

Reducing the number of DB transactions would also improve the robustness of block and transaction storage

- all blocks that are finalized would be written in a single transaction. Currently they are written by **decreasing height**, i.e., if block at height 30 is finalized and the previous finalized block is at 27 then the order of writes is block at height 30, then at 29, then at 28.
- since they are written by decreasing height it can happen that we have a block at height n but not at height n-1 if the node crashes at an unfortunate time
- since transactions are marked as finalized in a separate transaction (and by increasing block height) it can happen that all blocks and the finalization record that finalizes them are written, but transactions are not marked as such.

Some things need to be considered.
- is there a limit on transaction size in LMDB?

PR #181, opened 2021-10-04
Fix dead block access

### LINKED ISSUE #179: Block parent is sometimes deemed missing

abizjak:
## Purpose

Fix occasional crashes in high-stress situations.

In combination with #180 closes #179 

Partially addresses #135 by writing finalized blocks by increasing height.

1. For the persistent block storage block parent pointers are weak pointers.
2. During finalization blocks are in some cases marked as dead before their descendants. The node sometimes looks up a block parent that is dead.
3. Depending on the scheduling of the garbage collector the parent block is already reclaimed, or at the very least the weak pointer cannot be dereferenced, leading to a node crash.

## Changes

Instead of marking blocks as dead immediately during finalization we accumulate them and mark them as dead at the end after we no longer need to look up their parents.

This relies on #180.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -162,18 +162,29 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         -- We proceed backwards from the new finalized block, marking it and
         -- its ancestors as finalized, while other blocks at the same height
         -- are marked dead.
+        -- Instead of marking blocks dead immediately we accummulate them
+        -- and a return a list. The reason for doing this is that we never
+        -- have to look up a parent block that is already marked dead.
         let
-            pruneTrunk :: BlockPointerType m -> Branches m -> m ()
-            pruneTrunk _ Seq.Empty = return ()
-            pruneTrunk keeper (brs Seq.:|> l) = do
-                forM_ l $ \bp -> if bp == keeper then do
-                                    markFinalized (getHash bp) finRec
-                                    logEvent Skov LLDebug $ "Block " ++ show bp ++ " marked finalized"
-                                else do
-                                    markLiveBlockDead bp
-                                    logEvent Skov LLDebug $ "Block " ++ show bp ++ " marked dead"
+            pruneTrunk :: [BlockPointerType m] -- ^List of blocks to remove.
+                       -> BlockPointerType m -- ^The block that was finalized.
+                       -> Branches m
+                       -> m [BlockPointerType m]
+                       -- ^ The return value is a list of blocks to mark dead, ordered
+                       -- by increasing height.
+            pruneTrunk toRemove _ Seq.Empty = return toRemove
+            pruneTrunk toRemove keeper (brs Seq.:|> l) = do
+                toRemove1 <- foldM (\remove bp -> if bp == keeper then do
+                                      markFinalized (getHash bp) finRec
+                                      logEvent Skov LLDebug $ "Block " ++ show bp ++ " marked finalized"
+                                      return remove
+                                    else
+                                      return (bp:remove)
+                                  )
+                            toRemove
+                            l
                 parent <- bpParent keeper
-                pruneTrunk parent brs
+                toRemove2 <- pruneTrunk toRemove1 parent brs
                 -- Finalize the transactions of the surviving block.
                 -- (This is handled in order of finalization.)
                 finalizeTransactions (getHash keeper) (blockSlot keeper) (blockTransactions keeper)
@@ -184,8 +195,9 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
                         bcHeight = bpHeight keeper,
                         ..}
                 flushBlockSummaries ctx ati =<< getSpecialOutcomes =<< blockState keeper
+                return toRemove2
 
-        pruneTrunk newFinBlock (Seq.take pruneHeight oldBranches)
+        toRemoveFromTrunk <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
         -- Archive the states of blocks up to but not including the new finalized block
         let doArchive b = case compare (bpHeight b) lastFinHeight of
                 LT -> return ()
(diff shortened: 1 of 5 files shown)


concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -221,6 +241,10 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
                     _ -> return prunedbrs
         newBranches <- trimBranches unTrimmedBranches
         putBranches newBranches
+        -- mark dead blocks by increasing height
+        forM_ (toRemoveFromTrunk ++ reverse toRemoveFromBranches) $ \bp -> do
```
td202, line 245: I think they should be removed by decreasing height to avoid having a live block with a dead parent.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -162,18 +162,29 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         -- We proceed backwards from the new finalized block, marking it and
         -- its ancestors as finalized, while other blocks at the same height
         -- are marked dead.
+        -- Instead of marking blocks dead immediately we accummulate them
```
td202, line 165:
```suggestion
        -- Instead of marking blocks dead immediately we accumulate them
```

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -162,18 +162,29 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
+            pruneTrunk toRemove keeper (brs Seq.:|> l) = do
+                toRemove1 <- foldM (\remove bp -> if bp == keeper then do
+                                      markFinalized (getHash bp) finRec
+                                      logEvent Skov LLDebug $ "Block " ++ show bp ++ " marked finalized"
+                                      return remove
+                                    else
+                                      return (bp:remove)
+                                  )
+                            toRemove
+                            l
```
td202, line 185:
```suggestion
                markFinalized (getHash keeper) finRec
                logEvent Skov LLDebug $ "Block " ++ show keeper ++ " marked finalized"
                let toRemove1 = filter (/= keeper) l ++ toRemove
```

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -162,18 +162,29 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
                 parent <- bpParent keeper
-                pruneTrunk parent brs
+                toRemove2 <- pruneTrunk toRemove1 parent brs
```
td202, line 187: It might be preferable to call `markFinalized` after the recursive call to ensure that blocks are marked finalized in increasing height order. (This would also be useful for database recovery, since it would be convenient to assume that if a block is in the database then its parent will also be there.)

td202 (changes requested):

concordium-consensus/src/Concordium/Skov/Update.hs
abizjak: I think at this point it should not matter, but I don't have any objections to it either.

concordium-consensus/src/Concordium/Skov/Update.hs
abizjak: Added.

concordium-consensus/src/Concordium/Skov/Update.hs
abizjak: Changed.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -196,21 +203,29 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         doArchive =<< bpParent newFinBlock
         -- Prune the branches: mark dead any block that doesn't descend from
         -- the newly-finalized block.
+        -- Instead of marking blocks dead immediately we accummulate them
```
td202, line 206:
```suggestion
        -- Instead of marking blocks dead immediately we accumulate them
```

td202 (approved):
One typo, otherwise looks fine. If this will be a release, be sure to update `service/windows/installer/Node.wxs`.

concordium-consensus/src/Concordium/Skov/Update.hs
abizjak: Resolved.

PR #182, opened 2021-10-04
Fix incorrect lmdb env use

### LINKED ISSUE #176: Node crashes when queried at inopportune times.
**Bug Description**

When a node is queried for any data that requires the LMDB database access during finalization it will tend to crash with a corrupt database leaving only the option of deleting the database.

This happens because we are using the LMDB API somewhat incorrectly, namely http://www.lmdb.tech/doc/group__mdb.html#gaa2506ec8dab3d969b0e609cd82e619e5 `mdb_env_set_mapsize` requires that there are no outstanding database transactions when executed.

**Steps to Reproduce**

It is hard to reproduce deterministically without modifying the node, but it can be reliably reproduced by running queries `getBlocksAtHeight` + `getBlockInfo` + `getBlockSummary` in sequence by increasing block height while the node is catching up on the testnet. Likely just running `getBlocksAtHeight` would do the same.

This leads to a segmentation fault and analyzing the core dump leads to the following backtrace
```
[Current thread is 1 (Thread 0x7fb2a0b86700 (LWP 3313041))]
(gdb) backtrace
#0  0x00005644d2edd6fd in mdb_txn_renew0 ()
#1  0x00005644d2ede423 in mdb_txn_begin ()
#2  0x00007fb2a9a78d52 in ?? () at hsrc_lib/Database/LMDB/Raw.hsc:855
```

**Expected Result**

The node does not crash as a result of external queries.

**Actual Result**

The node crashes.

**Versions**
- Software Version: 1.1.2 (but I think any version starting with 1.0.0 is affected)

abizjak:
## Purpose

Closes #176 

## Changes

The cause of the issue is that we have been resizing the environment while there were outstanding active transactions.
The change in this PR is to introduce a RW lock that is used to guard access to the environment. Transactions need read access, but doing any modification to the environment itself needs write access.

Most of the code change is an implementation of a lock. Unfortunately I have not been able to find a lock that has the right fairness properties, namely that writers have priority which is what we want since we do not wish to starve consensus of progress by making repeated queries.

Depends on
- #180 
- #181 

So those should be merged before.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE
         ScopedTypeVariables,
         DefaultSignatures,
-        TypeFamilies
+        TypeFamilies,
+        TemplateHaskell
 #-}
 {- |This module provides abstractions for working with LMDB databases.
     Chief among them is the 'MDBDatabase' class, which can be used for
@@ -12,6 +13,7 @@
 module Concordium.GlobalState.LMDB.Helpers where
 
 import Control.Concurrent (runInBoundThread)
+import Control.Concurrent.MVar
 import Control.Exception
 import Control.Monad
 import Control.Monad.Trans.Except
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -28,6 +30,158 @@ import Foreign.Marshal.Utils
...
+initializeLock :: IO RWLock
+initializeLock = do
+  state <- newMVar (Free 0)
+  readLock <- newMVar ()
+  writeLock <- newMVar ()
+  return RWLock{..}
+
+-- |Acquire a read lock. This will block until there are no pending writers
+-- waiting to acquire the lock.
+acquireRead :: RWLock -> IO ()
```
EmilLa1, line 74: I suppose some of these functions are internal, could they be hidden from external modules?

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -28,6 +30,158 @@ import Foreign.Marshal.Utils
...
+        -- There is an active writer. Do nothing and wait until the writer is done.
+        putMVar state lockState
+        readMVar writeLock
+        go
+
+
+acquireWrite :: RWLock -> IO ()
+acquireWrite RWLock{..} = mask_ $ go False
+  where
+    -- the boolean flag indicates whether this is a first iteration of the loop (False) or not (True)
```
EmilLa1, line 110: Maybe add a remark of what the purpose of the boolean flag. Something along the lines of "... waitingWriters field is updated according to the the boolean flag i.e. the waitingWriters field is incremented if a new writer has arrived... " - Now you're at it (adding a comment)

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
EmilLa1: acquireRead, acquireWrite etc.

EmilLa1 (approved):
Looks fine to me - a few comments.

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
abizjak: I agree, but the only way to not export this is by listing every other function in the export list which is tedious.

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -132,67 +138,67 @@ acquireRead RWLock{..} = mask_ go
...
+              -- waitQSem to make progress.
               putMVar rwlState st
-              readMVar rwlReadLock
+              waitQSem rwlReadLock
               go
       lockState@(WriteLocked _) -> do
         -- There is an active writer. Do nothing and wait until the writer is done.
         putMVar rwlState lockState
-        readMVar rwlWriteLock
+        waitQSem rwlWriteLock
```
td202, line 155: I think this can also lead to a deadlock if this semaphore is not signalled.

abizjak: Yes, this is reverted to use MVars as it was originally.

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -28,6 +30,158 @@ import Foreign.Marshal.Utils
...
+      waitingWriters :: !Word64
+      }
+  -- |The lock is acquired by a single writer.
+  | WriteLocked {
+      waitingWriters :: !Word64
+      }
+    deriving(Show)
+
+-- |A reader writer lock with a bias for writers. If there are waiting writers
+-- readers will block.
```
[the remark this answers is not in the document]
abizjak: I documented more extensively now how the lock is meant to behave.

td202 (approved):
Looks good. Just a couple of notes on the documentation.

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -28,6 +60,212 @@ import Foreign.Marshal.Utils
...
+-- (WriteLocked n, ⊤, ⊥) -RW-> (Free n, ⊤, ⊤), WW
+--
+-- No other state should be reachable.
+data RWLock = RWLock {
+  -- |The state the lock is currently in.
+  rwlState :: !(MVar RWState),
+  -- |An MVar used to signal waiting readers to wake up. This is empty when
+  -- there is at least one active reader and full otherwise.
+  rwlReadLock :: !(MVar ()),
+  -- |An MVar used to signal waiting writers to wake up. This is empty when
```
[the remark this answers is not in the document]
abizjak: Done.

PR #193, opened 2021-10-20
Add docker build for database exporter

bisgardo:
## Purpose

As we're moving towards a multi-reporter setup, it's becoming less clear that the reporter node is the right place to run the database exporter service (used for exporting and publishing data for OOB catchup).

We're looking into performing the task in a K8s cron job that periodically starts up a node, lets it catch up, and then stops again.  Once stopped, the database exporter will run against the node's database. Also, the data may be archived and shipped off to S3 to be used to start a new node almost without catchup.

## Changes

Added Dockerfile to build a database exporter image.

## Checklist

      hard-to-understand areas.


scripts/database-exporter/publish.sh
```diff
@@ -0,0 +1,52 @@
...
+# Once the export has been generated and verified,
+# it will be uploaded to the specified AWS S3 bucket with public access
+# and the relevant entry of the specified AWS CloudFront distribution will be invalidated.
+
+# Expected environment variables:
+# - DB_PATH: Local path to the node's LMDB database (default: '/var/lib/concordium/data/database-v4')
+# - S3_BUCKET_NAME: Name of the S3 bucket (without 's3://' prefix) to which the export should be published.
+# - CF_DISTRIBUTION_ID: ID of the CloudFront distribution.
+
+db_path="${DB_PATH-/var/lib/concordium/data/database-v4}"
```
abizjak, line 15: Is this missing a `:`, i.e., `${DB_PATH:-/...}`?

bisgardo: Both syntaxes are valid. If `:` is included it will also pick the default if the variable is declared with empty value. I tend to not include it.

bisgardo:
> I'm less concerned that a short-running node would be a drain on the network. After all, the out-of-band catch-up is a benefit to the network, which I feel at least balances out the drain. However, I'm concerned that the time for the node to catch up could be quite variable (depending on the amount of transactions, for instance) and determining if the node has caught up with the network may not be entirely reliable. On that basis, I would also rather that the database exporter be run on a node that is constantly up.

I'm not against the node being constantly up, but I'm not sure I see why we should worry about the catchup time. Even if it takes a full day, the published catchup data will still be less than 2 days old. And should it ever take longer, a new job will start right after the first one finishes. Or we could just run the job more often than that.

We were discussing ways to determine if the node is caught up and thought that a reliable method is to ask prometheus for the currently finalized block height based on the metrics emitted by all of our own nodes and stop once that height is reached. It could query once when the job starts or periodically.

bisgardo:
> Catchup is slow. Even on mainnet and testnet now it takes many hours, depending on hardware (currently mostly disk speed). So it is really not feasible to start catchup from scratch every day. It is also just a waste of resources.

We do not intend to catchup from scratch every day. As the state is persisted we would start from the point when the node was last turned off.

I'll merge this for now. Let's discuss the details of now to run the node later.

PR #200, opened 2021-11-04
Earlier verification of ChainUpdates

EmilLa1:
## Purpose

Enable earlier verification checks for `ChainUpdate` transactions. 
This builds upon the work done in https://github.com/Concordium/concordium-node/pull/199
with the purpose of reducing network congestion caused by non valid transactions. 

## Changes

Now carrying out a signature validation check when receiving ChainUpdate transactions and reject / don't retransmit non verifiable ones. 

If the verification fails then we reject the transaction. 

On the other hand if the verification check succeeds then we cache this result, together with a hash of the 
`UpdateKeysCollection`. Before executing the transaction we check if the current set keys matches with the ones from the verification cache. If that is the case, then we skip the signature check and carry out the `ChainUpdate` otherwise we reject it with `IncorrectSignature`

## Checklist

      hard-to-understand areas.


EmilLa1:
Closing, changes are now at : https://github.com/Concordium/concordium-node/pull/216

PR #206, opened 2021-11-12
Drop connections on unrecognized update

### LINKED ISSUE #197: On unrecognized protocol update disconnect peers
**Description**

When an unrecognized protocol update takes effect the node currently stops consensus for everything apart from queries. It does not stop the network layer and some transactions are still retransmitted by the node, but blocks and finalization messages are not. This means that such nodes cause a disruption in the network.

The change is to shut down the network layer of the node as well as all RPC calls that change the state.
This at least involves 
- sendTransaction 
- PeerConnect

This will make such a node a better network participant.

abizjak:
## Purpose

Drop connections upon an urecognized protocol update.

Closes #197 

## Changes

- On an unrecognized protocol update drop all connections and peers, and block sending of transactions.
- Additionally fixes an issue where the node would not stop if there is an active grpc connection.
  We now have a timeout in case the rpc server does not shut down in due time.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
   `GetNextAccountNonce` by any alias.
 - `GetAccountInfo` object has an additional field `accountAddress` that contains
   the canonical address of the account.
+- The node now drops all connections on an unrecognized protocol update and
+  refuses to accept new transactions.
 
 ## concordium-node 1.1.3
 
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -189,7 +189,20 @@ async fn main() -> anyhow::Result<()> {
             if err.is_cancelled() {
                 info!("RPC server was successfully shutdown by force.");
             } else if err.is_panic() {
-                error!("RPC server panicked!");
+        // Force the rpc server to shut down in at most 10 seconds.
+        let timeout_duration = std::time::Duration::from_secs(10);
+        match tokio::time::timeout(timeout_duration, task).await {
+            Ok(res) => {
+                if let Err(err) = res {
+                    if err.is_cancelled() {
+                        info!("RPC server was successfully shutdown by force.");
+                    } else if err.is_panic() {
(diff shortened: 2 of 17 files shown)


EmilLa1 (approved):
Looks good

concordium-node/src/bin/cli.rs
```diff
@@ -185,11 +185,20 @@ async fn main() -> anyhow::Result<()> {
...
-                error!("RPC server panicked!");
+        // Force the rpc server to shut down in at most 10 seconds.
+        let timeout_duration = std::time::Duration::from_secs(10);
+        match tokio::time::timeout(timeout_duration, task).await {
+            Ok(res) => {
+                if let Err(err) = res {
+                    if err.is_cancelled() {
+                        info!("RPC server was successfully shutdown by force.");
+                    } else if err.is_panic() {
+                        error!("RPC server panicked!");
```
EmilLa1, line 196: For good measure we could log the reason.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - Faster state queries by avoiding locking the block state file when reading.
 - Fix a bug by shutting down RPC before the node, which caused the node to crash
   when attempting a graceful shutdown while processing RPC requests.
+- The node now drops all connections on an unrecognized protocol update and
+  refuses to accept new transactions.
```
limemloh, line 11: Missing a log for forcing down the rpc after 10 seconds.

limemloh (approved):
Looks good to me

CHANGELOG.md
EmilLa1: I don't think this change necessarily should be in the changelog. I wouldn't expect the rpc to be open after the node has shutdown. Also there is the other entry already. "Fix a bug by shutting down the RPC...."

abizjak: I agree since we did not release the version which had this "never terminate due to idle connections" bug. There is no reason to have two entries in the changelog for this.

PR #207, opened 2021-11-13
Improve missing config dir error

### LINKED ISSUE #198: Not very friendly error reporting in case of misconfiguration.
**Bug Description**

Bad error reporting when config directory is missing.

```
thread 'main' panicked at 'Can't write to config file '/dev/shm/testnet-base/main.config.json': No such file or directory (os error 2)', src/configuration.rs:905:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Steps to Reproduce**

Start a node, giving it a non-existent config directory.

**Expected Result**

There is a reasonable error message like there is for other configuration mistakes. In practice this means we should use anyhow and `.context` to contextualize the error and make it more palatable.

**Actual Result**

Rust panic is triggered.

**Versions**
- Software Version
- OS
- Browser
- Mobile device, if applicable

soerenbf:
## Purpose

Improve error message when node is started with missing configuration directory.

Fixes #198.

## Changes

- Provide error message with `anyhow::Context`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/configuration.rs
+++ b/concordium-node/src/configuration.rs
@@ -883,27 +883,24 @@ impl AppPreferences {
     /// Creates an `AppPreferences` object.
     pub fn new(override_conf: PathBuf, override_data: PathBuf) -> anyhow::Result<Self> {
         let file_path = Self::calculate_config_file_path(&override_conf, APP_PREFERENCES_MAIN);
-        let mut new_prefs = match OpenOptions::new().read(true).write(true).open(&file_path) {
-            Ok(file) => {
-                let mut reader = BufReader::new(&file);
-                let load_result = PreferencesMap::<String>::load_from(&mut reader);
-                let prefs = load_result.unwrap_or_else(|_| PreferencesMap::<String>::new());
+        let mut new_prefs = if let Ok(file) = OpenOptions::new().read(true).write(true).open(&file_path) {
+            let mut reader = BufReader::new(&file);
+            let load_result = PreferencesMap::<String>::load_from(&mut reader);
+            let prefs = load_result.unwrap_or_else(|_| PreferencesMap::<String>::new());
 
-                AppPreferences {
-                    preferences_map:     prefs,
-                    override_data_dir:   override_data,
-                    override_config_dir: override_conf,
-                }
-            },
-            _ => {
-                let _ = File::create(&file_path).context("Context error!")?;
-                let prefs = PreferencesMap::<String>::new();
+            AppPreferences {
+                preferences_map:     prefs,
+                override_data_dir:   override_data,
+                override_config_dir: override_conf,
+            }
+        } else {
+            let _ = File::create(&file_path).with_context(|| format!("Can't write to config file: '{}'", file_path.as_path().display()))?;
+            let prefs = PreferencesMap::<String>::new();
 
-                AppPreferences {
-                    preferences_map:     prefs,
-                    override_data_dir:   override_data,
-                    override_config_dir: override_conf,
-                }
+            AppPreferences {
+                preferences_map:     prefs,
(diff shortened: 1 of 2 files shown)


concordium-node/src/configuration.rs
```diff
@@ -881,10 +881,10 @@ pub struct AppPreferences {
 
 impl AppPreferences {
     /// Creates an `AppPreferences` object.
-    pub fn new(override_conf: PathBuf, override_data: PathBuf) -> Self {
+    pub fn new(override_conf: PathBuf, override_data: PathBuf) -> anyhow::Result<Self> {
         let file_path = Self::calculate_config_file_path(&override_conf, APP_PREFERENCES_MAIN);
-        let mut new_prefs = match OpenOptions::new().read(true).write(true).open(&file_path) {
-            Ok(file) => {
+        let mut new_prefs =
+            if let Ok(file) = OpenOptions::new().read(true).write(true).open(&file_path) {
```
abizjak, line 887: It looks like in this branch you only need read access to the file, so unless I'm missing something you could just do
```rust
if let Ok(file) = File::open(&file_path) {
...
```

concordium-node/src/configuration.rs
```diff
@@ -894,23 +894,20 @@ impl AppPreferences {
...
-                    AppPreferences {
-                        preferences_map:     prefs,
-                        override_data_dir:   override_data,
-                        override_config_dir: override_conf,
-                    }
-                }
-                Err(e) => {
-                    panic!("Can't write to config file '{}': {}", file_path.as_path().display(), e)
+            } else {
+                let _ = File::create(&file_path).with_context(|| {
```
abizjak, line 898: Perhaps it would be useful to log here, something like
```
info!("Node configuration file not found. Creating a new one.")
```

concordium-node/src/configuration.rs
```diff
@@ -894,23 +894,20 @@ impl AppPreferences {
...
+                    format!("Can't write to config file: '{}'", file_path.as_path().display())
```
abizjak, line 899:
```suggestion
                    format!("Could not create configuration file: '{}'", file_path.as_path().display())
```

concordium-node/src/configuration.rs
```diff
@@ -881,36 +881,34 @@ pub struct AppPreferences {
...
-                let mut reader = BufReader::new(&file);
-                let load_result = PreferencesMap::<String>::load_from(&mut reader);
-                let prefs = load_result.unwrap_or_else(|_| PreferencesMap::<String>::new());
-
-                AppPreferences {
-                    preferences_map:     prefs,
-                    override_data_dir:   override_data,
-                    override_config_dir: override_conf,
-                }
+        let mut new_prefs = if let Ok(file) = OpenOptions::new().open(&file_path) {
```
abizjak, line 886: This is pedantic, but the logic here is ugly since we are checking whether a file exists by trying to open it.

Instead, a cleaner logic would be something like (see https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file )
```
if file_path.is_file() {
   let file = std::fs::File::open(&file_path).context("Could not open configuration file.")?;
   ...
} else {
  // the current else branch
}
```

EmilLa1: Agree, we should take the time now to clean up this logic. 
But other than that it looks good! 👍

soerenbf: Got caught up in something else yesterday, taking a look at this now!

soerenbf: Tried doing this.

abizjak (approved):

EmilLa1 (approved):
Looks good

PR #208, opened 2021-11-16
Don't use the partial record selector to allow updating 1 -> 2 -> 3.

abizjak:
## Purpose

`genesisCore` is a partial record selector so it fails on the second update.

## Changes

Use the correct core function to extract the genesis core.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
Looks good

PR #209, opened 2021-11-17
Release node version 3.0.0.

abizjak:
## Purpose

Bump node version to 3 to match the protocol version and add the protocol specification hash.

## Changes

The only code change is the updated specification hash.


concordium-consensus/src/Concordium/ProtocolUpdate/P2/ProtocolP3.hs
```diff
@@ -71,7 +71,7 @@ import Concordium.Kontrol
 -- This is the hash of the published specification document.
 -- TODO: Write the document, insert correct hash.
```
EmilLa1, line 72: Can this todo be considered DONE?

abizjak: Yes.

EmilLa1 (approved):
Looks good

PR #210, opened 2021-11-17
Add configuration setting for "keep-alive".

### LINKED ISSUE #160: Stale connections may linger
**Bug Description**

A node may erroneously consider itself connected to a peer for an extended period after the connection was dropped.  This disallows the node from reconnecting.

**Steps to Reproduce**

- Run a node on stagenet (or a small network) so that it will connect to all available nodes.
- Interrupt the network connection (e.g. disconnect from VPN)
- Shutdown the node
- Restart the node
- Reconnect the network connection

**Expected Result**

The node should be able to reconnect to its peers within a short amount of time.

**Actual Result**

Reconnecting to peers takes >10 minutes. The peers show as connected to the node, but the node does not show as connected to them on the network dashboard.

**Versions**
- Software Version 1.1.1

soerenbf:
## Purpose

Reduce the time a potentially dead connection is kept.

Fixes #160 

#### Note
I have only tested that the relation between **housekeeping-interval** and **keep-alive** is ensured. I have not tested that the connection is indeed dropped after the time specified instead of the previous hardcoded 20 minutes.

## Changes

- Moved keep-alive constant to configuration setting
- Ensure correct relation between housekeeping-interval and keep-alive settings

## Checklist

     hard-to-understand areas.

### DIFF
--- a/concordium-node/src/configuration.rs
+++ b/concordium-node/src/configuration.rs
@@ -72,6 +72,9 @@ pub const SOFT_BAN_DURATION_SECS: u64 = 300;
 pub const MAX_PEER_NETWORKS: usize = 20;
 /// Database subdirectory name
 pub const DATABASE_SUB_DIRECTORY_NAME: &str = "database-v4";
+/// The factor "max_normal_keep_alive" needs to be greater than
+/// "housekeeping_interval" by. Decreasing this increases risk of inactive nodes
+/// being dropped prematurely.
 const HOUSE_KEEPING_INTERVALS_PER_NORMAL_KEEP_ALIVE: u8 = 3;
 
 #[cfg(feature = "database_emitter")]
@@ -512,12 +515,12 @@ pub struct ConnectionConfig {
     pub deduplication_hashing_algorithm: DeduplicationHashAlgorithm,
     #[structopt(
         long = "max-normal-keep-alive",
-        help = "Max seconds to hold connection to dead \"normal\" node before discarding. Must be \
-                at least 3 times greater than the value of \"housekeeping-interval\".",
+        help = "Max seconds to keep alive an inactive connection to a \"normal\" node before \
+                discarding.",
         default_value = "120",
         env = "CONCORDIUM_NODE_MAX_NORMAL_KEEP_ALIVE"
     )]
-    pub max_normal_keep_alive: u16,
+    pub max_normal_keep_alive: u64,
 }
 
 #[derive(StructOpt, Debug)]
@@ -867,8 +870,8 @@ pub fn parse_config() -> anyhow::Result<Config> {
     );
 
     ensure!(
-        (conf.connection.max_normal_keep_alive as u64)
-            > conf.connection.housekeeping_interval
+        conf.connection.max_normal_keep_alive
+            >= conf.connection.housekeeping_interval
                 * (HOUSE_KEEPING_INTERVALS_PER_NORMAL_KEEP_ALIVE as u64),
         "max-normal-keep-alive ({}) should be at least {} times greater than the value of \
          housekeeping-interval ({})",
--- a/concordium-node/src/configuration.rs
+++ b/concordium-node/src/configuration.rs
@@ -73,9 +73,9 @@ pub const MAX_PEER_NETWORKS: usize = 20;
 /// Database subdirectory name
 pub const DATABASE_SUB_DIRECTORY_NAME: &str = "database-v4";
 /// The factor "max_normal_keep_alive" needs to be greater than
-/// "housekeeping_interval" by. Decreasing this increases risk of inactive nodes
+/// "housekeeping_interval" by. Decreasing this increases risk of nodes
 /// being dropped prematurely.
-const HOUSE_KEEPING_INTERVALS_PER_NORMAL_KEEP_ALIVE: u8 = 3;
+const KEEP_ALIVE_FACTOR: u8 = 3;
 
 #[cfg(feature = "database_emitter")]
 #[derive(StructOpt, Debug)]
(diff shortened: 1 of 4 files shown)


concordium-node/src/configuration.rs
```diff
@@ -511,6 +509,14 @@ pub struct ConnectionConfig {
...
     )]
     pub deduplication_hashing_algorithm: DeduplicationHashAlgorithm,
+    #[structopt(
+        long = "max-normal-keep-alive",
+        help = "Max seconds to hold connection to dead \"normal\" node before discarding. Must be \
+                at least 3 times greater than the value of \"housekeeping-interval\".",
+        default_value = "120",
+        env = "CONCORDIUM_NODE_MAX_NORMAL_KEEP_ALIVE"
+    )]
+    pub max_normal_keep_alive: u64,
```
EmilLa1, line 519: Minor, but u64 is probably overkill.

concordium-node/src/configuration.rs
```diff
@@ -859,6 +865,14 @@ pub fn parse_config() -> anyhow::Result<Config> {
         "wait-until-minimum-nodes must be lower than or equal to peer-list-size"
     );
 
+    ensure!(
+        conf.connection.max_normal_keep_alive > conf.connection.housekeeping_interval * 3,
```
EmilLa1, line 869: Let the `3` be a constant and attach a comment to it, why it's 3..

concordium-node/src/p2p/connectivity.rs
```diff
@@ -564,6 +564,7 @@ pub fn connection_housekeeping(node: &Arc<P2PNode>) -> bool {
 
     let curr_stamp = get_current_stamp();
     let peer_type = node.peer_type();
+    let max_keep_alive_ms = node.config.max_normal_keep_alive * 1000;
```
EmilLa1, line 567: Why not store it in ms. instead of doing the ms -> sec conversion for every housekeeping.

soerenbf: Well.. true, could do the conversion when creating the node config, however I thought it was nice for the user to input in seconds when configuring the node.

For now I'll put the conversion in the node config.

concordium-node/src/configuration.rs
soerenbf: True.. u16 will hold ~17 hours in seconds, which will probably suffice?

td202: I would stick with u64 just for consistency with other time parameters.

concordium-node/src/configuration.rs
```diff
@@ -74,6 +72,7 @@ pub const SOFT_BAN_DURATION_SECS: u64 = 300;
 pub const MAX_PEER_NETWORKS: usize = 20;
 /// Database subdirectory name
 pub const DATABASE_SUB_DIRECTORY_NAME: &str = "database-v4";
+const HOUSE_KEEPING_INTERVALS_PER_NORMAL_KEEP_ALIVE: u8 = 3;
```
abizjak, line 75: Please add some documentation of this constant and the rationale for it.

concordium-node/src/configuration.rs
```diff
@@ -511,6 +510,14 @@ pub struct ConnectionConfig {
         env = "CONCORDIUM_NODE_CONNECTION_DEDUPLICATION_HASHING_ALGORITHM"
     )]
     pub deduplication_hashing_algorithm: DeduplicationHashAlgorithm,
+    #[structopt(
+        long = "max-normal-keep-alive",
+        help = "Max seconds to hold connection to dead \"normal\" node before discarding. Must be \
+                at least 3 times greater than the value of \"housekeeping-interval\".",
```
abizjak, line 516: I guess this 3 is the same as `HOUSE_KEEPING_INTERVALS_PER_NORMAL_KEEP_ALIVE`. Can we use that here as well?

concordium-node/src/configuration.rs
```diff
@@ -859,6 +866,17 @@ pub fn parse_config() -> anyhow::Result<Config> {
         "wait-until-minimum-nodes must be lower than or equal to peer-list-size"
     );
 
+    ensure!(
+        (conf.connection.max_normal_keep_alive as u64)
+            > conf.connection.housekeeping_interval
```
abizjak, line 871: The error and help descriptions suggest that this should be `>=`, no?

concordium-node/src/configuration.rs
```diff
@@ -511,6 +510,14 @@ pub struct ConnectionConfig {
         env = "CONCORDIUM_NODE_CONNECTION_DEDUPLICATION_HASHING_ALGORITHM"
     )]
     pub deduplication_hashing_algorithm: DeduplicationHashAlgorithm,
+    #[structopt(
+        long = "max-normal-keep-alive",
+        help = "Max seconds to hold connection to dead \"normal\" node before discarding. Must be \
```
abizjak, line 515:
```suggestion
        help = "Max seconds to keep alive an inactive connection to a \"normal\" node before discarding. Must be \
```

abizjak: You'll probably need to reformat so don't just apply this suggestion even if you agree with it. :)

concordium-node/src/configuration.rs
EmilLa1: I'm fine with this.

soerenbf: I'll stick to u64 to avoid needing to cast and increase readability of the code.

concordium-node/src/p2p/connectivity.rs
EmilLa1: Yes, I still think the node should be configured with secs.

concordium-node/src/configuration.rs
soerenbf: Wasn't possible to format text with constant. Discarded part with the "3" instead after discussion with @abizjak .

concordium-node/src/configuration.rs
soerenbf: Tried doing this. Would like feedback on whether it suffices!

EmilLa1: What does the `Normal` refer to?
How about renaming it to simply just `KEEP_ALIVE_FACTOR`.

And then use // for explaining it.
```
In order to avoid prematurely connection drops, then the 'KEEP_ALIVE_FACTOR' must be at least 3 times greater than the 'housekeeping_interval'.
```

EmilLa1: What does the Normal refer to?
How about renaming it to simply just KEEP_ALIVE_FACTOR.

And then use // for explaining it.

_In order to avoid prematurely connection drops, the 'KEEP_ALIVE_FACTOR' must be at least 3 times greater than 'housekeeping_interval'._

soerenbf: The `normal` is there, as there is is also a `MAX_BOOTSTRAP_KEEP_ALIVE`, however I didn't like the name to begin with.. I'll apply yours.

abizjak (approved):

EmilLa1 (approved):
Looks good

PR #214, opened 2021-11-19
Remove support for the legacy wire protocol version 0.

### LINKED ISSUE #190: Remove support for legacy wire protocol
**Task description**

Node version 1.1.x introduced a new wire protocol (version 1), while maintaining support for the old one (version 0) used by version 1.0.x peers. Since the protocol update is now effective on all our networks, there is no need to retain support for the version 0 wire protocol. This means that the compatibility shim can be removed.

**Sub-tasks**
- [ ] Disallow connections from peers that only support wire protocol version 0
- [ ] (Optional) disallow connections from v1.0.x peers. (Optional, since they will be disallowed by the wire protocol version implicitly.)
- [ ] Remove wire protocol shim

td202:
## Purpose

Node version 1.1.x introduced a new wire protocol (version 1), while maintaining support for the old one (version 0) used by version 1.0.x peers. Since the protocol update is now effective on all our networks, there is no need to retain support for the version 0 wire protocol. This means that the compatibility shim can be removed.

Closes #190

## Changes

- Connections on wire protocol version 0 are no longer accepted.
- The shim to convert messages between wire protocol versions 0 and 1 is removed.

## Checklist

      hard-to-understand areas.


EmilLa1 (commented):
I sympathize with this change and the PR looks good. 

But wouldn't the `version_adapter` be useful for future protocol updates which possibly updates the `wire_protocol`?

td202:
> I sympathize with this change and the PR looks good.
> 
> But wouldn't the `version_adapter` be useful for future protocol updates which possibly bumps the `wire_protocol`?

Not really. The implementation is very specific to the bump from version 0 to 1, and any future wire protocol update would be different. We might use similar functions in a similar place for a future protocol update, but I don't think keeping these ones around in anticipation would be helpful.

EmilLa1 (approved):

abizjak (approved):
Ok.

PR #218, opened 2021-11-26
Bugfix/incorrect state unrecognized protocol update

### LINKED ISSUE #144: `node_info` reports incorrect state upon an unrecognized protocol update
**Bug Description**

Upon an unrecognized protocol update the baker stops, but this is not reflected in the status.

**Steps to Reproduce**

1. Run a node, send a protocol update with an unrecognized specification hash.
2. Query `concordium-client raw GetNodeInfo`.

**Expected Result**

`Baker running` should be false since the baker has stopped at that point. Finalization committee status is reported correctly as `False`.

**Actual Result**

```
Baker ID: 0
Peer type: "Node"
Baker running: True
Consensus running: True
Consensus type: "Active"
Baker committee member: NodeInfoResponse'ACTIVE_IN_COMMITTEE
Finalization committee member: False
```

**Versions**
- Software Version: 1, 1.1

soerenbf:
## Purpose

As described in #144, when invoking GetNodeInfo after an unrecognized protocol update, it shows as still baking. These changes should fix this by introducing a new response for the scenario where consensus has been shut down.

Fixes #144.

## Changes

- Show "baker running" as "False" when consensus is shut down.
- Show "Consensus running" as "False" when consensus is shut down.
- Always show baker ID if it is available, regardless of the baker status reported.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/External.hs
```diff
@@ -1135,17 +1135,22 @@ checkIfWeAreFinalizer cptr = do
 -- Returns -1 if we are not added as a baker.
 -- Returns -2 if we are added as a baker, but not part of the baking committee yet.
 -- Returns -3 if we have keys that do not match the baker's public keys on the chain.
+-- Returns -5 if baking has stopped due to consensus being shut down.
```
td202, line 1138: Why -5 rather than -4?

concordium-node/src/rpc.rs
```diff
@@ -429,11 +429,20 @@ impl P2p for RpcServerImpl {
...
+                        consensus.is_baking()
+                    };
+
                 NodeInfoResponse {
                     node_id,
                     current_localtime,
                     peer_type,
-                    consensus_baker_running: consensus.is_baking(),
+                    consensus_baker_running,
                     consensus_running: true,
```
td202, line 446: If the consensus is shutdown, then this probably should be `false`.

concordium-node/src/rpc.rs
```diff
@@ -450,6 +459,10 @@ impl P2p for RpcServerImpl {
                         ConsensusIsInBakingCommitteeResponse::NotInCommittee => {
                             node_info_response::IsInBakingCommittee::NotInCommittee.into()
                         }
+                        ConsensusIsInBakingCommitteeResponse::ShutDown => {
+                            node_info_response::IsInBakingCommittee::AddedButNotActiveInCommittee
+                                .into()
+                        }
```
td202, line 465: It seems that this is not an accurate response. In particular, if there is no baker and the consensus is shut down, this should not be the response. Really, if the consensus is shut down then the node has no idea whether it is supposed to be in the baking committee or not, but it would generally be most accurate to reflect the status as of the last block of the chain.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -1135,17 +1135,22 @@ checkIfWeAreFinalizer cptr = do
...
 bakerIdBestBlock :: StablePtr ConsensusRunner -> IO Int64
 bakerIdBestBlock cptr = do
     (ConsensusRunner mvr) <- deRefStablePtr cptr
-    res <- runMVR Q.getBakerStatusBestBlock mvr
-    return $! case res of
-        NoBaker -> -1
-        InactiveBaker _ -> -2
-        BadKeys _ -> -3
-        ActiveBaker bid -> fromIntegral bid
+    isDown <- runMVR Q.checkIsShutDown mvr
```
td202, line 1144: It might be better to have a separate query for checking if the node is shutdown. Otherwise, extend the definition of `BakerStatus` and revise the implementation of `getBakerStatusBestBlock`. There should be minimal logic in this module, beyond what is needed to interface between Rust and Haskell.

td202 (commented):
Can we define a more detailed specification of what the node info response should be, and in what case? That might help figure out what we actually want here.

For instance, it might be useful to report the baker ID even when the node is not in the baking/finalization committee. (Since the baking credentials now explicitly specify the baker ID, this info should always be available now, though it wasn't when this API was first introduced.)

concordium-consensus/src/Concordium/External.hs
soerenbf: Yes.. good question.. I think  I had a talk with @abizjak about the solution to this, where -5 was mentioned, and I never gave it extra thought. I'll change it.

concordium-consensus/src/Concordium/External.hs
soerenbf: Makes sense. I'll revise as suggested.

concordium-consensus/src/Concordium/External.hs
abizjak: Don't blame me. `-5` was representative of "numbers not yet in use".

concordium-node/src/rpc.rs
abizjak: That was indeed the entire point of the bug, that this reports true when it is shut down.

abizjak (changes requested):
The bug as reported is specifically about the fact that the node reports baker as running when it is not.

We should not attempt to change the API in other ways. To change this API will require a feature proposal since it is an API that is used by tools and sdks, so any change in the formats should be publicized more widely.

So if you add a separate external call to isShutDown you can achieve this. The baker ID status is as of the best block, so that can stay as it was.

This response should be made more precise at some point (in the sense of having less freedom than it does now, and more information) but that requires change in the response format which, again, will require a more widely available proposal.

concordium-consensus/src/Concordium/External.hs
soerenbf: > Don't blame me. `-5` was representative of "numbers not yet in use".

Can't argue with that.

soerenbf:
> Can we define a more detailed specification of what the node info response should be, and in what case? That might help figure out what we actually want here.
> 
> For instance, it might be useful to report the baker ID even when the node is not in the baking/finalization committee. (Since the baking credentials now explicitly specify the baker ID, this info should always be available now, though it wasn't when this API was first introduced.)

I changed it so the baker ID is always in the response if available.

soerenbf:
> The bug as reported is specifically about the fact that the node reports baker as running when it is not.
> 
> We should not attempt to change the API in other ways. To change this API will require a feature proposal since it is an API that is used by tools and sdks, so any change in the formats should be publicized more widely.
> 
> So if you add a separate external call to isShutDown you can achieve this. The baker ID status is as of the best block, so that can stay as it was.
> 
> This response should be made more precise at some point (in the sense of having less freedom than it does now, and more information) but that requires change in the response format which, again, will require a more widely available proposal.

As we discussed, I've changed to response to always include the baker ID if it is available. Other than that, it should now state that both consensus and baker is not running when consensus has been shut down.

After the changes the response after an unrecognized protocol update should be the following:

```
Node ID: "0000000000000000"
Current local time: 1638279083
Baker ID: 0
Peer type: "Node"
Baker running: False
Consensus running: False
Consensus type: "Active"
Baker committee member: NodeInfoResponse'ACTIVE_IN_COMMITTEE
Finalization committee member: False
```

concordium-node/src/rpc.rs
soerenbf: Removed the response type.

soerenbf:
Updated description of PR, as a few changes have been made since the last revision.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,37 @@ impl ConsensusContainer {
...
-    pub fn in_baking_committee(&self) -> ConsensusIsInBakingCommitteeResponse {
-        wrap_c_committee_call!(self, |consensus| bakerIdBestBlock(consensus))
+    pub fn in_baking_committee(&self) -> (ConsensusIsInBakingCommitteeResponse, Option<u64>) {
+        let consensus = self.consensus.load(Ordering::SeqCst);
+
+        unsafe {
+            let mut baker_id: u64 = 0;
+            let mut has_baker_id: u8 = 0;
+
+            let result = bakerStatusBestBlock(consensus, &mut baker_id, &mut has_baker_id);
```
abizjak, line 724: Only the foreign call needs to be unsafe. We should strive to have the smallest possible scope for the unsafe blocks.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,37 @@ impl ConsensusContainer {
...
+            let status =
+                ConsensusIsInBakingCommitteeResponse::try_from(result).unwrap_or_else(|code| {
+                    panic!("Unknown Consensus Baking Committee FFI return code: {}", code)
```
abizjak, line 727: I'm not in favor of panics like this. Since this function is only used in rpc as far as I can tell I think a better approach would be to forward the error and then have the RPC respond with "InternalServerError ( error description )".

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,37 @@ impl ConsensusContainer {
...
+                });
+
+            let baker_id_option = match has_baker_id {
+                0 => None,
+                1 => Some(baker_id),
+                code => panic!("Expected value from FFI to be either 0 or 1, received {}", code),
```
abizjak, line 733: And here. Alternatively you could just define non-zero => Some(baker_id).

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,37 @@ impl ConsensusContainer {
...
+
+            (status, baker_id_option)
+        }
     }
 
     pub fn in_finalization_committee(&self) -> bool {
         wrap_c_bool_call!(self, |consensus| checkIfWeAreFinalizer(consensus))
     }
 
+    pub fn is_running(&self) -> bool {
```
abizjak, line 744: Please document what this is meant to be doing. `is_running` does not really say what it is that is supposed to be running.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,37 @@ impl ConsensusContainer {
         ))
     }
 
-    pub fn in_baking_committee(&self) -> ConsensusIsInBakingCommitteeResponse {
-        wrap_c_committee_call!(self, |consensus| bakerIdBestBlock(consensus))
+    pub fn in_baking_committee(&self) -> (ConsensusIsInBakingCommitteeResponse, Option<u64>) {
```
abizjak, line 717: Please document the response, i.e., what is the meaning of the Option.

concordium-node/src/consensus_ffi/ffi.rs
abizjak: If you want to keep the panic I would recommend using `unreachable!` instead since that documents the reason better.

soerenbf: Went with `unreachable!`

concordium-node/src/consensus_ffi/ffi.rs
soerenbf: tried giving it a better name as well.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,39 @@ impl ConsensusContainer {
...
+        let consensus = self.consensus.load(Ordering::SeqCst);
+        let mut baker_id: u64 = 0;
+        let mut has_baker_id: u8 = 0;
+
+        let result = unsafe { bakerStatusBestBlock(consensus, &mut baker_id, &mut has_baker_id) };
+
+        let status =
+            ConsensusIsInBakingCommitteeResponse::try_from(result).unwrap_or_else(|code| {
+                unreachable!("Unknown Consensus Baking Committee FFI return code: {}", code)
```
abizjak, line 728: THis is a minor nitpick, but if this case ever happens, the way the errors are phrased it will print something like
```
Unknown Consensus Baking Committee FFI return code: Unsupported FFI return code for committee status ({})
```

abizjak (approved):

concordium-node/src/consensus_ffi/ffi.rs
soerenbf: Good catch, changed it to something less redundant.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -1131,22 +1131,37 @@ checkIfWeAreFinalizer cptr = do
...
+-- Returns 1 if we are not added as a baker.
+-- Returns 2 if we are added as a baker, but not part of the baking committee yet.
+-- Returns 3 if we have keys that do not match the baker's public keys on the chain.
+-- Returns 0 if we are part of the baking committee. 
+bakerStatusBestBlock :: StablePtr ConsensusRunner -> Ptr Word64 -> Ptr Word8 -> IO Word8
+bakerStatusBestBlock cptr bakerIdPtr hasBakerIdPtr = do
+    (ConsensusRunner mvr) <- deRefStablePtr cptr
+    (bs, mBid) <- runMVR Q.getBakerStatusBestBlock mvr
+    case mBid of
+        Nothing -> return $! getBakerStatusCode bs
```
td202, line 1153: I think it would also be wise to `poke hasBakerIdPtr 0` here. Also document the meaning of this value.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -709,14 +714,38 @@ impl ConsensusContainer {
...
+
+        let status = ConsensusIsInBakingCommitteeResponse::try_from(result).unwrap_or_else(|err| {
+            unreachable!("An error occured when trying to convert FFI return code: {}", err)
+        });
+
+        let baker_id_option = match has_baker_id {
+            0 => None,
+            1 => Some(baker_id),
+            code => unreachable!("Expected value from FFI to be either 0 or 1, received {}", code),
+        };
```
td202, line 734: Why not just treat 0 as false and anything else as true?

concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -575,34 +575,38 @@ getTransactionStatusInBlock trHash blockHash =
...
-      ActiveBaker !BakerId
+      ActiveBaker
     | -- |The account has a baker, but it is not yet in the committee
-      InactiveBaker !BakerId
+      InactiveBaker
     | -- |The baker id does not correspond with a current baker
       NoBaker
     | -- |The baker may exist, but the keys do not match
-      BadKeys !BakerId
+      BadKeys
```
td202, line 591: Can we match the names between the Haskell and Rust versions?

td202 (approved):
A couple of small changes. Otherwise fine.

concordium-node/src/consensus_ffi/ffi.rs
soerenbf: Well, I took the logic from the macro that wraps ffi bool return calls, where this was the way it was done. And I also think it's nice to throw an error, since clearly someone has misunderstood the purpose of the value, if anything other than 0/1 is used.

concordium-consensus/src/Concordium/Queries.hs
soerenbf: Yes, that would make sense.

concordium-node/src/consensus_ffi/ffi.rs
td202: In that case, can you update the documentation on `bakerStatusBestBlock` to make the values explicit.

PR #219, opened 2021-11-29
Handle existing service files when upgrading

Bargsteen:
## Purpose

Keep existing service files when upgrading the node.
Update: Also sets the configuration options to match those of the previous install (report to network dashboard, launch on boot).

## Changes

- The preinstall script only removes specific files in `/Library/Concordium Node` instead of the whole folder.
   - It thereby keeps the service files and files created by the user.
- The node name field is hidden during installation if the collector service file exists.
   - In its place, an explanation about how to reconfigure the name is shown.
   - It is checked separately for mainnet and testnet.
- The installer will not replace existing service files.
- Sets the configuration options to match those of the previous installation.
   - Delete the launch daemon symlinks in `/Library/LaunchDaemons/` during preinstall (a minor bug, which caused problems for the configuration options check). 

## Checklist

      hard-to-understand areas.


scripts/distribution/macOS-package/template/scripts/preinstall
```diff
@@ -34,19 +34,19 @@ function removeLaunchDaemons(){
...
+function removeFilesInNodeDir(){
+    local nodeDir="/Library/Concordium Node"
+    if [ -d "$nodeDir" ]
     then
-        echo "-- Deleting '/Library/Concordium Node/'"
-        sudo rm -r "/Library/Concordium Node/"
+        echo "-- Deleting the following files from '$nodeDir':"
+        removeFileIfExists "$nodeDir/concordium-node"
+        removeFileIfExists "$nodeDir/node-collector"
+        removeFileIfExists "$nodeDir/REPORT_TO_NETWORK_DASHBOARD_MAINNET"
```
EmilLa1, line 48: Is this a user configuration? Or what does it mean?
Do we re-create these on every installation? Shouldn't they be untouched if the user do not want 
to report the the network dashboard? 

Alternatively, they could be called  DONT_REPORT_TO_NETWORK_DASHBOARD_MAINNET. 
Then a `blank` installation wouldn't have to create the file and subsequent installations shouldn't bother with adding nor removing these files.

EmilLa1 (commented):
Looks good. 
I think it (the new behaviour) should have a small entry in the changelog. 
Like: "The macOS installer no longer overrides user data".

scripts/distribution/macOS-package/template/distribution.xml
```diff
@@ -20,7 +20,7 @@ function installCheck() {
     }
     if(system.files.fileExistsAtPath('/Library/Concordium Node/')) {
         my.result.title = 'Previous Installation Detected';
-        my.result.message = 'A previous installation of Concordium Node exists at /Library/Concordium Node/. This installer will remove the previous installation prior to installing. Your keys and the chain data will remain untouched.';
+        my.result.message = 'A previous installation of Concordium Node exists at /Library/Concordium Node/. This installer will remove some files from the previous installation prior to installing. Your keys, chain data, and service files will remain untouched.';
```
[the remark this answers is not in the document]
abizjak: I agree, @Bargsteen please be more precise or at least explain what the files it will remove are for.

scripts/distribution/macOS-package/template/scripts/preinstall
abizjak: If this file exists then the collector is started to report to the network dashboard.

abizjak (approved):
Overall I think it is OK, but I have no way of testing since it is a mac thing.

scripts/distribution/macOS-package/template/scripts/preinstall
Bargsteen: Yes, the collector is only started when the file exists.

Reporting to the dashboard is an option in the installer. That's why I remove the old setting, i.e., the old file if it exists.

abizjak (approved):

scripts/distribution/macOS-package/template/scripts/preinstall
EmilLa1: Yes. I forgot it was possible to toggle this behaviour during the installation process.

EmilLa1 (approved):

abizjak:
Will you please merge this @Bargsteen ?

Bargsteen:
Update: Sets the configuration options to match those of the previous installation (as suggested by @EmilLa1 ;) )

scripts/distribution/macOS-package/template/scripts/preinstall
```diff
@@ -22,8 +22,14 @@ function removeServiceIfExists() {
 }
 
 function removeSymlinks(){
+    echo "Removing symlinks:"
+
     removeFileIfExists "/usr/local/bin/concordium-node"
     removeFileIfExists "/usr/local/bin/concordium-node-collector"
+    removeFileIfExists "/Library/LaunchDaemons/software.concordium.mainnet.node.plist"
```
EmilLa1, line 29: 1st. this is not a symlink?
2nd. should the configurations be removed?

Bargsteen: 1. These *are* symlinks to actual files that exist in `/Library/Concordium Node/LaunchDaemons/`.
  - macOS will launch any services that exist in `/Library/LaunchDaemons/` on boot. So if you choose to launch the mainnet service on boot, the `postinstall` script will make a symlink from `/Library/LaunchDaemons/software.concordium.mainnet.node.plist -> /Library/Concordium Node/LaunchDaemons/software.concordium.mainnet.node.plist`
2. We could let the `postinstall` script remove or add them based on the configuration options you chose in the installer. But I thought it would be cleaner to simply remove all of them in `preinstall` and then create the necessary ones in `postinstall`.

PR #221, opened 2021-11-29
Cache module interfaces on start.

### LINKED ISSUE #220: Module interfaces are not cached on startup
Module interfaces (specifically the code to run) are not cached on startup from existing state (or rather, to be precise, the reference from the contract instance points to an uncached module interface, interfaces are cached in the module table).
This means that for each execution of a smart contract we need to load the interface from disk which can be slow.

I don't have measurements on how much performance is affected so I don't know how to classify the priority and severity. Module interfaces are relatively straightforward to deserialize, it is only the reading from the file that is of concern.

At the same time contract state is cached, and generally most of the rest of the state is cached in memory. And since module interfaces are already in memory (since they are cached as part of the module table), there is no reason why the pointers from the instances should not be cached.

abizjak:
## Purpose

Fixes #220 

## Changes

Cache module interfaces on startup.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/Instances.hs
```diff
@@ -119,10 +119,10 @@ instance MonadBlobStore m => BlobStorable m PersistentInstance where
 
 instance MonadBlobStore m => Cacheable m PersistentInstance where
     cache p@PersistentInstance{..} = do
-        -- TODO: We do not currently cache the pinstanceCachedParameters.
-        -- This behaviour is probably fine.
+        -- we only cache parameters and the interface. The rest is already in memory at this point.
```
EmilLa1, line 122: For the unenlightened ones it could be beneficial to add a small comment saying why only these 
two fields of the contract are required to be cached.

EmilLa1 (approved):
Looks good.

concordium-consensus/src/Concordium/GlobalState/Persistent/Instances.hs
abizjak: I attempted to clarify the comment.

EmilLa1 (approved):
Looks good.

PR #229, opened 2021-12-27
Replace panics triggered by malformed requests by graceful handling.

### LINKED ISSUE #228: Invalid queries can cause the node tasks to panic
**Bug Description**

If a query is made with an invalid input (concretely a string containing a null character) the task that handles it panics.
This is fortunately not very severe since tokio gracefully handles the panic and remains alive, responding to further queries.
But it is nevertheless an issue that should be fixed and this input problem should be handled correctly.

**Steps to Reproduce**

Make a query which expects a string as input (e.g., any consensus query that requires a block hash) with a malformed block hash with a 0 byte in the string.

**Expected Result**

Invalid request or similar error is returned.

**Actual Result**

The task handling the request panics and the connection is terminated.

**Versions**
- Software Version: All existing versions.

abizjak:
## Purpose

Fix #228 

## Changes

- Handle invalid input by responding with InvalidArgument status code.
- Remove needless copies of byte buffers when passing them to Haskell. This is not directly related to the issue above, but it is in the same place in the code and so I noticed it.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 - The macOS installer no longer overwrites the service files when reinstalling.
 - Cache smart contract modules on startup from existing state to improve smart
   contract execution.
+- Make consensus queries more robust, by validating input more extensively.
+  This affects all queries whose input was a block or transaction hash.
+  These queries now return `InvalidArgument` error, as opposed to `Unknown`
+  which they returned previously.
   
 ## concordium-node 3.0.1
 
--- a/concordium-node/src/consensus_ffi/ffi.rs
+++ b/concordium-node/src/consensus_ffi/ffi.rs
@@ -571,12 +571,12 @@ impl ConsensusContainer {
         wrap_c_call_string!(self, consensus, |consensus| getConsensusStatus(consensus))
     }
 
-    pub fn get_block_info(&self, block_hash: &str) -> String {
-        let c_str = CString::new(block_hash).unwrap();
-        wrap_c_call_string!(self, consensus, |consensus| getBlockInfo(
+    pub fn get_block_info(&self, block_hash: &str) -> anyhow::Result<String> {
+        let c_str = CString::new(block_hash)?;
+        Ok(wrap_c_call_string!(self, consensus, |consensus| getBlockInfo(
             consensus,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - The macOS installer no longer overwrites the service files when reinstalling.
 - Cache smart contract modules on startup from existing state to improve smart
   contract execution.
+- Make consensus queries more robust, by validating input more extensively.
```
EmilLa1, line 11: Maybe add the GRPC code that invalid ones will return i.e., `InvalidArgument`?

EmilLa1 (commented):
Should we validate more? Etc. the length of the block hashes submitted?

abizjak:
> Should we validate more? Etc. the length of the block hashes submitted?

This is validated by Haskell functions, e.g., `decodeBlockHash` fails if the hash is incorrect length.

We could change this setup so that we would not be allocating CStrings and instead validated everything Rust-side and just passed a *const u8 to Haskell, which could assume that the data is valid.

But that is a bigger refactoring, this is just a bugfix.

EmilLa1 (approved):

CHANGELOG.md
abizjak: Done.

td202 (approved):
One suggestion (see comment). Otherwise seems fine.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -571,12 +561,12 @@ impl ConsensusContainer {
...
     }
 
-    pub fn get_block_info(&self, block_hash: &str) -> String {
-        let c_str = CString::new(block_hash).unwrap();
-        wrap_c_call_string!(self, consensus, |consensus| getBlockInfo(
+    pub fn get_block_info(&self, block_hash: &str) -> anyhow::Result<String> {
+        let c_str = CString::new(block_hash)?;
+        Ok(wrap_c_call_string!(self, consensus, |consensus| getBlockInfo(
             consensus,
             c_str.as_ptr() as *const u8
```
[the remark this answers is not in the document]
abizjak: Changed.

PR #235, opened 2022-01-07
Added build dependency management through nix-shell

mkmks:
## Purpose

This pull request adds a Nix expression which evaluates to a shell environment containing all non-Haskell and non-Rust dependencies necessary to build `concordium-node`. It removes the need to manually install these dependencies by following the [build checklist](https://concordium.atlassian.net/wiki/spaces/GEN/pages/606666786/How-to-build+Your+Own+Node).

The major advantage of defining this environment as a Nix expression is that it becomes easily reproducible with a single command (`nix-shell`). It's useful, for example, when setting up the development environment for `concordium-node` on a new machine or to make sure that all developers in a team use exactly the same development environment.

## Changes

The added `shell.nix` file is the conventional location for package and shell environment definitions under the Nix package manager. Additionally, the `.envrc` file provides integration of said environment with [`direnv`](https://direnv.net/) and [`lorri`](https://github.com/nix-community/lorri) (a build server which removes the need to launch `nix-shell` manually when `shell.nix` changes).   

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+eval "$(lorri direnv)"
+
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,0 @@
-eval "$(lorri direnv)"
-
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ hie.yaml
 **/**/TAGS
 .DS_Store
 /concordium-node/*.log
+.envrc
 
 # MacOS distribution
 /scripts/distribution/macOS-package/tools
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,18 @@
+let
+  # 'stack --nix' will look for a GHC in the user channels
+  # we don't want a GHC from an unknown user channel,
+  # but a GHC from a known, build environment-specific channel
+  # thus, please use 'stack --no-nix --system-ghc' instead
+  pkgs2105 = import (builtins.fetchTarball "channel:nixos-21.05") {};
+  pkgs2111 = import (builtins.fetchTarball "channel:nixos-21.11") {};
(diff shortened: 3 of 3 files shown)


.envrc
```diff
@@ -0,0 +1,2 @@
+eval "$(lorri direnv)"
```
abizjak, line 1: This will break for anybody who uses direnv without nix.

abizjak (changes requested):
I am against adding this at present. The simple reason that only one person can be expected to maintain it is enough.

We used to have a nix person and these things in the repos, they need constant maintenance, and seeing that there is exactly one person for whom this is useful, it should be in their local environment.

By the same principle I could add a bash script for each of
- install dependencies on ubuntu 20.04
- install dependencies on ubuntu 20.10
- install dependencies on fedora ...

And it would be a pain to maintain like this will be. And it is impossible to test without setting up nix-shell or nixos or similar.

.envrc
mkmks: True, if anyone uses direnv configurations that aren't part of this repo. If they do, they still have a choice of not enabling this `.envrc` with `direnv allow`. Direnv configuration is opt-in for exactly this reason.

However, I'm not insisting that any of the above should be included in the repo. I'm simply sharing mine, with the idea in mind that others might find it useful too.

mkmks:
I agree that having to maintain separate scripts to install dependencies on a variety of Linux distros (+ on MacOS) is going to be a burden. Which is exactly one of the reasons Nix became somewhat popular (at least in the Haskell community) – it allows for cross-platform reproducible builds.

Perhaps, the need for that isn't currently strong in this project but there's a broader trend towards cross-platform reproducible builds not only for production deployments but also for development environments. I must say this topic interests me.

PR #242, opened 2022-01-31
Finalization persistence in a single LMDB transaction

### LINKED ISSUE #135: Optimize and improve the use of LMDB database

mkmks:
Combined persisting finalization records, finalized block state and finalized
transaction statuses into a single LMDB transaction.

Closes #135

## Purpose

At the moment, there are no less than two LMDB write transactions happening on each finalization: one to store the finalization record and one for for each block being marked as finalized. Additionally, block transaction statuses might be updated too, in a separate LMDB write transaction. If a node shuts down before all these transactions are committed, the node state becomes inconsistent which might prevent the node to start again.

## Changes

I added the `writeFinalizationComposite` function in `Concordium.GlobalState.LMDB.Helpers` to write the finalization record, blocks marked as finalized and block transaction statuses in a single LMDB transaction. I also moved the LMDB write effects outside the `pruneTrunk` function (in `processFinalization`), to make it possible to call `writeFinalizationComposite` without relying on monadic state to collect its arguments.

I also added the `wrapupFinalization` operation in `TreeState` which is defined through `writeFinalizationComposite` in the persistent implementation and as a no-op in the in-memory implementation.

A more elegant and generic solution would be to come up with a composition operation on LMDB transactions. I'm not sure if now would be a good time to work on one since the only part of the node that requires LMDB transaction composition is finalization.

<s>Since I added a new operation to `TreeState`, I also had to give it a `Paired` implementation. I'm sure I didn't do the right thing there, and I'll gladly take comments on that.</s> Fixed thanks to Thomas!

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -159,22 +159,24 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         -- We proceed backwards from the new finalized block, collecting blocks
         -- to mark as dead. When we exhaust the branches we then mark blocks as finalized
         -- by increasing height.
-        -- Instead of marking blocks dead immediately we accumulate them
-        -- and a return a list. The reason for doing this is that we never
-        -- have to look up a parent block that is already marked dead.
+        -- 
+        -- Instead of marking blocks dead immediately, we accumulate them in a list by decreasing
+        -- height. The reason for doing this is that we never have to look up a parent block that is
+        -- already marked dead.
         let pruneTrunk :: [BlockPointerType m] -- ^List of blocks to remove.
-                       -> BlockPointerType m -- ^The block that was finalized.
+                       -> [BlockPointerType m] -- ^List of finalized blocks.
                        -> Branches m
                        -> m ([BlockPointerType m], [BlockPointerType m])
-                       -- ^ The return value is a list of blocks to mark dead, ordered
-                       -- by increasing height.
-            pruneTrunk toRemove _ Seq.Empty = return ([], toRemove)
-            pruneTrunk toRemove keeper (brs Seq.:|> l) = do
-                parent <- bpParent keeper
-                let toRemove1 = filter (/= keeper) l ++ toRemove
-                (toFinalize, toRemove2) <- pruneTrunk toRemove1 parent brs
-                return (keeper : toFinalize, toRemove2)
-        (toFinalize, toRemoveFromTrunk) <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
+                       -- ^ The return value is a list of blocks to mark dead (ordered by decreasing
+                       -- height) and a list of blocks to mark finalized (ordered by increasing
+                       -- height).
+            pruneTrunk toRemove toFinalize Seq.Empty = return (toRemove, toFinalize)
+            pruneTrunk toRemove toFinalize (brs Seq.:|> l) = do
+              let keeper = head toFinalize
+                  toRemove1 = toRemove ++ filter (/= keeper) l
+              parent <- bpParent keeper
+              pruneTrunk toRemove1 (parent : toFinalize) brs
+        (toRemoveFromTrunk, toFinalize) <- pruneTrunk [] [newFinBlock] (Seq.take pruneHeight oldBranches)
         -- Add the finalization to the finalization list
         addFinalization newFinBlock finRec
         -- mark blocks as finalized in the order returned by `pruneTrunk`, so that blocks are marked
@@ -240,7 +242,7 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         newBranches <- trimBranches unTrimmedBranches
         putBranches newBranches
         -- mark dead blocks by decreasing height
-        forM_ (toRemoveFromBranches ++ reverse toRemoveFromTrunk) $ \bp -> do
+        forM_ (toRemoveFromBranches ++ toRemoveFromTrunk) $ \bp -> do
           markLiveBlockDead bp
           logEvent Skov LLDebug $ "Block " ++ show (bpHash bp) ++ " marked dead"
         -- purge pending blocks with slot numbers predating the last finalized slot
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -10,6 +10,7 @@ import Data.Foldable
 import Data.List (intercalate)
 import GHC.Stack
 import Data.Maybe (fromMaybe)
+import Data.Time (diffUTCTime)
 
 import Concordium.Types
 import Concordium.Types.Accounts
@@ -139,6 +140,7 @@ processFinalization :: forall pv m. (TreeStateMonad pv m, SkovMonad pv m, OnSkov
 processFinalization newFinBlock finRec@FinalizationRecord{..} = do
     nextFinIx <- getNextFinalizationIndex
     when (nextFinIx == finalizationIndex) $ do
+        startTime <- currentTime
         -- We actually have a new block to finalize.
         logEvent Skov LLInfo $ "Block " ++ show (bpHash newFinBlock) ++ " is finalized at height " ++ show (theBlockHeight $ bpHeight newFinBlock) ++ " with finalization delta=" ++ show finalizationDelay
         updateFinalizationStatistics
@@ -249,6 +251,8 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         -- purge pending blocks with slot numbers predating the last finalized slot
         purgePending
         onFinalize finRec newFinBlock
+        endTime <- currentTime
+        logEvent Skov LLDebug $ "Processed finalization in " ++ show (diffUTCTime endTime startTime)
 
 -- |Try to add a block to the tree.  
 -- Besides taking the `PendingBlock` this function takes a list
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -165,18 +165,19 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         -- already marked dead.
         let pruneTrunk :: [BlockPointerType m] -- ^List of blocks to remove.
                        -> [BlockPointerType m] -- ^List of finalized blocks.
+                       -> BlockPointerType m -- ^Finalized block to consider now.
+                                            -- At the same height as the latest branch in the next argument
                        -> Branches m
                        -> m ([BlockPointerType m], [BlockPointerType m])
                        -- ^ The return value is a list of blocks to mark dead (ordered by decreasing
                        -- height) and a list of blocks to mark finalized (ordered by increasing
                        -- height).
-            pruneTrunk toRemove toFinalize Seq.Empty = return (toRemove, toFinalize)
-            pruneTrunk toRemove toFinalize (brs Seq.:|> l) = do
-              let keeper = head toFinalize
-                  toRemove1 = toRemove ++ filter (/= keeper) l
+            pruneTrunk toRemove toFinalize _ Seq.Empty = return (toRemove, toFinalize)
+            pruneTrunk toRemove toFinalize keeper (brs Seq.:|> l) = do
+              let toRemove1 = toRemove ++ filter (/= keeper) l
               parent <- bpParent keeper
-              pruneTrunk toRemove1 (parent : toFinalize) brs
-        (toRemoveFromTrunk, toFinalize) <- pruneTrunk [] [newFinBlock] (Seq.take pruneHeight oldBranches)
+              pruneTrunk toRemove1 (keeper : toFinalize) parent brs
+        (toRemoveFromTrunk, toFinalize) <- pruneTrunk [] [] newFinBlock (Seq.take pruneHeight oldBranches)
         -- Add the finalization to the finalization list
         addFinalization newFinBlock finRec
         -- mark blocks as finalized in the order returned by `pruneTrunk`, so that blocks are marked
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -7,6 +7,7 @@ import Control.Monad
 import qualified Data.Sequence as Seq
 import Lens.Micro.Platform
 import Data.Foldable
+import Data.List (intercalate)
 import GHC.Stack
 import Data.Maybe (fromMaybe)
 
@@ -147,9 +148,6 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
         -- This is to ensure that the focus block is always a live (or finalized) block.
         unless focusBlockSurvives $ updateFocusBlockTo newFinBlock
         lastFinHeight <- getLastFinalizedHeight
-        -- Add the finalization to the finalization list
-        -- TODO: The way this is stored will probably change.
-        addFinalization newFinBlock finRec
         -- Prune the branches, which consist of all the non-finalized blocks
         -- grouped by block height.
         oldBranches <- getBranches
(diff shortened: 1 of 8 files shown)


abizjak (commented):
I think the changes overall are fine, but before we can merge this we need to understand limitations and do some tests.

- Is there a limit on LMDB transaction sizes, and what is. What happens if it is reached?
- How well does this work when 100 blocks are finalized at a time in bad conditions, e.g., 100 full blocks are finalized at the same time (with current parameters this would be about 300MB of data, so this is probably fine, but it would be useful to have reassurance).

The latter can easily be tested either by the deterministic runner by using sufficiently many transactions and finalization minimum skip of 100 (or more of course). I would ideally like to see this in a running node as well.

Lastly, I would like some benchmarks of how much this improves catchup performance. DB transactions are at present a significant proportion of catchup time, since most blocks are empty. It would be good to measure before and after catchup of, say 10000 blocks.

You can use the execute-chain runner for that with either mainnet or testnet catchup file truncated to a reasonable size (the testnet catchup file is at https://catchup.testnet.concordium.com/blocks_to_import.mdb , the mainnet one at https://catchup.mainnet.concordium.software/blocks_to_import.mdb )

A simple way of truncating is just to truncate the file (e.g., cut it down to N MB) for the desired number of blocks to import.

abizjak:
Also, it is not useful to spend time on abstracting since we have one example.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -165,34 +162,41 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
-                        ..}
-                flushBlockSummaries ctx ati =<< getSpecialOutcomes =<< blockState keeper
-                return toRemove2
-
-        toRemoveFromTrunk <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
+                (toFinalize, toRemove2) <- pruneTrunk toRemove1 parent brs
+                return (keeper : toFinalize, toRemove2)
+        (toFinalize, toRemoveFromTrunk) <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
+        -- Add the finalization to the finalization list
+        -- TODO: The way this is stored will probably change.
```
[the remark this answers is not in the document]
mkmks: This TODO is from February 2020. I kept when refactoring `pruneTrunk`, but I'll remove it if it isn't useful anymore.

mkmks:
> * Is there a limit on LMDB transaction sizes, and what is. What happens if it is reached?

I couldn't find a mention of a transaction size limit in LMDB, only the environment size limit. Since the LMDB environment consists of all memory-mapped data structures, it must include the data structures used to execute transactions. If the environment is too small, writes will fail with the `MDB_MAP_FULL` error code. This case seems to be already taken care of in the consensus layer (failed writes are retried after a resize).

I built both `main` and my own branch and run `deterministic-runner` and `execute-chain` from both. For benchmarking purposes, I run each 3 times and averaged the running time. My machine has a 4-core Intel i7 mobile CPU from 4 years ago and an SSD.

> * How well does this work when 100 blocks are finalized at a time in bad conditions, e.g., 100 full blocks are finalized at the same time (with current parameters this would be about 300MB of data, so this is probably fine, but it would be useful to have reassurance).

As I understood it, I'd have to implement my own event generator to try this scenario within `deterministic-runner`, which is something, I'm afraid, I didn't figure out too quickly. 1,000,000 steps with the default event generator finished in:

`main`: 53.5 +/- 0.4 s
my branch: 50.4 +/- 0.2 s

That is, a 9% running time decrease.

> Lastly, I would like some benchmarks of how much this improves catchup performance. DB transactions are at present a significant proportion of catchup time, since most blocks are empty. It would be good to measure before and after catchup of, say 10000 blocks.

`execute-chain` processed the mainnet catchup file was truncated to 5M (10135 blocks) in:

`main`: 373 +/- 1 s
my branch: 355 +/- 1 s

That is, a 5% running time decrease.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -165,34 +162,40 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
-                        bcHash = getHash keeper,
-                        bcHeight = bpHeight keeper,
-                        ..}
-                flushBlockSummaries ctx ati =<< getSpecialOutcomes =<< blockState keeper
-                return toRemove2
-
-        toRemoveFromTrunk <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
+                (toFinalize, toRemove2) <- pruneTrunk toRemove1 parent brs
+                return (keeper : toFinalize, toRemove2)
+        (toFinalize, toRemoveFromTrunk) <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
```
abizjak, line 177: If I'm understanding this correctly then `toFinalize` is a list of blocks "by decreasing height".

`finalizeTransactions` assumes that all transactions prior to the block it is processing are finalized. Thus this will go very badly.

Either reverse the `toFinalize` list here, or construct it in reverse order via an accumulator.

If you do the latter, you can also construct `toRemoveFromTrunk` in reverse order, avoiding reversal below, and making `pruneTrunk` tail recursive.

abizjak: And the comment of pruneTrunk should probably be updated to reflect the new reality.

abizjak (changes requested):
I tested and benchmarked it now (to test I needed to fix the issue I pointed out, namely blocks are now being marked as finalized in the wrong order, which crashes the node).

The benchmark I did was to import 10MB of testnet blocks both on main and in this branch, and both with the database in memory (that is, located in /dev/shm) and on disk. The results are in-line with what I expected. Assuming the database in `/dev/shm` is the best we can do (and that is reflected in 100% CPU utilization), the `single-transaction-finaliation` branch is roughly in the middle between the reference and the `main` branch. This fits the expectation since I am importing primarily empty blocks, so previously there were on average 2 fsync calls per finalization, now there is 1.

```
* Database on disk (main)
	User time (seconds): 77.61
	System time (seconds): 1.69
	Percent of CPU this job got: 55%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 2:22.20
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 147560
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 24
	Minor (reclaiming a frame) page faults: 21605
	Voluntary context switches: 240181
	Involuntary context switches: 79320
	Swaps: 0
	File system inputs: 0
	File system outputs: 2294952
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

* Database in memory (/dev/shm) (main)
  	User time (seconds): 74.14
	System time (seconds): 0.63
	Percent of CPU this job got: 100%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:14.19
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 147760
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 1
	Minor (reclaiming a frame) page faults: 22397
	Voluntary context switches: 72522
	Involuntary context switches: 8178
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

* Database on disk (single-transaction-finalization)
	User time (seconds): 75.27
	System time (seconds): 1.31
	Percent of CPU this job got: 70%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:49.07
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 148032
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 4
	Minor (reclaiming a frame) page faults: 22925
	Voluntary context switches: 160696
	Involuntary context switches: 40065
	Swaps: 0
	File system inputs: 0
	File system outputs: 1745456
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0

* Database in memory (/dev/shm) (single-transaction-finalization)
	User time (seconds): 73.99
	System time (seconds): 0.54
	Percent of CPU this job got: 100%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:13.95
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 147896
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 1
	Minor (reclaiming a frame) page faults: 21546
	Voluntary context switches: 68454
	Involuntary context switches: 7466
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

concordium-consensus/src/Concordium/Skov/Update.hs
mkmks: I believe that `toFinalize` was already constructed in __increasing__ order but, indeed, it's easier to see that if `pruneTrunk` uses an accumulator to construct it: `parent` is always prepended before `toFinalize`.

I also reversed the order in which `toRemove` is constructed, it's in decreasing order now.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -161,37 +159,44 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
-                return toRemove2
-
-        toRemoveFromTrunk <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
+                       -> m ([BlockPointerType m], [BlockPointerType m])
+                       -- ^ The return value is a list of blocks to mark dead and a list of blocks
+                       -- to mark finalized, ordered by increasing height.
+            pruneTrunk toRemove toFinalize Seq.Empty = return (toRemove, toFinalize)
+            pruneTrunk toRemove toFinalize (brs Seq.:|> l) = do
+              let keeper = head toFinalize
+                  toRemove1 = toRemove ++ filter (/= keeper) l
```
abizjak, line 175: Since we are traversing the branches by decreasing height, I believe this should be as it was, so
```
filter (/= keeper) l ++ toRemove
```
since that will add blocks with lower height at the beginning of the list.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -161,37 +159,44 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
-                        bcHash = getHash keeper,
-                        bcHeight = bpHeight keeper,
-                        ..}
-                flushBlockSummaries ctx ati =<< getSpecialOutcomes =<< blockState keeper
-                return toRemove2
-
-        toRemoveFromTrunk <- pruneTrunk [] newFinBlock (Seq.take pruneHeight oldBranches)
+                       -> m ([BlockPointerType m], [BlockPointerType m])
+                       -- ^ The return value is a list of blocks to mark dead and a list of blocks
+                       -- to mark finalized, ordered by increasing height.
```
abizjak, line 171: I believe this returns the list of blocks to remove ordered by decreasing height. Which is also what is expected below.

So either keep the comment as is and change `toRemove1` back to what it was
```
toRemove1 = filter (/= keeper) l ++ toRemove
```
and then reverse the list below
```
reverse toRemoveFromTrunk
```

Or fix the comment.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -161,37 +159,46 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
+          ati <- bpTransactionAffectSummaries block
+          bcTime <- getSlotTimestamp (blockSlot block)
+          let ctx = BlockContext{
+                bcHash = getHash block,
+                bcHeight = bpHeight block,
+                ..}
+          flushBlockSummaries ctx ati =<< getSpecialOutcomes =<< blockState block
+        -- block states and transaction statuses need to be added into the same LMDB transaction
+        -- with the finalization record, if persistent tree state is used
+        wrapupFinalization mfs fts
```
abizjak, line 200: Is there a reason this function does not take a finalization record and instead looks it up?

I believe passing it here explicitly would be conceptually cleaner and harder to screw up.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -161,37 +159,46 @@ processFinalization newFinBlock finRec@FinalizationRecord{..} = do
...
+        -- Finalize the transactions of surviving blocks in the order of their finalization.
+        fts <- forM toFinalize $ \block -> do
+          finalizeTransactions (getHash block) (blockSlot block) (blockTransactions block)
+        forM_ toFinalize $ \block -> do
+          ati <- bpTransactionAffectSummaries block
+          bcTime <- getSlotTimestamp (blockSlot block)
+          let ctx = BlockContext{
+                bcHash = getHash block,
+                bcHeight = bpHeight block,
+                ..}
```
abizjak, line 196: Is there a reason you are iterating `toFinalize` three times instead of once, doing all three things for each of the blocks?

You could also change the type of wrapupFinalization to take a list of pairs to make this even simpler.

concordium-consensus/src/Concordium/Skov/Update.hs
mkmks: At one point, I was thinking to pass lists of blocks and transactions to be finalized through the tree state, just like the last finalization record already was. Then I figured that it's probably not optimal because these lists would only be used by `wrapupFinalization`. But since the last finalization record was part of the (persistent) tree state anyway, I could just look it up.

Since `wrapupFinalization` is only doing anything in its persistent implementation, I also wanted to make its signature "quieter" (having fewer parameters). But of course, passing a parameter explicitly makes the data flow more obvious than passing it through the state. If you think it's better that way, I'll make it explicit.

abizjak: I would prefer it, but it is your decision.

concordium-consensus/src/Concordium/Skov/Update.hs
mkmks: I guess the reason was exactly that: I made `wrapupFinalization` take a pair of lists and not a list of pairs, which required the two lists to to be produced independently of each other. And the persistent implementation of `wrapupFinalization` is no more than a wrapper around `writeFinalizationComposite` which I also made to take a pair of lists and not a list of pairs.

So, yeah, there isn't a special reason to do that. I'll change it.

abizjak (approved):

PR #266, opened 2022-04-05
Node collector keep trying

rimbi:
## Purpose

Removes the parameter for maximum allowed times a gRPC call can fail and keeps collector querying forever.

## Changes

Remove the parameter and the relevant logic to terminate the app.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/bin/collector.rs
+++ b/concordium-node/src/bin/collector.rs
@@ -3,14 +3,7 @@ use collector_backend::{IsInBakingCommittee, NodeInfo};
 use concordium_node::utils::setup_macos_logger;
 use concordium_node::{common::grpc_api, req_with_auth, utils::setup_logger};
 use serde_json::Value;
-use std::{
-    borrow::ToOwned,
-    fmt,
-    process::exit,
-    str::FromStr,
-    thread,
-    time::Duration,
-};
+use std::{borrow::ToOwned, fmt, process::exit, str::FromStr, thread, time::Duration};
 use structopt::StructOpt;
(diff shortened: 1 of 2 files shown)


concordium-node/src/bin/collector.rs
```diff
@@ -188,18 +169,12 @@ async fn main() {
...
                     );
                 }
             }
-
-            if grpc_failure_count + 1 >= conf.max_grpc_failures_allowed as usize {
-                error!("Too many gRPC failures, exiting!");
-                exit(1);
-            }
         }
         trace!("Sleeping for {} ms", conf.collector_interval);
```
abizjak, line 179: I know this is from before already, but I think because we use `tokio`, it would be better to use the [Interval](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html) instead of `std::thread::sleep`.

concordium-node/src/bin/collector.rs
```diff
@@ -162,11 +146,8 @@ async fn main() {
         thread::sleep(Duration::from_millis(conf.artificial_start_delay));
```
abizjak, line 146: I know this is from before already, but I think because we use `tokio`, it would be better to use the [`tokio::sleep`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html) instead of thread::sleep. The former works better in combination with async futures.

abizjak (commented):
Could you also add a line to the changelog?

abizjak:
Can you please rebase your fix on top of main instead?

abizjak (approved):
Please update the changelog before merging.

Issue #245, opened 2022-02-11
Collector hangs after error


td202:
**Bug Description**

The collector hangs after an error sending the payload to the collector backend.

(From SUPPORT-1047)
```
Feb 08 19:56:53 host systemd[1]: Stopping Concordium Mainnet Node Collector...
Feb 01 13:12:59 host concordium-mainnet-node-collector-3.0.1[163781]: 2022-02-01T12:12:59.871117255Z: INFO: Collecting node information via gRPC from host/http://localhost:10000/rpcadmin
Feb 01 13:12:54 host concordium-mainnet-node-collector-3.0.1[163781]: 2022-02-01T12:12:54.869352026Z: ERROR: Error sending payload to collector backend due to "error sending request for url (https://dashboard.mainnet.concordium.software/nodes/post): connection closed before message completed"
Feb 01 13:12:49 host concordium-mainnet-node-collector-3.0.1[163781]: 2022-02-01T12:12:49.815110006Z: INFO: Collecting node information via gRPC from host/http://localhost:10000/rpcadmin
Feb 01 13:12:44 host concordium-mainnet-node-collector-3.0.1[163781]: 2022-02-01T12:12:44.742820925Z: INFO: Collecting node information via gRPC from host/http://localhost:10000/rpcadmin
```

**Steps to Reproduce**

**Expected Result**

**Actual Result**

**Versions**
- Software Version
- OS
- Browser
- Mobile device, if applicable

PR #268, opened 2022-04-05
A probable fix for the issue #245: Collector hangs after error

rimbi:
## Purpose

To prevent node-collecter from hanging when the node does not respond to gRPC calls.

## Changes

Sets the timeout for gRPC calls to 30secs.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/bin/collector.rs
+++ b/concordium-node/src/bin/collector.rs
@@ -111,6 +111,13 @@ struct ConfigCli {
     )]
     pub artificial_start_delay: u64,
     #[structopt(
+        long = "grpc-timeout",
+        help = "Time (in seconds) for gRPC request timeouts",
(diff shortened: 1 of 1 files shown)


concordium-node/src/bin/collector.rs
```diff
@@ -23,6 +23,9 @@ use std::alloc::System;
 #[global_allocator]
 static A: System = System;
 
+// The timeout for the RPC calls.
+const GRPC_TIMEOUT: Duration = Duration::from_secs(30);
```
abizjak, line 27: This does look plausible, but you you please make this a configurable option with 30s default, like the other options we have?

abizjak (approved):
Looks good and might fix the exact issue.

PR #269, opened 2022-04-06
Fixes #259: Expose foundation account address

### LINKED ISSUE #259: Expose foundation account address
**Description**

There is currently no _good_ way to get the current foundation account's address.
It would be nice if there was a way to get it.

Context: The desktop wallet displays the current foundation account, when updating the foundation account, however it does that by extracting the address from a special event in the BlockSummary. Furthermore it seems that this event will be removed in the delegation update.

(The BlockSummary exposes the index of foundation account, but not the address, and it doesn't seem possible to get the address from the index)

rimbi:
## Purpose

Adds the option of querying AccountInfo by using account index value.

## Changes

Modified the functions to query the account info such that they can also use account index as input.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
   which they returned previously.
 - Fix issue #244: Collector to keep querying. Remove the parameter for maximum allowed
   times a gRPC call can fail and keeps `node-collector` querying forever.
+- Fix the issue #259: Expose foundation account address. The account information now can
+  be queried by using an account index as well.
 
 ## concordium-node 3.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,7 @@
   which they returned previously.
 - Fix issue #244: Collector to keep querying. Remove the parameter for maximum allowed
   times a gRPC call can fail and keeps `node-collector` querying forever.
-- Fix the issue #259: Expose foundation account address. The account information now can
-  be queried by using an account index as well.
+- `GetAccountInfo` endpoint supports querying the account via the account index.
 
 ## concordium-node 3.0.1
 
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -891,16 +891,16 @@ decodeBlockHash blockcstr = readMaybe <$> peekCString blockcstr
 decodeAccountAddress :: CString -> IO (Either String AccountAddress)
 decodeAccountAddress acctstr = addressFromBytes <$> BS.packCString acctstr
 
--- |Decode a null-terminated string as either an account address (base-58), account index (Int) or a
+-- |Decode a null-terminated string as either an account address (base-58), account index (AccountIndex) or a
 -- credential registration ID (base-16).
 decodeAccountAddressOrCredId :: CString -> IO (Maybe AccountIdentifier)
 decodeAccountAddressOrCredId str = do
     bs <- BS.packCString str
     return $ case addressFromBytes bs of
-        Left _ -> Just $ 
+        Left _ ->
             case bsDeserializeBase16 bs of
-                Nothing -> AI $ read (BS.unpack bs)
-                Just cid -> CID cid
+                Nothing -> AI <$> readMaybe (BS.unpack bs)
+                Just cid -> Just $ CID cid
         Right acc -> Just $ AA acc
 
 -- |Decode an instance address from a null-terminated JSON-encoded string.
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -29,7 +29,6 @@ import qualified Data.FixedByteString as FBS
 import Concordium.Afgjort.Finalize.Types (FinalizationInstance (FinalizationInstance))
 import Concordium.Birk.Bake
 import Concordium.Constants.Time (defaultEarlyBlockThreshold, defaultMaxBakingDelay)
-import Concordium.Crypto.ByteStringHelpers
 import Concordium.GlobalState
 import Concordium.GlobalState.Persistent.LMDB (addDatabaseVersion)
 import Concordium.GlobalState.Persistent.TreeState (InitException (..))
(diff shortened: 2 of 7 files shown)


concordium-consensus/src/Concordium/External.hs
```diff
@@ -891,14 +891,17 @@ decodeBlockHash blockcstr = readMaybe <$> peekCString blockcstr
 decodeAccountAddress :: CString -> IO (Either String AccountAddress)
 decodeAccountAddress acctstr = addressFromBytes <$> BS.packCString acctstr
 
--- |Decode a null-terminated string as either an account address (base-58) or a
+-- |Decode a null-terminated string as either an account address (base-58), account index (Int) or a
 -- credential registration ID (base-16).
-decodeAccountAddressOrCredId :: CString -> IO (Maybe (Either CredentialRegistrationID AccountAddress))
+decodeAccountAddressOrCredId :: CString -> IO (Maybe AccountIdentifier)
```
abizjak, line 896: What is "AccountIdentifier"?

concordium-consensus/src/Concordium/External.hs
```diff
@@ -891,14 +891,17 @@ decodeBlockHash blockcstr = readMaybe <$> peekCString blockcstr
 decodeAccountAddress :: CString -> IO (Either String AccountAddress)
 decodeAccountAddress acctstr = addressFromBytes <$> BS.packCString acctstr
 
--- |Decode a null-terminated string as either an account address (base-58) or a
+-- |Decode a null-terminated string as either an account address (base-58), account index (Int) or a
```
abizjak, line 894: Account index is not an `Int`. There is a type `AccountIndex` defined in concordium-base (and it is defined as a wrapper around Word64).

concordium-consensus/src/Concordium/External.hs
```diff
@@ -891,14 +891,17 @@ decodeBlockHash blockcstr = readMaybe <$> peekCString blockcstr
...
-decodeAccountAddressOrCredId :: CString -> IO (Maybe (Either CredentialRegistrationID AccountAddress))
+decodeAccountAddressOrCredId :: CString -> IO (Maybe AccountIdentifier)
 decodeAccountAddressOrCredId str = do
     bs <- BS.packCString str
     return $ case addressFromBytes bs of
-        Left _ -> Left <$> bsDeserializeBase16 bs
-        Right acc -> Just $ Right acc
+        Left _ -> Just $ 
+            case bsDeserializeBase16 bs of
+                Nothing -> AI $ read (BS.unpack bs)
```
abizjak, line 902: `read` is not a safe function, it can cause an exception. Use [readMaybe](https://hackage.haskell.org/package/base-4.16.1.0/docs/Text-Read.html#v:readMaybe) instead.

concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -459,19 +459,21 @@ getModuleList :: BlockHash -> MVR gsconf finconf (Maybe [ModuleRef])
...
     MVR gsconf finconf (Maybe AccountInfo)
 getAccountInfo blockHash acct =
     join
         <$> liftSkovQueryBlock
             ( \bp -> do
                 bs <- blockState bp
-                macc <- either (BS.getAccountByCredId bs) (BS.getAccount bs) acct
+                macc <- case acct of AA addr -> BS.getAccount bs addr
+                                     AI idx -> BS.getAccountByIndex bs idx
+                                     CID crid -> BS.getAccountByCredId bs crid
```
abizjak, line 476:
```suggestion
                macc <- case acct of 
                                AA addr -> BS.getAccount bs addr
                                AI idx -> BS.getAccountByIndex bs idx
                                CID crid -> BS.getAccountByCredId bs crid
```
Usually the cases would not start on the same line as the case statement.

abizjak (changes requested):

abizjak:
@rimbi please also link the relevant issue so that it will be closed as a result of merging this PR.

concordium-consensus/src/Concordium/External.hs
rimbi: You are right. I actually defined it using `AccountIndex` in the `AccountIdentifier`.

concordium-consensus/src/Concordium/External.hs
rimbi: It is defined in `consensus-base`, forgot to push it. coming...

concordium-consensus/src/Concordium/External.hs
rimbi: :thumbsup:

abizjak:
@rimbi Please also update the changelog.

rimbi:
@abizjak PRs staring with `Fixes #XXX` should automatically close the relevant issue.

concordium-consensus/src/Concordium/External.hs
td202: This should probably be renamed `decodeAccountIdentifier`.

rimbi: done 👍

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@
   which they returned previously.
 - Fix issue #244: Collector to keep querying. Remove the parameter for maximum allowed
   times a gRPC call can fail and keeps `node-collector` querying forever.
+- Fix the issue #259: Expose foundation account address. The account information now can
+  be queried by using an account index as well.
```
abizjak, line 28:
```suggestion
- `GetAccountInfo` endpoint supports querying the account via the account index.
```

I missed that in the previous commit you did, but if you look at the changelogs we have, they do not mention specific github issues. SO please keep them like that.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -891,14 +891,17 @@ decodeBlockHash blockcstr = readMaybe <$> peekCString blockcstr
 decodeAccountAddress :: CString -> IO (Either String AccountAddress)
 decodeAccountAddress acctstr = addressFromBytes <$> BS.packCString acctstr
 
--- |Decode a null-terminated string as either an account address (base-58) or a
+-- |Decode a null-terminated string as either an account address (base-58), account index (AccountIndex) or a
 -- credential registration ID (base-16).
-decodeAccountAddressOrCredId :: CString -> IO (Maybe (Either CredentialRegistrationID AccountAddress))
```
abizjak, line 896: I don't think this function is needed anymore so should be removed.

rimbi: It was actually, just did not push it, was waiting for the merge of the change in `concordium-base`.

abizjak:
> @abizjak PRs staring with `Fixes #XXX` should automatically close the relevant issue.

Are you sure about this? Because no issues are linked. Whereas if you add `Fixes #...` or `Resolves #...` to the body then it is automatically linked.

abizjak (approved):

abizjak:
> > @abizjak PRs staring with `Fixes #XXX` should automatically close the relevant issue.
> 
> Are you sure about this? Because no issues are linked. Whereas if you add `Fixes #...` or `Resolves #...` to the PR description then it is automatically linked.

You named it `Fixes the issue #...` instead of `Fixes #...` which is probably the reason for this.

rimbi:
Yeah, that might be true.

PR #271, opened 2022-04-13
Fixes #162: Remove the need to enter a name, when a node is not going…

### LINKED ISSUE #162: Native Mac Node: Remove the need to enter a name, when a node is not going to be run anyway.
**Description**
When running the installer for the native Mac node, the installer requires names in both Mainnet and Testnet name fields, even though one of them might not be chosen.

It is a bit confusing, that you have to enter a name for a node which is not going to be started anyway.

image

**Steps to Reproduce**
- Download installer
- Browse through the proces, until the screen with Mainnet + Testnet setup is seen.
- Deselect all checkmarks for one of the nets. Click continue.
- The installer will throw an error.

**Expected Result**
- The installer let's the user continue without an error, if all checkmarks and the name field is left empty for one net. 

**Actual Result**
- The installer gives an error, if one of the name fields are empty, even if all checkmarks of the given net is deselected.

**Versions**
- Software Version: 1.1.1
- OS: MacOS 11.4

rimbi:
… to be run anyway.

This is a fix for the native mac node installer.

## Purpose

Remove the need to enter a name, when a node is not going to be run anyway.

## Changes

Changed the relevant checks in the installer.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,8 @@
 - Fix issue #244: Collector to keep querying. Remove the parameter for maximum allowed
   times a gRPC call can fail and keeps `node-collector` querying forever.
 - `GetAccountInfo` endpoint supports querying the account via the account index.
-- Mac installer: Users now can leave one of the (but not both) net configurations empty
-  when they don't want to configure e node for it.
+- Mac installer: Users now can leave one (but not both) of the net configurations empty
+  when they don't want to configure a node for it. On the initial installation, leaving a net configuration empty means that the start/stop app shortcuts and the application support folder for that net won't be installed.
 
 ## concordium-node 3.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@
 - Fix issue #244: Collector to keep querying. Remove the parameter for maximum allowed
   times a gRPC call can fail and keeps `node-collector` querying forever.
 - `GetAccountInfo` endpoint supports querying the account via the account index.
+- Mac installer: Users now can leave one of the (but not both) net configurations empty
+  when they don't want to configure e node for it.
 
 ## concordium-node 3.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,9 @@
   times a gRPC call can fail and keeps `node-collector` querying forever.
 - `GetAccountInfo` endpoint supports querying the account via the account index.
 - Mac installer: Users now can leave one (but not both) of the net configurations empty
-  when they don't want to configure a node for it. On the initial installation, leaving a net configuration empty means that the start/stop app shortcuts and the application support folder for that net won't be installed.
+  when they don't want to configure a node for it. 
+  - On the initial installation, leaving a net configuration empty means that
+    the start/stop app shortcuts and the application support folder for that net won't be installed.
 
 ## concordium-node 3.0.1
 
--- a/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
+++ b/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
@@ -202,10 +202,14 @@
 // Handles the changes in the Mainnet options
 - (IBAction)onChangeInMainnetOptions:(id)sender {
     [self enableOrDisableMainnetNodeNameField];
+    // Disable 'Continue' button if both net configurations are empty.
+    [self setNextEnabled:(self.oMainnetNodeName.isEnabled || self.oTestnetNodeName.isEnabled)];
 }
 
 // Handles the changes in the Testnet options
 - (IBAction)onChangeInTestnetOptions:(id)sender {
     [self enableOrDisableTestnetNodeNameField];
+    // Disable 'Continue' button if both net configurations are empty.
+    [self setNextEnabled:(self.oMainnetNodeName.isEnabled || self.oTestnetNodeName.isEnabled)];
 }
 @end
--- a/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
+++ b/scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
@@ -139,11 +139,11 @@
     }
 }
 
-// Checks that both the mainnet and testnet node names:
-// - have length > 0 and <= 100
-// - only contain characters in [a-zA-Z0-9-_ ]
+// Checks that a node name:
+// - has length > 0 and <= 100 if at least one of `run on startup`, `run after install`, or `report to network dashboard` is checked
+// - only contains characters in [a-zA-Z0-9-_ ]
 // - cannot start or end with spaces
-- (BOOL) nodeNamesAreValid
+- (BOOL) nodeNameIsValid:(NSString*)nodeName :(BOOL)runOnStartup :(BOOL)runAfterInstall :(BOOL)reportToNetworkDashboard
 {
     // NB: This character set must be accepted by the service file XML parser.
     // This naturally excludes '<', '>', etc., but the service won't start with a sequence of ',.' in the node name either.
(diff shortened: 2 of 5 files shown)


Bargsteen (changes requested):
This is a good start :blush:, but it only addresses the initial part of the solution.

I think you should follow through and make the following changes/checks:
- Don't install the `/Library/Application Support/Concordium Node/{net}` folder if `{net}` (testnet/mainnet) isn't selected for installation.
- Don't install the start/stop applications for `{net}` if it isn't selected.
- Don't install the `/Library/Concordium Node/LaunchDaemons/software.concordium.{net}.node[-collector].plist` service files if `{net}` isn't selected.
- Update the `preinstall`,`postinstall` scripts to make sure they work as intended with the new changes (perhaps no changes are needed, I'm not sure atm.).
- Make sure uninstalling works as expected.
   - Installing only mainnet, then run the installer again and choose only testnet (and vice versa).
   - Try the same thing again, but with an uninstall in the middle (install mainnet, uninstall, install testnet).

And, if you haven't done so already, please read https://github.com/Concordium/concordium-node/tree/main/scripts/distribution/macOS-package#concordium-node-for-macos to get a better idea of how the mac node works :blush:

scripts/distribution/macOS-package/template/scripts/postinstall
```diff
@@ -82,35 +82,52 @@ function configureService() {
     fi
 }
 
-function configureServices () {
+# Uninstalls apps and support files related to a net.
+function uninstallNet() {
+    local netName=${1:?"Missing net name to uninstall"}
+    echo "-- Removing $netName node apps and support files"
+    rm -rf "/Applications/Concordium Node/Concordium Node Start $netName.app"
```
Bargsteen, line 89: You should check whether the folders exist before trying to delete them, as the script will fail due to `set -e` if they don't exist.

scripts/distribution/macOS-package/template/scripts/postinstall
```diff
@@ -82,35 +82,52 @@ function configureService() {
...
-        "TESTNET" \
-        "/tmp/Concordium Node/LaunchDaemons/software.concordium.testnet.node.plist" \
-        "/tmp/Concordium Node/LaunchDaemons/software.concordium.testnet.node-collector.plist" \
-        "/Library/Concordium Node/LaunchDaemons/software.concordium.testnet.node.plist" \
-        "/Library/Concordium Node/LaunchDaemons/software.concordium.testnet.node-collector.plist" \
-        "$CONCORDIUM_NODE_INSTALL_TESTNET_RUN_ON_STARTUP" \
-        "$CONCORDIUM_NODE_INSTALL_TESTNET_REPORT_TO_NETWORK_DASHBOARD" \
-        "$CONCORDIUM_NODE_INSTALL_TESTNET_RUN_AFTER_INSTALL" \
-        "$CONCORDIUM_NODE_INSTALL_TESTNET_NODE_NAME"
+    if [ -z "$CONCORDIUM_NODE_INSTALL_MAINNET_NODE_NAME" ]; then
```
Bargsteen, line 100: This does not work correctly when used reinstalling, because the name will be set to `ALREADY_CONFIGURED`.

Bargsteen, line 100: This does not work correctly when used reinstalling, because the name will be set to `ALREADY_CONFIGURED`.

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -152,24 +152,40 @@ - (BOOL) nodeNamesAreValid
     // Default maximum node name length that the collector-backend will accept. Measured in UTF-8 encoded bytes.
     int maxLen = 100;
 
+    // NB: Get the actual length in bytes if allowedChars is changed to contain non-ascii characters.
+    NSUInteger nodeNameLen = [nodeName length];
+
+    if (nodeNameLen == 0 && !(runOnStartup || runAfterInstall || reportToNetworkDashboard)) {
```
Bargsteen, line 158: I think it would be better if we disabled the entry field (so it is greyed out) when you disable all three options.
We should probably allow the field to be non-empty, so you can 
- write a node name `foo`
- disable all three options (see that `foo` is still there, but the field is _disabled_)
- enable at least one of the options again
- see that `foo` is set as the node name

Later in the process, we should then handle that the node is disabled. I believe the correct solution is to omit the `CONCORDIUM_NODE_INSTALL_{netname}_NODE_NAME` flag in the config file.

Bargsteen, line 158: I think it would be better if we disabled the entry field (so it is greyed out) when you disable all three options.
We should probably allow the field to be non-empty, so you can 
- write a node name `foo`
- disable all three options (see that `foo` is still there, but the field is _disabled_)
- enable at least one of the options again
- see that `foo` is set as the node name

Later in the process, we should then handle that the node is disabled. I believe the correct solution is to omit the `CONCORDIUM_NODE_INSTALL_{netname}_NODE_NAME` flag in the config file.

Bargsteen (changes requested):

scripts/distribution/macOS-package/template/scripts/postinstall
rimbi: Good point but there is no need to do that because
1- By the time this line executes this folder is supposed to be created 
2- `rm -rf` does not fail

scripts/distribution/macOS-package/template/scripts/postinstall
rimbi: This does. not conflict with the above implementation.

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
rimbi: The solution I have just pushed has a slightly different approach. When the field is disabled the text field somehow clears its text. This is the default behavior. To have the functionality you proposed some more effort is needed. I am taking the initiative to postpone this effort until a request comes from the product management.

Bargsteen: Okay, no problem. That is also a fine solution 😊

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
```diff
@@ -139,37 +142,70 @@ - (void) saveConfigurationToDisk {
...
 }
 
-@end
+// Checks that both the mainnet and testnet node names are valid.
+- (BOOL) nodeNamesAreValid
+{
+    BOOL mainnetNameIsValid = [self nodeNameIsValid:_oMainnetNodeName];
+    BOOL testnetNameIsValid = [self nodeNameIsValid:_oTestnetNodeName];
 
+    return mainnetNameIsValid && testnetNameIsValid;
```
Bargsteen, line 181: If you deselect all options, then this will also be `true`.
I am not sure whether this is a feature or a bug because I am not sure about the value you would get from it.
What do you think?

rimbi: Good catch. I am not sure either. But let's prevent leaving both empty.  I'll push another commit.

scripts/distribution/macOS-package/template/scripts/postinstall
rimbi: You are right. I think the installer is for installing/reconfiguring. For uninstalling the user can use the uninstaller.

scripts/distribution/macOS-package/NodeConfigurationInstallerPlugin/NodeConfigurationInstallerPlugin/MyInstallerPane.m
rimbi: done

Bargsteen:
Looks good. Just missing a changelog entry 😊

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@
 - Fix issue #244: Collector to keep querying. Remove the parameter for maximum allowed
   times a gRPC call can fail and keeps `node-collector` querying forever.
 - `GetAccountInfo` endpoint supports querying the account via the account index.
+- Mac installer: Users now can leave one of the (but not both) net configurations empty
+  when they don't want to configure e node for it.
```
Bargsteen, line 29:
```suggestion
- Mac installer: Users now can leave one (but not both) of the net configurations empty
  when they don't want to configure a node for it. On the initial installation, leaving a net configuration empty means that the start/stop app shortcuts and the application support folder for that net won't be installed.
```

Bargsteen (approved):

Issue #274, opened 2022-04-22
Node vanished from the dashboard after several days


mh-concordium:
**Bug Description**
I was running one passive node on Windows and one baker on Mac. Both nodes just vanished from the dashboard after couple of days. Both nodes were responsive after they vanished, passive one was catching up and I was able to query it. Baker kept getting rewards, seemed like everything else is ok.

Looking a bit more into the Win node, nothing was visible in the Event Viewer, last entry was just regarding starting the node service. node-collector service was also running. I saved logs for Win node. Thinking back, the only thing that possibly happened was that my laptop went to sleep mode and back.

**Steps to Reproduce**
Have a node that is running for a week.

**Expected Result**
Node should be on the dashboard.

**Actual Result**
Node vanished from the dashboard
Restarting the node seems to help, node is back on the dashboard after being restarted.

**Versions**
- Software Version: 4.0.6
- OS: Windows

Log:
[stagenet.log](https://github.com/Concordium/concordium-node/files/8538822/stagenet.log)

mh-concordium:
I wasn't able to reproduce the issue.
Node version:  4.0.9
Client version: 4.0.2

abizjak:
Most likely fixed by #275

PR #275, opened 2022-04-22
Skip missed submissions in case the collector is delayed for some time.

abizjak:
## Purpose

In case the collector is suspended, or the computer goes to sleep, the current behaviour (before this commit) would be that the collector would query and submit in quick succession to make up for the lost time. This is undesirable and not useful.

## Changes

Skip any missed submissions.

In order to prevent indefinite hangs, also add a timeout to request submission.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/bin/collector.rs
+++ b/concordium-node/src/bin/collector.rs
@@ -160,8 +160,8 @@ async fn main() {
     // in such a case, try to make up the missed ticks by submitting statistics as
     // quickly as possible.
     interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
-    #[allow(unreachable_code)]
     loop {
+        interval.tick().await;
         for (node_name, grpc_host) in conf.node_names.iter().zip(conf.grpc_hosts.iter()) {
             trace!("Processing node {}/{}", node_name, grpc_host);
             match collect_data(node_name.clone(), grpc_host.to_owned(), &conf).await {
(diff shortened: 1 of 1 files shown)


abizjak:
This possibly addresses #274 or at the very least should provide more insight into it since the collector should not just hang.

concordium-node/src/bin/collector.rs
```diff
@@ -154,16 +154,31 @@ async fn main() {
...
         for (node_name, grpc_host) in conf.node_names.iter().zip(conf.grpc_hosts.iter()) {
             trace!("Processing node {}/{}", node_name, grpc_host);
             match collect_data(node_name.clone(), grpc_host.to_owned(), &conf).await {
                 Ok(node_info) => {
                     trace!("Node data collected successfully from {}/{}", node_name, grpc_host);
                     match rmp_serde::encode::to_vec(&node_info) {
                         Ok(msgpack) => {
-                            let client = reqwest::Client::new();
+                            let client_builder = reqwest::Client::builder()
+                                .connect_timeout(Duration::from_secs(conf.collector_interval))
```
rimbi, line 173:
```suggestion
                                .connect_timeout(Duration::from_millis(conf.collector_interval))
```

concordium-node/src/bin/collector.rs
```diff
@@ -154,16 +154,31 @@ async fn main() {
...
+                                .timeout(Duration::from_secs(conf.collector_interval));
```
rimbi, line 174:
```suggestion
                                .timeout(Duration:: from_millis(conf.collector_interval));
```

rimbi (commented):
Looks good, with 2 minor probable fixes.

concordium-node/src/bin/collector.rs
abizjak: Ah, good catch. Don't know what I was thinking :P

rimbi (approved):

PR #277, opened 2022-04-26
Return BakerRemoved event when baker is removed.

eb-concordium:
## Purpose

A minor bug was found where if a baker was removed (in P4), the `BakerStakeDecreased` was returned instead of the `BakerRemoved` event. 

## Changes

* The above bug has been fixed.

## Checklist

      hard-to-understand areas.


abizjak (approved):

abizjak (approved):
Good catch.

PR #278, opened 2022-04-26
Tweak `in_baking_committee` to work around issue #261

### LINKED ISSUE #261: bakerID is shown as 0
**Bug Description**
bakerID is shown as 0 on network dashboard and raw `GetNodeInfo` but baker credentials have `bakerId: 80019`.

**Expected Result**
bakerID is according to baker credentials.

**Actual Result**
bakerID is 0.

**Versions**
- Windows
- Concordium Client: 4.0.0-delegation-test-1
- Concordium Node: 4.0.5-delegation-test-5

td202:
## Purpose

Fix #261.

## Changes

Change the return type of `in_baking_committee` to work around the fact that on Windows the `node_info` endpoint reports  baker ID being 0 (if there is a baker), even when it should be something else.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/consensus_ffi/ffi.rs
+++ b/concordium-node/src/consensus_ffi/ffi.rs
@@ -770,7 +770,11 @@ impl ConsensusContainer {
 
     /// Gets baker status of the node along with the baker ID registered in the
     /// baker credentials used, if available.
-    pub fn in_baking_committee(&self) -> (ConsensusIsInBakingCommitteeResponse, Option<u64>) {
+    ///
+    /// Note: the return type does not use an Option<u64>, which would be more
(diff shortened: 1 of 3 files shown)


abizjak (approved):
If this fixes it then I suppose it should be merged.

I would like to understand what exactly goes wrong, but I have no means of reproducing, nor really the time to do it.

soerenbf (approved):
Haven't reproduced this either, so pretty much the same as @abizjak .

td202:
I would also like to understand the cause. Although I was able to reproduce the issue, I've not managed to work out the root cause. I'd like to come back to it, but for now I think having a fix and moving on is enough.

PR #281, opened 2022-04-27
Fixes #126: Node does not stop during out of band catchup

### LINKED ISSUE #126: Node does not stop during out of band catchup
**Bug Description**

The node cannot be stopped  in a nice way when doing out-of-band catchup. It can be killed of course, but it does not respond to SIGINT to be shut down gracefully.

This is due to the way the signal handler is installed. We install the handler which terminates the main node loop upon SIGINT, we then import all blocks, and then start the main node loop. Thus the signal is handled, but it will only have effect after all the blocks have been imported.

**Steps to Reproduce**

- Start the node with --import-blocks-from.
- Press Ctrl+C during block import

**Expected Result**

The expected behaviour (from a user perspective) would be that the node shuts down in a reasonable amount of time, e.g., after importing a few more blocks.

**Actual Result**

The node finishes importing all blocks, then terminates.

**Versions**
- Software Version: 1.0.*
- OS: Linux

rimbi:
- Add `stopImportingBlocks` ffi function to set a flag to stop importing process.
- Import logic checks this flag and stops the process.

## Purpose

To stop the node promptly during out of band catchup

## Changes

- Add `stopImportingBlocks` ffi function to set a flag to stop importing process.
- Call this function in the signal handler
- Import logic checks this flag and stops the process.

## Checklist

      hard-to-understand areas.


concordium-node/src/bin/cli.rs
```diff
@@ -137,6 +119,9 @@ async fn main() -> anyhow::Result<()> {
     )?;
     info!("Consensus layer started");
 
+    // Setup task with signal handling.
+    let shutdown_signal = setup_shutdown_signal_handling(consensus.clone());
+
```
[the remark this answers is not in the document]
rimbi: I see your point. On the other hand, there is no difference between the old place or the new place of setting up signal handler as long as it comes before the point where the `rx` channel is used.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -960,8 +970,8 @@ receiveTransaction transactionBS = do
                     liftSkovUpdate vc $ Skov.receiveTransaction transaction
 
 -- |Import a block file for out-of-band catch-up.
-importBlocks :: FilePath -> MVR gsconf finconf UpdateResult
-importBlocks importFile = do
+importBlocks :: FilePath -> MVar Bool -> MVR gsconf finconf UpdateResult
+importBlocks importFile mvShouldStopImportingBlocks = do
```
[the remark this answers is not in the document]
rimbi: Good point, I have covered it with the recent push. 👍

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -334,6 +334,8 @@ data MultiVersionRunner gsconf finconf = MultiVersionRunner
       mvVersions :: !(IORef (Vec.Vector (EVersionedConfiguration gsconf finconf))),
       -- |Global write lock.
       mvWriteLock :: !(MVar ()),
+      -- |Flag to stop importing blocks.
+      mvShouldStopImportingBlocks :: !(MVar Bool),
```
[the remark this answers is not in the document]
rimbi: done 👍

CHANGELOG.md
```diff
@@ -58,12 +58,14 @@
   - synchronous contract calls
   - fallback entrypoints
   - increased smart contract module size limit, 512kB
+- Node can now be stopped during out of band catchup by using signals like SIGINT.
```
abizjak, line 61: Specifically SIGINT and SIGTERM is what is handled explicitly. The other signals have system defined behaviour.

abizjak: I suggest you are more precise here.

concordium-consensus/src/Concordium/External.hs
```diff
@@ -1398,3 +1410,4 @@ foreign export ccall checkIfRunning :: StablePtr ConsensusRunner -> IO Word8
 foreign export ccall freeCStr :: CString -> IO ()
 
 foreign export ccall importBlocks :: StablePtr ConsensusRunner -> CString -> Int64 -> IO Int64
+foreign export ccall stopImportingBlocks :: StablePtr ConsensusRunner -> IO ()
```
abizjak, line 1413: This export must be added to the `lib.def` file.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -334,6 +334,8 @@ data MultiVersionRunner gsconf finconf = MultiVersionRunner
       mvVersions :: !(IORef (Vec.Vector (EVersionedConfiguration gsconf finconf))),
       -- |Global write lock.
       mvWriteLock :: !(MVar ()),
+      -- |Flag to stop importing blocks.
```
abizjak, line 337: A bit more context would be useful.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -963,6 +971,9 @@ receiveTransaction transactionBS = do
 importBlocks :: FilePath -> MVR gsconf finconf UpdateResult
 importBlocks importFile = do
     vvec <- liftIO . readIORef =<< asks mvVersions
+    -- reset the stop flag to False, in case it was set to True
```
abizjak, line 974: Why is this the correct behaviour?

This flag is only set upon Ctrl+C/SIGINT. So how could they accidentally set this flag?

concordium-node/src/bin/cli.rs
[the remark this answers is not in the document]
abizjak: Are you sure that is correct? Isn't the default SIGINT handler going to abort the process?

I don't see a need to change the place where the handler is installed, so please do not change it, or provide a strong reason for changing it.

abizjak (changes requested):

concordium-node/src/bin/cli.rs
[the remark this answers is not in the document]
rimbi: We need `consensus` instance to get initialized. It is used in the handler. That's the reason for the place change.

concordium-consensus/src/Concordium/MultiVersion.hs
rimbi: This is a more proper approach. Consider a case where there is a need to start import process more than once. That would not be possible without resetting the flag.

concordium-consensus/src/Concordium/External.hs
rimbi: done 👍

CHANGELOG.md
rimbi: done 👍

concordium-consensus/src/Concordium/MultiVersion.hs
rimbi: done 👍

concordium-node/src/bin/cli.rs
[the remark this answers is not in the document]
abizjak: We can use a oneshot channel for this without changing the semantics.

abizjak (approved):

PR #309, opened 2022-05-04
Fix crash when active finalizer is not a baker

### LINKED ISSUE #308: Removing active finalizer causes bakers to crash
When a baker that is an active finalizer is removed, it can lead to a situation where the baker signs off on a finalization record that is included in a block after the finalizer is removed as a baker. This causes a crash in the block construction when attempting to mark the finalizer as awake, since `bsoMarkFinalizationAwakeBaker` should only be called on current-epoch bakers.

Example error log:
```
May 04 11:31:47 stagenet-aws-ireland-0-baker-0 node-entrypoint.sh[505]: [INFO  concordium_node::consensus_ffi::ffi] Baker: Won lottery in 2081141(lottery power: 1000000000000 % 5001464417983; election difficulty: 2.5e-2)
May 04 11:31:47 stagenet-aws-ireland-0-baker-0 node-entrypoint.sh[505]: [INFO  concordium_node::consensus_ffi::ffi] Runner: Exiting baker thread
May 04 11:31:47 stagenet-aws-ireland-0-baker-0 node-entrypoint.sh[505]: concordium-node: Invariant violation: unable to find baker in baker pool capital vector
May 04 11:31:47 stagenet-aws-ireland-0-baker-0 node-entrypoint.sh[505]: CallStack (from HasCallStack):
May 04 11:31:47 stagenet-aws-ireland-0-baker-0 node-entrypoint.sh[505]:   error, called at src/Concordium/GlobalState/Persistent/BlockState.hs:2222:13 in concordium-consensus-0.1.0.0-HxrDhotXrasBczA1Yhubzf:Concordium.GlobalState.Persistent.BlockState
```

td202:
## Purpose
Fix a crash bug when a finalizer that signs a finalization record is no longer a baker, and so cannot be marked as awake.

Closes #308 

## Changes

- The previous `bsoMarkFinalizationAwakeBaker` (which only tolerates marking current-epoch bakers as finalization-awake) is replaced with `bsoMarkFinalizationAwakeBakers` (which ignores any baker ids that are no longer considered bakers in the current epoch).
- `bsoMarkFinalizationAwakeBakers` is refined 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Basic/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Basic/BlockState.hs
@@ -1699,9 +1699,21 @@ instance (IsProtocolVersion pv, Monad m) => BS.BlockStateOperations (PureBlockSt
       let accrueAmountBPR bpr = bpr{PoolRewards.transactionFeesAccrued = applyAmountDelta delta (PoolRewards.transactionFeesAccrued bpr)}
       in modifyBakerPoolRewardDetailsInPoolRewards bs bid accrueAmountBPR
 
-    bsoMarkFinalizationAwakeBaker bs bid =
-      let setAwake bpr = bpr{PoolRewards.finalizationAwake = True}
-      in modifyBakerPoolRewardDetailsInPoolRewards bs bid setAwake
+    bsoMarkFinalizationAwakeBakers bs bids = do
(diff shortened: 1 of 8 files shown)


abizjak (approved):

abizjak:
I will merge this now since we need to release, but it would be really useful to add regression tests in all these bugfix cases to make sure these hard to recreate bugs are not reintroduced.

It would be great if this was part of a followup pull request.

PR #319, opened 2022-05-05
Fix regression in transaction verification logic.

### LINKED ISSUE #318: Incompatibility between node versions < 4 and 4. Node version 4 does not catch up with testnet.
**Bug Description**

This is a regression bug due to an incorrect bound check, using `>` instead of `>=`.

It means that some transactions that are processed by the node version 3 are not processed by node version 4.

This means that a node version 4 is not compatible with existing mainnet and testnet, if they have transactions with stated energy = maxBlockEnergy.

**Versions**
- Software Version: 4.*

abizjak:
## Purpose

Transactions which were previously (in node version 3.*) accepted were now rejected. Specifically
transactions that allow for maxBlockEnergy amount of energy to be used should be
allowed since they were allowed by previous protocol and node versions.

Fixes #318

## Changes

- Correct the check for maximum block energy.
- Add regression test.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/tests/consensus/ConcordiumTests/ReceiveTransactionsTest.hs
+++ b/concordium-consensus/tests/consensus/ConcordiumTests/ReceiveTransactionsTest.hs
@@ -103,6 +103,7 @@ test = do
       check resultingState results 20 ResultSuccess True -- here to check that the TransactionVerifier uses the `correct` nonce for checking validity.
       check resultingState results 21 ResultEnergyExceeded False
       check resultingState results 22 ResultDuplicateNonce False
+      check resultingState results 23 ResultSuccess True -- exactly max block energy
     specify "Credential deployments received individually" $ do
       let credentialDeploymentExpiryTime = 1596409020
           now = posixSecondsToUTCTime $ credentialDeploymentExpiryTime - 1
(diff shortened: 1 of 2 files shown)


EmilLa1 (approved):

concordium-consensus/tests/consensus/ConcordiumTests/ReceiveTransactionsTest.hs
```diff
@@ -345,6 +353,8 @@ data TestEnergyParam
   -- ^The tranasction will be provided with a too little amount of energy for
   -- the particular transaction.
   | TooMuch
+  -- ^The tranasction will be provided with exactly maximum block energy amount.
+  | MaxBlockEnergy
   -- ^The energy provided was more than what is allowed in a block,
   -- and as such the transaction will be rejected.
```
td202, line 359:
```suggestion
  -- ^The energy provided was more than what is allowed in a block,
  -- and as such the transaction will be rejected.
  | MaxBlockEnergy
  -- ^The transaction will be provided with exactly maximum block energy amount.
```

concordium-consensus/tests/consensus/ConcordiumTests/ReceiveTransactionsTest.hs
```diff
@@ -377,6 +387,7 @@ mkAccountTransaction expiry validSignature nonce validSender energyAmount =
     theCost = case energyAmount of
       TooLittle -> 0
       TooMuch -> maxBound
```
td202, line 389: Perhaps making this `maxBlockEnergy + 1` would be a better test.

td202 (approved):

PR #333, opened 2022-05-11
Updates ghc to `9.0.2` and resolver to `lts-19.6`.

rimbi:
## Purpose

Update Haskell toolchain and lib dependencies.

## Changes

Updates ghc to `9.0.2` and resolver to `lts-19.6`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unrelease changes
+
 - The `SendTransaction` function exposed via the gRPC interface now provides the caller with detailed error messages if the 
   transaction was rejected instead of just `False`. The function still returns `True` if 
   the transaction was accepted.
@@ -60,6 +61,7 @@
   - increased smart contract module size limit, 512kB
   - a number of cryptographic primitives
 - Node can now be stopped during out of band catchup by using signals, SIGINT and SIGTERM.
+- Upgrade resolver to version `lts-19.6` and ghc to version `9.0.2`.
 
 ## concordium-node 3.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,6 @@
   - increased smart contract module size limit, 512kB
   - a number of cryptographic primitives
 - Node can now be stopped during out of band catchup by using signals, SIGINT and SIGTERM.
-- Upgrade resolver to version `lts-19.6` and ghc to version `9.0.2`.
 
 ## concordium-node 3.0.1
 
--- a/concordium-consensus/stack.static.yaml
+++ b/concordium-consensus/stack.static.yaml
@@ -1,4 +1,4 @@
-resolver: lts-17.4
+resolver: lts-19.6
 
 packages:
 - .
@@ -12,4 +12,4 @@ extra-lib-dirs:
 - ./smart-contracts/lib
 
 ghc-options:
-    "$everything": -fPIC
+    "$everything": -fPIC -fsimpl-tick-factor=10000
--- a/concordium-consensus/stack.static.yaml
+++ b/concordium-consensus/stack.static.yaml
@@ -12,4 +12,5 @@ extra-lib-dirs:
 - ./smart-contracts/lib
 
 ghc-options:
+    # `simpl-tick-factor` parameter here is necessary due to a bug in the ghc: https://gitlab.haskell.org/ghc/ghc/-/issues/14637#note_413425
     "$everything": -fPIC -fsimpl-tick-factor=10000
--- a/concordium-consensus/stack.yaml
+++ b/concordium-consensus/stack.yaml
@@ -18,7 +18,7 @@
 #
 # resolver: ./custom-snapshot.yaml
 # resolver: https://example.com/snapshots/2018-01-01.yaml
-resolver: lts-17.4
+resolver: lts-19.6
 
 # User packages to be built.
 # Various formats can be used as shown in the example below.
@@ -47,6 +47,11 @@ extra-deps:
 extra-lib-dirs:
 - ../concordium-base/lib
 - ./smart-contracts/lib
+- /opt/homebrew/lib
+
+extra-include-dirs:
+- /opt/homebrew/include/
+- /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/include/ffi/
 
 # Override default flag values for local packages and extra-deps
 # flags: {}
--- a/concordium-consensus/stack.yaml
+++ b/concordium-consensus/stack.yaml
@@ -47,11 +47,6 @@ extra-deps:
 extra-lib-dirs:
 - ../concordium-base/lib
 - ./smart-contracts/lib
-- /opt/homebrew/lib
-
-extra-include-dirs:
-- /opt/homebrew/include/
-- /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/include/ffi/
 
 # Override default flag values for local packages and extra-deps
 # flags: {}
--- a/concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/ABBA.hs
+++ b/concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/ABBA.hs
@@ -66,7 +66,7 @@ selectFromSeq' g s =
     (Seq.index s n, Seq.deleteAt n s, g')
 
 atParty :: Party -> Traversal' (Vec.Vector a) a
-atParty = ix . fromIntegral
+atParty p = ix $ fromIntegral p
 
 runABBATestRG :: RandomGen g => g -> BS.ByteString -> Int -> Int -> Vec.Vector VRF.KeyPair -> Seq.Seq (Party, ABBAInput) -> IO Property
 runABBATestRG g0 baid nparties allparties vrfkeys = go g0 iStates iResults
--- a/concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/WMVBA.hs
+++ b/concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/WMVBA.hs
@@ -47,8 +47,8 @@ genBlsKeys n = do
   vectorOf n genBlsKeyPair
 
 -- TODO: generate random string
-genByteString' :: Gen BS.ByteString
-genByteString' = BS.pack <$> vector 8
+genRandomByteString :: Gen BS.ByteString
+genRandomByteString = BS.pack <$> vector 8
 
 -- Used to build streams of test data
 data WMVBAInput
@@ -137,7 +137,7 @@ testData2 baid v keys corrupted wrongkey = inputs
 runNonCorruptedTest :: Property
 runNonCorruptedTest = monadicIO $ do
   keys <- pick $ genKeys nparties
-  baid <- pick $ genByteString'
+  baid <- pick $ genRandomByteString
   stream <- pick $ shuffle $ testData1 baid blockA keys
   liftIO $ runWMVBATest me baid nparties keys (StartWMVBA blockA : stream)
     where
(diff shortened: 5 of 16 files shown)


concordium-consensus/stack.yaml
```diff
@@ -47,6 +47,11 @@ extra-deps:
 extra-lib-dirs:
 - ../concordium-base/lib
 - ./smart-contracts/lib
+- /opt/homebrew/lib
```
abizjak, line 50: This is not part of the repository, so cannot  remain. It is also mac specific.

concordium-consensus/stack.yaml
```diff
@@ -47,6 +47,11 @@ extra-deps:
...
+
+extra-include-dirs:
+- /opt/homebrew/include/
+- /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/include/ffi/
```
abizjak, line 54: Same here.

concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/ABBA.hs
```diff
@@ -66,7 +66,7 @@ selectFromSeq' g s =
     (Seq.index s n, Seq.deleteAt n s, g')
 
 atParty :: Party -> Traversal' (Vec.Vector a) a
-atParty = ix . fromIntegral
+atParty x = ix $ fromIntegral x
```
abizjak, line 69: I don't know why this is needed. In general these kind of changes may affect resource usage of the function since they change the inlining opportunities. This is only in tests, but still we should refrain from making these kinds of changes. They can cause unexpected problems.

concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/WMVBA.hs
```diff
@@ -47,8 +47,8 @@ genBlsKeys n = do
   vectorOf n genBlsKeyPair
 
 -- TODO: generate random string
-genByteString :: Gen BS.ByteString
-genByteString = BS.pack <$> vector 8
+genByteString' :: Gen BS.ByteString
+genByteString' = BS.pack <$> vector 8
```
abizjak, line 51: Why? 

This is not a good name. What does the `'` signify?

concordium-node/build.rs
```diff
@@ -3,9 +3,9 @@ use std::{env, path::Path};
 use std::{process::Command, str};
 
 #[cfg(all(not(feature = "static"), target_os = "linux"))]
-const GHC_VARIANT: &str = "x86_64-linux-ghc-8.10.4";
+const GHC_VARIANT: &str = "x86_64-linux-ghc-9.0.2";
 #[cfg(all(not(feature = "static"), target_os = "macos"))]
-const GHC_VARIANT: &str = "x86_64-osx-ghc-8.10.4";
+const GHC_VARIANT: &str = "aarch64-osx-ghc-9.0.2";
```
abizjak, line 8: This change means that the node will not properly build on an Intel mac.

We need another solution with more targeted platforms.

concordium-node/build.rs
```diff
@@ -158,6 +158,11 @@ fn link_ghc_libs() -> std::io::Result<std::path::PathBuf> {
         ]))
     });
     let rts_dir = Path::new(&ghc_lib_dir).join("rts");
+    let rts_dir = if rts_dir.exists() {
+        rts_dir
+    } else {
+        Path::new(&ghc_lib_dir).join(GHC_VARIANT)
+    };
```
abizjak, line 165: Even if this works on all platforms it needs documentation as to why we are doing this.

I am also not a fan of this solution. It works by accident. It would be better to be explicit
- if linux do the first thing
- if intel mac do ...
- if M1 mac do the third thing

The way it is now is very indirect so hard to understand.

stack.yaml
```diff
@@ -13,3 +13,8 @@ extra-deps:
 extra-lib-dirs:
 - ./concordium-base/lib
 - ./concordium-consensus/smart-contracts/lib
+- /opt/homebrew/lib
+
+extra-include-dirs:
+- /opt/homebrew/include/
+- /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/include/ffi/
```
abizjak, line 20: Same comment as above.

concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/ABBA.hs
rimbi: Frankly, I did not spend time investigating the cause of the issue but the change was required due to a compilation error. I have seen similar problems in 1 or 2 other places in other repos too.

concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/WMVBA.hs
rimbi: Again a compiler error due to name resolution. I put the `'` to resolve the issue. Any better idea?

concordium-node/build.rs
rimbi: Good catch! 👍

concordium-consensus/tests/consensus/ConcordiumTests/Afgjort/ABBA.hs
td202: This is due to the simplified subsumption introduced in GHC 9. https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0287-simplify-subsumption.rst#costs-and-drawbacks
This does mean that eta-expansion is required in various places.

rimbi: Thanks 👍 
Good to know, bad to not understand :)

CHANGELOG.md
```diff
@@ -60,6 +61,7 @@
   - increased smart contract module size limit, 512kB
   - a number of cryptographic primitives
 - Node can now be stopped during out of band catchup by using signals, SIGINT and SIGTERM.
+- Upgrade resolver to version `lts-19.6` and ghc to version `9.0.2`.
```
abizjak, line 64: The changelog is not for these kinds of changes. This is not visible to any users.

concordium-consensus/stack.static.yaml
```diff
@@ -12,4 +12,4 @@ extra-lib-dirs:
 - ./smart-contracts/lib
 
 ghc-options:
-    "$everything": -fPIC
+    "$everything": -fPIC -fsimpl-tick-factor=10000
```
abizjak, line 15: Can you please document why this latter option is necessary. And link to the GHC issue https://gitlab.haskell.org/ghc/ghc/-/issues/14637#note_413425 for future maintainers to be able to clean up.

concordium-node/README.md
```diff
@@ -1,26 +1,30 @@
...
-* collector - enables the build of the node-collector and backend
-* dedup_benchmarks - enable support in the benchmarks for deduplication queues
+
+- instrumentation - switches the default internal counter implementation out with prometheus
+- instrumentation - enables stats data exporting to [prometheus](https://crates.io/crates/prometheus)
+- network_dump - makes the network dumping capabilites available.
+- static - build against static haskell libraries (Linux only)
+- profiling - build against haskell libraries with profiling support enabled (Linux only)
+- collector - enables the build of the node-collector and backend
+- dedup_benchmarks - enable support in the benchmarks for deduplication queues
```
abizjak, line 27: I do not see what is changed here. Please refrain from these kinds of changes. I don't see what is improved here at all.

concordium-node/README.md
```diff
@@ -92,38 +97,71 @@ Environment variables only apply to the default build. This links with shared Ha
...
-* [protoc](https://github.com/protocolbuffers/protobuf/releases) >= 3.7.1
-* PostGreSQL and LMDB should be installed under `stack`'s `msys2` installation, which can be done with the following commands:
-```
+- Haskell [stack](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
+- [Rust](https://www.rust-lang.org/tools/install)
+  - For building the node, the toolchain `1.53.0-x86_64-pc-windows-gnu` is required, which can be installed with the command: `rustup toolchain install 1.53.0-x86_64-pc-windows-gnu`.
+  - For building the node runner service (optional), the toolchain `1.53.0-x86_64-pc-windows-msvc`  is required, which can be installed with the command: `rustup toolchain install 1.53.0-x86_64-pc-windows-msvc`.
+- [flatc](https://github.com/google/flatbuffers/releases/tag/v2.0.0) 2.0.0 (should be in the path)
+- [protoc](https://github.com/protocolbuffers/protobuf/releases) >= 3.7.1
+- PostGreSQL and LMDB should be installed under `stack`'s `msys2` installation, which can be done with the following commands:
```
abizjak, line 157: Also here, I do not see what is changed or improved. Changing `*` to `-` does not seem useful.

concordium-node/README.md
```diff
@@ -92,38 +97,71 @@ Environment variables only apply to the default build. This links with shared Ha
...
+
+```bash
 stack exec -- pacman -Syuq --noconfirm
 stack exec -- pacman -Syq mingw-w64-x86_64-postgresql mingw-w64-x86_64-lmdb --noconfirm
 ```
-* If building the installer, the [Wix Toolset](https://wixtoolset.org/releases/) is required, and should be in the path.
+
+- If building the installer, the [Wix Toolset](https://wixtoolset.org/releases/) is required, and should be in the path.
```
abizjak, line 164: Also here,.

concordium-node/README.md
```diff
@@ -134,19 +172,22 @@ However, running the node requires the consensus DLL, which is compiled to `..\c
 If you wish to run the node directly, you should copy `HSdll.dll` to the location of `concordium-node.exe`.
 The node also depends on a number of DLLs that are part of `stack`'s `msys2` installation.
 To run the node with these DLLs in the path, you can use `stack exec`, as in:
-```
+
+```bash
```
abizjak, line 176: It would be good to be consistent if adding these. Either use `bash` or `shell` or `console` consistently.

concordium-node/README.md
```diff
@@ -134,19 +172,22 @@ However, running the node requires the consensus DLL, which is compiled to `..\c
...
 stack exec -- concordium-node.exe --help
 ```
 
 ### Build issue: long paths
+
 If the root directory of the repository is too long, then the build may fail with `No such file or directory` errors, such as:
-```
+
+```console
```
abizjak, line 184: See above.

concordium-node/build.rs
```diff
@@ -157,7 +159,10 @@ fn link_ghc_libs() -> std::io::Result<std::path::PathBuf> {
             "--print-libdir",
         ]))
     });
+    #[cfg(all(not(feature = "static"), target_os = "linux"))]
     let rts_dir = Path::new(&ghc_lib_dir).join("rts");
+    #[cfg(all(not(feature = "static"), target_os = "macos"))]
+    let rts_dir = Path::new(&ghc_lib_dir).join(GHC_VARIANT);
```
abizjak, line 165: Please document your findings here so that future maintainers can act.

That is, document that for the M1 GHC distribution the shared libraries for the RTS are in the GHC_VARIANT directory, not where they are for x86 distributions.

abizjak (approved):

concordium-node/README.md
rimbi: It's a markdown linter fix, the symbol for the list items should be. consistent in a file, either. `-` or `*`

concordium-node/README.md
rimbi: It's a markdown linter fix, the symbol for the list items should be. consistent in a file, either. `-` or `*`

concordium-node/README.md
rimbi: It's a markdown linter fix, the symbol for the list items should be. consistent in a file, either. `-` or `*`

concordium-node/README.md
rimbi: agree, do you have any preference?

concordium-node/build.rs
rimbi: isn't the code explaining what you are telling above?

CHANGELOG.md
rimbi: done

concordium-consensus/stack.static.yaml
rimbi: done

concordium-node/build.rs
rimbi: Also, this difference is valid for Intel macs too.

concordium-node/README.md
rimbi: Chosen `console`

PR #350, opened 2022-05-21
Avoid loading genesis

### LINKED ISSUE #340: Node startup time should not increase "linearly" with the number of protocol updates
Linearly here is a rough guideline, but essentially during startup all the states for all existing protocol versions (since the genesis version) are loaded and all the protocol updates are processed. So with each new protocol update the startup time is increased. The more accounts and contracts there are at that point the more it is increased. Almost all of the state for the old protocol versions is dropped. Only the current one is maintained.

This delay can be seen by observing "starting chain with genesis" in the node logs.

Since the node's database is trusted there is no need to process all old protocol updates. The node does need to support queries for old blocks, transactions and state, but this just means loading a little bit of metadata and correct database handles.

Improvement here will mean

 -  avoid loading block state for all but the last protocol version
 -  avoid processing protocol updates again. This will significantly decrease memory use during startup as well.

## Acceptance criteria

The number of protocol updates doesn't change the startup time.

## References

- The function where the essence of loading the state is located 
https://github.com/Concordium/concordium-node/blob/main/concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs#L271

- The actual loading of the state happens in `makeBlockPointerCached`. 

Related to epic #289.

abizjak:
## Purpose

Improve node startup time by avoiding to load data that is not needed.

Closes #340

## Changes

- [x] Avoid processing protocol updates that have already been processed by the node and whose results are already stored in the state.
- [x] To achieve the above it was necessary to avoid storing genesis data as well. Genesis data was stored for each genesis index in two places, in the genesisData field of SkovPersistentData, as well as behind the genesisBlockPointer field. Both of those now store `GenesisConfiguration` only which has a small, fixed size. The effect of this should be considerably reduced memory use in case of multiple protocol updates with large number of accounts or contracts.
- [x] In order to avoid loading genesis some changes in LMDB were needed. The new node will no longer store the full genesis block in the tree state. Instead only genesis configuration is stored. The new node is compatible with existing state. It will manage to load GenesisConfiguration from the stored genesis block. However the opposite is not true. The state produced by node version 4.1 cannot be used by node versions < 4.1.
- [x] The effect of the changes is that the node will no longer check whether the genesis data it was given is compatible with the state that it has.

## Depends on 
- [x] https://github.com/Concordium/concordium-base/pull/152

### Startup time of mainnet node
- After the changes around 1min 50s
```
2022-05-23T06:55:17.228017214Z: INFO: Starting up concordium_node version 4.1.0!
...
2022-05-23T06:57:04.318102248Z: INFO: Starting RPC server
```

Memory use around 2.5GB after startup.

- Before the changes around 3min 40s
```
May 23 09:03:10 capybara concordium-mainnet-node-4.0.11[2258237]: [INFO  concordium_node::utils] Starting up concordium_node version 4.0.11!
May 23 09:06:51 capybara concordium-mainnet-node-4.0.11[2258237]: [INFO  concordium_node] Starting RPC server
```

Memory use around 4GB (unshared) after startup.

## Checklist

      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unrelease changes
 
+- Improve startup time by avoiding to process already processed protocol
+  updates.
+- Decrease memory usage by not storing genesis blocks. This has the effect that
+  the database of node versions >= 4.1.* are not compatible with those of < 4.1.
+
+## 4.0.11
 - The `SendTransaction` function exposed via the gRPC interface now provides the caller with detailed error messages if the 
   transaction was rejected instead of just `False`. The function still returns `True` if 
   the transaction was accepted.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 ## Unrelease changes
 
-- Improve startup time by avoiding to process already processed protocol
+- Improve startup time by avoiding processing already processed protocol
   updates.
 - Decrease memory usage by not storing genesis blocks. This has the effect that
-  the database of node versions >= 4.1.* are not compatible with those of < 4.1.
+  the database produced by node versions >= 4.2.* cannot be used by node
+  versions <= 4.1. The other direction works.
 
 ## 4.0.11
 - The `SendTransaction` function exposed via the gRPC interface now provides the caller with detailed error messages if the 
--- a/concordium-consensus/src/Concordium/GlobalState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState.hs
@@ -345,20 +345,20 @@ deriving via TreeStateBlockStateM pv g c r s m
 -----------------------------------------------------------------------------
 
 -- |Configuration that uses in-memory, Haskell implementations for both tree state and block state.
-newtype MemoryTreeMemoryBlockConfig (pv :: ProtocolVersion) = MTMBConfig {
+newtype MemoryTreeMemoryBlockConfig = MTMBConfig {
     mtmbRuntimeParameters :: RuntimeParameters
     }
 
 -- |Configuration that uses the in-memory, Haskell implementation of tree state and the
 -- persistent Haskell implementation of block state.
-data MemoryTreeDiskBlockConfig (pv :: ProtocolVersion) = MTDBConfig {
+data MemoryTreeDiskBlockConfig = MTDBConfig {
     mtdbRuntimeParameters :: !RuntimeParameters,
     mtdbBlockStateFile :: !FilePath
     }
 
 -- |Configuration that uses the disk implementation for both the tree state
 -- and the block state
-data DiskTreeDiskBlockConfig (pv :: ProtocolVersion) = DTDBConfig {
+data DiskTreeDiskBlockConfig = DTDBConfig {
     dtdbRuntimeParameters :: !RuntimeParameters,
     dtdbTreeStateDirectory :: !FilePath,
     dtdbBlockStateFile :: !FilePath
@@ -367,7 +367,7 @@ data DiskTreeDiskBlockConfig (pv :: ProtocolVersion) = DTDBConfig {
 -- |Configuration that uses the disk implementation for both the tree state
 -- and the block state, as well as an external database for producing
 -- an index of transactions affecting a given account.
-data DiskTreeDiskBlockWithLogConfig (pv :: ProtocolVersion) = DTDBWLConfig {
+data DiskTreeDiskBlockWithLogConfig = DTDBWLConfig {
     dtdbwlRuntimeParameters :: !RuntimeParameters,
     dtdbwlTreeStateDirectory :: !FilePath,
     dtdbwlBlockStateFile :: !FilePath,
@@ -387,7 +387,7 @@ instance Show GlobalStateInitException where
 instance Exception GlobalStateInitException
 
 -- |This class is implemented by types that determine configurations for the global state.
-class GlobalStateConfig (c :: ProtocolVersion -> Type) where
+class GlobalStateConfig (c :: Type) where
     -- |The read-only context type associated with a global state configuration.
     type GSContext c (pv :: ProtocolVersion)
     -- |The (mutable) state type associated with a global state configuration.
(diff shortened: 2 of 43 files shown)


concordium-consensus/src/Concordium/GlobalState/Block.hs
```diff
@@ -330,15 +330,18 @@ instance BlockData (Block pv) where
     {-# INLINE blockTransactions #-}
 
 instance (IsProtocolVersion pv) => EncodeBlock pv (Block pv) where
-    putBlock _ (GenesisBlock gd) = put genesisSlot >> put gd
+    putBlock _ (GenesisBlock gd) = put genesisSlot >> (putGenesisConfiguration gd)
```
td202, line 333: It is not clear to me that `getGenesisConfiguration` is and always will be the inverse of `putGenesisConfiguration`. This seems precarious and at the very least requires documentation.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -365,10 +338,39 @@ loadSkovPersistentData rp _treeStateDirectory _genesisData pbsc atiContext = do
...
-      bstate <- runReaderT (PBS.runPersistentBlockStateMonad (cacheBlockState =<< loadBlockState (blockStateHash sbBlock) sbState)) pbsc
-      makeBlockPointerFromPersistentBlock sbBlock bstate defaultValue sbInfo
+
+activateSkovPersistentData :: forall ati pv. (IsProtocolVersion pv, CanExtend (ATIValues ati))
+                           => PBS.PersistentBlockStateContext
+                           -> SkovPersistentData pv ati (PBS.HashedPersistentBlockState pv)
+                           -> LogIO (SkovPersistentData pv ati (PBS.HashedPersistentBlockState pv))
+activateSkovPersistentData pbsc uninitState = do
+  cachedLastFinalized <- liftIO (makeBlockPointerCached (_lastFinalized uninitState))
+  -- We need to establish is the transaction table invariants.
```
mkmks, line 348: I see that this comment came unchanged from the body of `loadSkovPersistentData` after `getTransactionTable` was moved. It'd be better to list the invariants that `activateSkovPersistentData` checks upfront, like it's done in the comment documenting `loadSkovPersistentData`.

There's, perhaps, only one invariant that's checked here ("every account in the account lists is stored in the database"), but still the function description will be easier to read if it says explicitly what needs to be true to run this function.

There's also no need to say anything about next available nonces when talking about this invariant, I believe — that's not what's being checked.

concordium-consensus/src/Concordium/GlobalState.hs
```diff
@@ -401,7 +397,22 @@ class GlobalStateConfig (c :: ProtocolVersion -> Type) where
...
+    -- expensive and is not needed if the state already exists.
+    initialiseGlobalState :: IsProtocolVersion pv => c pv -> LogIO (Either (GSContext c pv, GSState c pv, GSLogContext c pv) (GenesisData pv -> LogIO (GSContext c pv, GSState c pv, GSLogContext c pv)))
+
+    -- |A wrapper around 'initialiseGlobalState' in case genesis data is already available.
+    initialiseGlobalStateWithGenesis :: IsProtocolVersion pv => GenesisData pv -> c pv -> LogIO (GSContext c pv, GSState c pv, GSLogContext c pv)
+    initialiseGlobalStateWithGenesis gd cfg = initialiseGlobalState cfg >>= \case
+        Left x -> return x
+        Right k -> k gd
+
+    activateGlobalState :: IsProtocolVersion pv => Proxy (c pv) -> GSContext c pv -> GSState c pv -> LogIO (GSState c pv)
```
mkmks, line 415: Please explain how this function is different from `initialiseGlobalState`. These two names appear very close semantically. When do I use one and when do I use another?

I see that there's a comment like the one I'm asking about here in `SkovConfiguration`, which is describing the use of `activateSkovState`.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -559,11 +570,129 @@ makeMultiVersionRunner
...
+             -> Maybe (EVersionedConfiguration gsconf finconf)
+             -- ^If this is the first iteration of the loop then this will be 'Nothing'. Otherwise it is the
+             -- versioned configuration produced in the previous iteration of the loop.
+             -> GenesisIndex
+             -- ^Genesis index at which to attempt to load the state.
+             -> AbsoluteBlockHeight
+             -- ^Absolute block height of the genesis block of the new chain.
+             -> MVR gsconf finconf ()
+        loop (SomeProtocolVersion (_ :: SProtocolVersion pv)) first vcIndex vcGenesisHeight = do
+            let comp = MVR $
```
mkmks, line 630: This function builds an `Either` value with quite a few `Maybe`s nested inside. Perhaps, using a custom type (with 5 constructors, if I counted right) could help to improve the readability of the pattern matching tree that consumes this value.

mkmks (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 
-## Unrelease changes
+## Unreleased changes
+- Improve startup time by avoiding to process already processed protocol
+  updates.
+- Decrease memory usage by not storing genesis blocks. This has the effect that
+  the database of node versions >= 4.1.* are not compatible with those of < 4.1.
```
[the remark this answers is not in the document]
abizjak: Rephrased.

concordium-consensus/src/Concordium/GlobalState/Block.hs
abizjak: It is documented as part of `getGenesisConfiguration`.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -80,30 +83,56 @@ data StoredBlock pv st = StoredBlock {
...
+instance (FixedSizeSerialization a, FixedSizeSerialization b) => FixedSizeSerialization (a, b) where
+  serializedSize _ = serializedSize (Proxy @a) + serializedSize (Proxy @b)
+
+instance (IsProtocolVersion pv, FixedSizeSerialization st) => MDBDatabase (BlockStore pv st) where
   type DBKey (BlockStore pv st) = BlockHash
   type DBValue (BlockStore pv st) = StoredBlock pv st
   encodeKey _ = hashToByteString . blockHash
+  encodeValue _ sb = S.runPutLazy (putStoredBlock sb)
+  decodeValue _ blockHash mdbVal = do
+    bs <- unsafeByteStringFromMDB_val mdbVal
```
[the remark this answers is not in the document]
abizjak: Even if the resulting type is harboring bytestrings it is generally not an issue since `getByteString` creates a copy. We would have to be using `getBytes` directly and retain the resulting bytestring to retain pointers into the source, but I don't see us doing that.

While I see that it is a bit risky, I think the particular use that we have is OK and avoids copying potentially a large amount of data for genesis blocks.

I have provided more documentation.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -570,7 +599,7 @@ writeBlock block = resizeOnFull blockSize
         storeReplaceRecord txn (dbh ^. finalizedByHeightStore) (_bpHeight b) (_bpHash b)
         storeReplaceRecord txn (dbh ^. blockStore) (_bpHash b) block
   where
-    blockSize = 2*digestSize + fromIntegral (LBS.length (S.encodeLazy block))
+    blockSize = 1000 + fromIntegral (_bpTransactionsSize (sbInfo block))
```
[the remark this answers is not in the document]
abizjak: It's a leftover from some intermediate version. Removed.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
abizjak: I don't know exactly what you are referring to. This function is establishing the invariants. It is not checking them.

And the state invariant it is establishing is that the pending table must record the correct "next nonce" for each account.

concordium-consensus/src/Concordium/MultiVersion.hs
abizjak: I am in favor of this.

I have documented the branches quite explicitly which in my opinion documents the results of this function quite well.

If I added a new type then those comments would just be next to the type. Seeing that this is just an inline loop I do not see the advantage.

PR #351, opened 2022-05-22
Info log when blocks are received.

EmilLa1:
## Purpose

Info log whenever blocks are received in contrast to today when it is only logged when running with debug log level. 

Today we already info log when blocks 'arrive' into the state. 
Info logging the receive time as well could potentially be useful when inspecting logs and understanding the network. 

Info logging the block received event will not pollute the log and as such it would be ok to do. 

## Changes

Changed log level from debug to info for when blocks are received. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Skov/Update.hs
+++ b/concordium-consensus/src/Concordium/Skov/Update.hs
@@ -488,7 +488,7 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
                 -- Check that the claimed key matches the signature/blockhash
                 checkClaimedSignature pb $ do
                 -- The block is new, so we have some work to do.
-                logEvent Skov LLDebug $ "Received block " ++ show pb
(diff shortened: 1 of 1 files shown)


abizjak (commented):
I'm not opposed to the change as such, but this seems like it's a workaround for what you actually want, seeing how much time it took to process the block.

We log how much time it took to construct a block, so it would make sense to also log how much time it took to process it explicitly. The former is already at info level.

EmilLa1:
It was not per se related to measuring block execution but more for a providing better information at the default info log level. As the 'arrive' time is already there I think the 'receive' time would be beneficial to include at info level as well. 

The receive time can also be used for analysing block propagation times wrt. the slot time etc.

abizjak (approved):
I don't want this to evolve into bikeshedding.

But in general I would probably prefer we reduce the info level logging rather than add to it. There is already too much noise there, and not a whole lot of useful information.

In any case, let's merge this since it does not matter.

Issue #363, opened 2022-05-25
Incorrect event logged for V1 contract transfer


abizjak:
**Bug Description**

An incorrect event is logged when a contract transfers an amount. The amount value is incorrect.

Concretely in the scheduler
```
                -- Add the transfer to the current changeset and return the corresponding event.
                lift (withContractToAccountAmountV1 (instanceAddress senderInstance) targetAccount tAmount $
                      return [Transferred addr tAmount (AddressAccount accAddr)])
```
is what we should have and
```
                -- Add the transfer to the current changeset and return the corresponding event.
                lift (withContractToAccountAmountV1 (instanceAddress senderInstance) targetAccount tAmount $
                      return [Transferred addr transferAmount (AddressAccount accAddr)])
```
is what we have.

This affects node version 4.

Note that the bug is in the transfer amounts, the total balances of contracts and accounts are correct.

The events are hashed and become part of a block hash, so the fix will need some thinking.

abizjak:
This was noticed on testnet by Martin from fintech builders.

At least the block `39c9a5135f04215f60397cd551c5cffc7bcbfd9e31fcd4d04503b679c10c7b24` is affected.

td202:
The most obvious options are:

1. Make the simple fix. This is incompatible with testnet, and would effectively require a (not-insignificant) rollback. This would be quite bad, and we may have difficultly getting all nodes on testnet to rollback.
2. Make the fix in a new protocol version (P5) that is otherwise identical to P4. We would update testnet to P5, and update mainnet directly from P3 to P5, skipping the bugged P4. 
3. Live with the bug until a future protocol update, which will include the fix.

abizjak:
1. I don't think this is realistically acceptable. But we can discuss it next week.
2. That is in some sense cleanest of the options you listed, but it is very heavy and will complicate our lives in the future.
3. I don't think this is acceptable either. In particular wallet-proxy will have the wrong view of the world.

I think pragmatically the following option could work. We fix the bug in P4 so that the correct amount is logged in the event.
We then "fix" the hashing by substituting the transferAmount into the Transferred event in `getTransactionOutcomesHash` only.
The transferAmount appears inside the Updated event already, so it should be straightforward to do this patching.

We can then fix this hashing in a followup protocol version (where we would ideally also fix some other hashing, such as mapping all "reject reasons" to a single value, and hashing transactions in a tree structure instead of a flat list).

abizjak:
Closed by #387

PR #365, opened 2022-05-28
Fix the event that is generated when a transfer is made from a V1 con…

abizjak:
## Purpose

An alternative way of fixing https://github.com/Concordium/concordium-node/issues/363

Depends on 
- [ ] https://github.com/Concordium/concordium-base/pull/158

## Changes

Correct the amount in the transfer event.

## Checklist

      hard-to-understand areas.


abizjak:
Due to a severe issue with delegation we will not be doing this directly.

PR #370, opened 2022-05-31
Node caching: Implement LRU cache

rimbi:
## Purpose

Adds LRU Cache as a caching strategy alternative.

## Changes

Added LRU Cache by making use of `lrucache` package.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/haskell-bench/LRUCache.hs
+++ b/concordium-consensus/haskell-bench/LRUCache.hs
@@ -23,10 +23,10 @@ data CacheOp k v
   = PutItem k v
   | GetItem k
 
-testCache :: forall v. (Cache v, HasCache v v) => v -> [CacheKey v] -> CacheValue v -> Int -> IO ()
+testCache :: forall v. (Cache v) => v -> [CacheKey v] -> CacheValue v -> Int -> IO ()
 testCache cache keys value txCount = do
   let cacheProxy = Proxy :: Proxy v
-  flip runReaderT cache $ do
+  flip runReaderT (CacheContext cache) $ do
     res <-
       forM
         (take txCount (reverse keys))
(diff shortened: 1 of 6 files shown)


td202 (approved):

concordium-consensus/package.yaml
```diff
@@ -59,6 +59,7 @@ dependencies:
 - recursion-schemes >= 5.1
 - file-embed >= 0.0.11
 - lmdb
+- lrucache
```
abizjak, line 62: What evaluation was done of this dependency? Looking at it on hackage, it is not very widely used.

Looking at the code a bit, it does not seem to be written with high performance as a major goal. For example it uses Data.Map which is for most purposes quite a bit slower than HashMap or IntMap (if this applies). It partially also uses the strict map, but in many places it also uses `Map.insert` for a lazy map.

These are not problems on their own, but in order to understand the benefits we need to understand the performance of this implementation. Including the potential for space leaks due to laziness.

rimbi: `LRUCache` is just one of the alternatives and the main reason to add it was the presence of the `lrucache` package (thus a minimal effort). We have not evaluated the performances yet. If its performance is inadequate, we could switch to another cache implementation or perhaps delete it entirely.

abizjak: I don't understand this reasoning. What are you evaluating exactly then?

The main thing we should be considering is the performance of our node and what would make most sense. If we use a library for caching that has poor performance for the specific algorithm then we are not evaluating whether the specific caching strategy would make sense. And that is what we should be doing.

rimbi: Nope, that's not quite right. The current LRU solution might still outperform the others. Let's start simple by using the 3rd party solution and maybe it will be sufficient. if not, we can develop our own solution too. Let's wait and see how the performances turn out. Then we may go into more detail.

EmilLa1: I would be in favor of not using this dependency because of the reasons already stated above.

We supposedly can do much better using better data structures, `IntMap` instead of the generic `Data.Map` in particular. - We need a cache with performance of very high priority - and I think using this for benchmarks would potentially rise as many questions as it will answers.

I think we should try implement our own LRU with performance in mind.

EmilLa1 (commented):
I would be in favor of not using this dependency because of the reasons already stated above.

We supposedly can do much better using better data structures, IntMap instead of the generic Data.Map in particular. - We need a cache with performance of very high priority - and I think using this for benchmarks would potentially rise as many questions as it will answers.

I think we should try implement our own LRU with performance in mind.

EDIT:
But if you want this in now for testing purposes I am also fine with that - I'm just unsure about the real benefit we will gain.

concordium-consensus/src/Concordium/GlobalState/Persistent/Cache.hs
```diff
@@ -23,15 +26,17 @@ class Cache c where
 
     putCachedValue :: (MonadCache r c m) => Proxy c -> CacheKey c -> CacheValue c -> m (CacheValue c)
     lookupCachedValue :: (MonadCache r c m) => Proxy c -> CacheKey c -> m (Maybe (CacheValue c))
+    cacheSize :: (MonadCache r c m) => Proxy c -> m Int
```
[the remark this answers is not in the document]
rimbi: done

concordium-consensus/haskell-bench/LRUCache.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Main where
+
+import Concordium.GlobalState.Persistent.BlobStore
+import Concordium.GlobalState.Persistent.Cache
+-- import Data.Random
```
EmilLa1, line 8: Remove.

concordium-consensus/src/Concordium/GlobalState/Persistent/Cache.hs
```diff
@@ -41,6 +48,12 @@ class Cache c where
 
     putCachedValue :: (MonadCache c m) => Proxy c -> CacheKey c -> CacheValue c -> m (CacheValue c)
     lookupCachedValue :: (MonadCache c m) => Proxy c -> CacheKey c -> m (Maybe (CacheValue c))
+    getCacheSize :: (MonadCache c m) => Proxy c -> m Int
+
+newtype CacheContext c = CacheContext { theCacheContext :: c }
```
EmilLa1, line 53: At some point we need to start documenting our abstractions - but I suppose this can be postponed until the integration task has been completed.

concordium-consensus/haskell-bench/LRUCache.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Main where
+
```
EmilLa1, line 5: Can you perhaps put some module level documentation on what this benchmark does and how. 
(You already said it a few times now, but it will serve us nicely in the future)

concordium-consensus/src/Concordium/GlobalState/Persistent/Cache.hs
rimbi: done

concordium-consensus/haskell-bench/LRUCache.hs
rimbi: done

concordium-consensus/haskell-bench/LRUCache.hs
rimbi: done

concordium-consensus/package.yaml
rimbi: The benchmars have shown that the `lrucache` does not perform very well when compared to our own `FIFOCache` implementation. We'll implement replace it with our own LRUCache implementation.

PR #389, opened 2022-06-11
Loosen the sandboxing and allow writable and executable regions.

### LINKED ISSUE #388: Static node build fails to run on linux.
**Bug Description**

The static build of the node for ubuntu for main does not work.

```
[INFO  concordium_node::consensus_ffi::consensus] Starting up the consensus layer
concordium-mainnet-node-4.2.0: internal error: mmapForLinkerMarkExecutable: mprotect: Operation not permitted
    (GHC version 9.0.2 for x86_64_unknown_linux)
    Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
 Main process exited, code=killed, status=6/ABRT
````

**Steps to Reproduce**

Build a statically linked node from main (e.g., `fbc7f5d`). 

**Expected Result**

The node can start.

**Actual Result**

The node does not start, and the error above is reported.

abizjak:
## Purpose

Fixes #388 

It is unclear to me why this is not a problem when compiling with 8.10.4, but it is a problem when compiling with GHC 9.0.2, but something about dynamic code generation must have changed in between the two versions.

## Changes

Remove the `MemoryDenyWriteExecute` restriction in the debian package.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/distribution/ubuntu-packages/template/debian/concordium-node.service
+++ b/scripts/distribution/ubuntu-packages/template/debian/concordium-node.service
@@ -26,7 +26,10 @@ ProtectKernelTunables=yes
 CapabilityBoundingSet=
 LockPersonality=yes
 RestrictRealtime=yes
-MemoryDenyWriteExecute=yes
+# The current node version generates some code at runtime to bridge the FFI
+# between Haskell and rust related to smart contracts. Hence we have to allow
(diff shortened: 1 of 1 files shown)


EmilLa1 (approved):
lgtm

Issue #385, opened 2022-06-09
High resource use of `GetConsensusInfo`


abizjak:
**Bug Description**

There is unexpected and rather bizarre behaviour of GetConsensusInfo. Querying GetConsensusInfo twice a second uses a large amount of CPU. How much depends on the runtime parameters used to run the node.

If `-N4` then it seems to use around 120% **even though queries take at most 4ms**, and there are two of them per second.
If `-N4` and `-qn1` then CPU use is around 50%, **but queries take 80ms or so**.
If `-N1` then queries take 30ms, sometimes more, and CPU usage is around 30%.

**Steps to Reproduce**

I have tested this on a network with protocol version 4 (where genesis started at P3) and around 80000 accounts.

I do not know how consensusInfo used to behave, but it just seems unnacceptable that a query that takes no time suddenly means that a node is using 100% all the time.

abizjak:
Probably related to #201

abizjak:
GetConsensusInfo is largely a red herring. The issue is caused by the idle garbage collector http://downloads.haskell.org/~ghc/latest/docs/html/users_guide/runtime_control.html#rts-flag--I%20%E2%9F%A8seconds%E2%9F%A9 running a major GC ever 0.3s, which, with 80000 accounts, is quite slow and resource intensive.

It is quite interesting that making queries 0.25s apart makes both the node CPU use low, as well as the queries fast, and making them 0.35s apart makes them **slower** and node CPU usage much higher.

Running with `-I0` seems to reduce CPU use drastically, although I do not know what the impact on memory use is. That needs more long-term testing.

abizjak:
The outcome of this is that there are no changes in the node needed apart from those that are already included on main (reduction of genesis block hashing computation).

We will experiment with different GC flags, but running over the last day on stagenet seems to indicate that `-I0` is an effective solution with no downsides.

PR #391, opened 2022-06-13
Update installer versions, and use -I0 in mac and ubuntu distributions.

abizjak:
## Purpose

Bump installer versions.

Address #385 

## Changes

- [x] Change version and GUID of the windows installer.
- [x] Use `-I0` in ubuntu and mac distributions. This disables the idle garbage collector and only affects non-baking nodes (at least with slot duration set for testnet/mainnet).

## Checklist

      hard-to-understand areas.


abizjak:
> What about windows?

I added it to Windows as well. Originally I did not since no runtime flags are currently set on Windows and the behaviour might well be different.

But as long as we test thoroughly that this does not cause a regression we should be OK.

EmilLa1 (approved):
lgtm

Bargsteen (approved):
Mac parts look good

PR #392, opened 2022-06-15
Added the ability to download the catch-up file from configured URL

### LINKED ISSUE #381: Allow the node to fetch the catchup file
Currently, out-of-band catchup requires the node operator to download the catchup file and to specify its location using the `--import-blocks-from` flag.

We'd like to add a node configuration option (say, `CONCORDIUM_NODE_OUT_OF_BAND_CATCH_UP_URL`) to specify the catchup file URL and enable the node to download the catchup file without operator intervention if the option is set.

# Acceptance criteria

A node started on the mainnet with an empty database and `CONCORDIUM_NODE_OUT_OF_BAND_CATCH_UP_URL` set to     https://catchup.mainnet.concordium.software/blocks_to_import.mdb should complete the catch up without `--import-blocks-from` and `--bootstrap-node` flags specified.

mkmks:
## Purpose

Adds the ability to download catch-up files during node startup from URLs given to `--import-blocks-from`.

Closes #381 

## Changes

The `--import-blocks-from` configuration option now accepts URLs as arguments, in addition to local file paths. If the argument can be parsed as a URL, the node will attempt to download the file from that URL during startup and use it to perform out-of-band catch-up.

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -3,12 +3,13 @@
 extern crate log;
 
 // Force the system allocator on every platform
-use futures::FutureExt;
-use std::{alloc::System, convert::identity};
+use futures::{stream::StreamExt, FutureExt};
+use std::{alloc::System, io::Write};
 #[global_allocator]
 static A: System = System;
 
 use anyhow::Context;
+use chrono::Utc;
 use concordium_node::{
     common::PeerType,
     configuration as config,
@@ -36,7 +37,7 @@ use concordium_node::{
 use mio::{net::TcpListener, Poll};
 use parking_lot::Mutex as ParkingMutex;
 use rand::Rng;
-use std::{io::Cursor, sync::Arc, thread::JoinHandle};
+use std::{sync::Arc, thread::JoinHandle};
 #[cfg(unix)]
 use tokio::signal::unix as unix_signal;
 #[cfg(windows)]
@@ -141,28 +142,42 @@ async fn main() -> anyhow::Result<()> {
 
     // Out-of-band catch-up
     if let Some(ref import_blocks_from) = conf.cli.baker.import_blocks_from {
-        let import_path =
-            data_dir_path.to_path_buf().join(concordium_node::configuration::CATCHUP_FILE);
-        let from = match Url::parse(import_blocks_from) {
-            Ok(import_url) => {
-                info!("Downloading catch-up files");
-                let res: Result<(), Box<dyn std::error::Error + Send + Sync>> = async {
-                    let mut file = std::fs::File::create(import_path.clone())?;
-                    let response = reqwest::get(import_url).await?;
-                    let mut content = Cursor::new(response.bytes().await?);
-                    std::io::copy(&mut content, &mut file)?;
-                    Ok(())
+        let from = if let Ok(import_url) = Url::parse(import_blocks_from) {
+            let default_filename = config::CATCHUP_FILE_BASENAME.to_owned()
+                + "_"
+                + &Utc::now().to_string()
+                + &config::CATCHUP_FILE_EXT;
+            let filename = import_url
+                .path_segments()
+                .and_then(|x| x.last())
+                .map(|x| x.to_string())
+                .unwrap_or(default_filename.clone());
+            let import_path = data_dir_path.to_path_buf().join(&filename);
+            let location_msg = if filename.eq(&default_filename) {
+                " to ".to_owned() + &import_path.display().to_string()
+            } else {
+                "".to_owned()
+            };
+            info!("Downloading the catch-up file from {}{}", import_url, location_msg);
+            let mut file = std::fs::File::create(&import_path)?;
+            let res: anyhow::Result<()> = async {
+                let mut stream = reqwest::get(import_url).await?.bytes_stream();
+                while let Some(bytes) = stream.next().await {
+                    let bytes = bytes?;
+                    file.write_all(&bytes)?;
                 }
-                .await;
-                match res {
-                    Ok(_) => import_path.to_str().map_or_else(|| import_blocks_from, identity),
-                    Err(_) => {
-                        info!("Downloading catch-up files failed");
-                        import_blocks_from
-                    }
+                Ok(())
+            }
+            .await;
+            match res {
+                Ok(()) => import_path.display().to_string(),
+                Err(e) => {
+                    error!("Downloading catch-up files failed: {}", e);
+                    import_blocks_from.to_string()
                 }
             }
-            Err(_) => import_blocks_from,
+        } else {
+            import_blocks_from.to_string()
         };
         info!("Starting out of band catch-up");
         consensus.import_blocks(from.as_bytes());
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -459,8 +459,7 @@ fn start_consensus_message_threads(
 
 fn handle_queue_stop<F>(msg: QueueMsg<ConsensusMessage>, dir: &'static str, f: F) -> bool
 where
-    F: FnOnce(ConsensusMessage) -> anyhow::Result<()>,
-{
+    F: FnOnce(ConsensusMessage) -> anyhow::Result<()>, {
     match msg {
         QueueMsg::Relay(msg) => {
             if let Err(e) = f(msg) {
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -142,46 +142,47 @@ async fn main() -> anyhow::Result<()> {
 
     // Out-of-band catch-up
     if let Some(ref import_blocks_from) = conf.cli.baker.import_blocks_from {
-        let from = if let Ok(import_url) = Url::parse(import_blocks_from) {
-            let default_filename = config::CATCHUP_FILE_BASENAME.to_owned()
-                + "_"
-                + &Utc::now().to_string()
-                + &config::CATCHUP_FILE_EXT;
+        let from: anyhow::Result<String> = if let Ok(import_url) = Url::parse(import_blocks_from) {
+            let default_filename = format!(
+                "{}_{}{}",
+                config::CATCHUP_FILE_BASENAME,
+                Utc::now().timestamp(),
+                config::CATCHUP_FILE_EXT
+            );
             let filename = import_url
                 .path_segments()
                 .and_then(|x| x.last())
                 .map(|x| x.to_string())
-                .unwrap_or(default_filename.clone());
+                .unwrap_or(default_filename);
             let import_path = data_dir_path.to_path_buf().join(&filename);
-            let location_msg = if filename.eq(&default_filename) {
-                " to ".to_owned() + &import_path.display().to_string()
-            } else {
-                "".to_owned()
-            };
-            info!("Downloading the catch-up file from {}{}", import_url, location_msg);
-            let mut file = std::fs::File::create(&import_path)?;
-            let res: anyhow::Result<()> = async {
-                let mut stream = reqwest::get(import_url).await?.bytes_stream();
-                while let Some(bytes) = stream.next().await {
-                    let bytes = bytes?;
-                    file.write_all(&bytes)?;
-                }
-                Ok(())
+
+            info!(
+                "Downloading the catch-up file from {} to {}",
+                import_url,
+                import_path.display().to_string()
+            );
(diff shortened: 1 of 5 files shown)


concordium-node/src/configuration.rs
```diff
@@ -76,6 +76,8 @@ pub const SOFT_BAN_DURATION_SECS: u64 = 300;
 pub const MAX_PEER_NETWORKS: usize = 20;
 /// Database subdirectory name
 pub const DATABASE_SUB_DIRECTORY_NAME: &str = "database-v4";
+/// Default out-of-band catch-up file name
+pub const CATCHUP_FILE: &str = "blocks_to_import.mdb";
```
abizjak, line 80: I don't think we should be having this default. It just makes the node behaviour harder to understand. It does not improve the use of the node.

concordium-node/src/configuration.rs
```diff
@@ -246,10 +248,11 @@ pub struct BakerConfig {
     pub transactions_purging_delay: u32,
     #[structopt(
         long = "import-blocks-from",
-        help = "Path to a file exported by the database exporter",
+        help = "Where to find a file exported by the database exporter. \
```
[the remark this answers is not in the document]
abizjak: I think the previous phrasing was better than the current one.

concordium-node/src/configuration.rs
```diff
@@ -746,19 +749,19 @@ pub struct MacOsConfig {
...
+    pub connection: ConnectionConfig,
     #[structopt(flatten)]
-    pub cli:          CliConfig,
+    pub cli: CliConfig,
     #[structopt(flatten)]
     pub bootstrapper: BootstrapperConfig,
     #[cfg(target_os = "macos")]
     #[structopt(flatten)]
-    pub macos:        MacOsConfig,
+    pub macos: MacOsConfig,
```
abizjak, line 764: This reformatting here suggests you have the wrong rust formatter. Please format with the correct nightly version as described in the README.

concordium-node/src/bin/cli.rs
```diff
@@ -152,9 +153,33 @@ async fn main() -> anyhow::Result<()> {
         None
     };
 
-    if let Some(ref import_path) = conf.cli.baker.import_path {
+    // Out-of-band catch-up
+    if let Some(ref import_blocks_from) = conf.cli.baker.import_blocks_from {
+        let import_path =
+            data_dir_path.to_path_buf().join(concordium_node::configuration::CATCHUP_FILE);
```
abizjak, line 159: Please remove this default name. It just adds "magical" behaviour to the node for no good reason.

I'd recommend using the file name from the URL using https://docs.rs/url/latest/url/struct.Url.html#method.path_segments and only if it is empty or unusable try a default (which would ideally have a temporal component for clarity), making sure to log extensively that this is what you are doing.

concordium-node/src/bin/cli.rs
```diff
@@ -152,9 +153,33 @@ async fn main() -> anyhow::Result<()> {
...
+        let from = match Url::parse(import_blocks_from) {
+            Ok(import_url) => {
+                info!("Downloading catch-up files");
```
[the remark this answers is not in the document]
abizjak: Yes, please add the url
```rust
info!("Downloading catch-up file from {}.", import_url)
```

concordium-node/src/bin/cli.rs
```diff
@@ -152,9 +153,33 @@ async fn main() -> anyhow::Result<()> {
...
+                let res: Result<(), Box<dyn std::error::Error + Send + Sync>> = async {
+                    let mut file = std::fs::File::create(import_path.clone())?;
+                    let response = reqwest::get(import_url).await?;
+                    let mut content = Cursor::new(response.bytes().await?);
+                    std::io::copy(&mut content, &mut file)?;
```
abizjak, line 167: Since you are `await`ing the response, I'm guessing this will store the entire contents of the file in memory before writing it out. You should probably use https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes_stream instead.

abizjak, line 167: Instead of loading the entire file into memory you should stream it into a file. Something like

```rust
response.bytes_stream().try_for_each(|bytes| async { file.write_all(&bytes) }).await?
```

abizjak: https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each

concordium-node/src/bin/cli.rs
```diff
@@ -152,9 +153,33 @@ async fn main() -> anyhow::Result<()> {
...
+                    Ok(())
+                }
+                .await;
+                match res {
+                    Ok(_) => import_path.to_str().map_or_else(|| import_blocks_from, identity),
+                    Err(_) => {
+                        info!("Downloading catch-up files failed");
```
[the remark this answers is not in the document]
abizjak: We should do that, yes.

concordium-node/src/bin/cli.rs
```diff
@@ -152,9 +153,33 @@ async fn main() -> anyhow::Result<()> {
...
+                let res: Result<(), Box<dyn std::error::Error + Send + Sync>> = async {
+                    let mut file = std::fs::File::create(import_path.clone())?;
+                    let response = reqwest::get(import_url).await?;
+                    let mut content = Cursor::new(response.bytes().await?);
+                    std::io::copy(&mut content, &mut file)?;
+                    Ok(())
+                }
+                .await;
+                match res {
+                    Ok(_) => import_path.to_str().map_or_else(|| import_blocks_from, identity),
```
abizjak, line 172: Please use `Ok(())` to make it clear that we are not ignoring anything.

That, or `if let Err(e) = ... then {...} else {...}`

concordium-node/src/bin/cli.rs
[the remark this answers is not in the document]
abizjak: That would also be an `error` log, not `info`.

concordium-node/src/bin/cli.rs
```diff
@@ -152,9 +153,33 @@ async fn main() -> anyhow::Result<()> {
...
+                    Err(_) => {
+                        info!("Downloading catch-up files failed");
+                        import_blocks_from
+                    }
+                }
```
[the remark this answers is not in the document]
abizjak: The current behaviour is to ignore (non)existence of the file. So this is consistent.

concordium-node/src/configuration.rs
```diff
@@ -18,7 +18,7 @@ use structopt::{clap::AppSettings, StructOpt};
 
 /// Client's details for local directory setup purposes.
 pub const APP_INFO: AppInfo = AppInfo {
-    name:   "concordium",
+    name: "concordium",
```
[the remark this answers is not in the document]
abizjak: Most nightly versions will actually work. It's only the stable one that does not support the configurations we use.

abizjak (dismissed):
I am wondering about a general design here. You are downloading this file from the Rust part of the node and this works OK when it is just one file.

However all the knowledge about what state the node is in resides in the Haskell part. So if we want to evolve this into something more clever where the node would figure out which chunks it needs to download then this downloading functionality should reside in the Haskell part.

If it does, then we can also start processing blocks while they are being downloaded, as opposed to now, where we download them all and then, and only then, start processing them.

td202:
Perhaps if the node downloads a catch-up file then it should delete it after import.

limemloh (changes requested):
I have added some comments.
You should also make sure to run `cargo clippy` and fix issues.

concordium-node/src/bin/cli.rs
```diff
@@ -138,10 +140,49 @@ async fn main() -> anyhow::Result<()> {
...
+                import_url,
+                import_path.display().to_string()
+            );
+            let file = std::fs::File::create(&import_path)?;
+            let mut buffer = std::io::BufWriter::new(file);
+            let mut stream = reqwest::get(import_url).await?.bytes_stream();
+            while let Some(Ok(bytes)) = stream.next().await {
+                buffer.write_all(&bytes)?;
+            }
+            buffer.flush()?;
```
limemloh, line 170: The node will now stop if any of these result in an error, since these are no longer in a closure.
We don't want the node to stop, but just log some error.

concordium-node/src/bin/cli.rs
```diff
@@ -138,10 +140,49 @@ async fn main() -> anyhow::Result<()> {
...
+                "{}_{}{}",
+                config::CATCHUP_FILE_BASENAME,
+                Utc::now().timestamp(),
+                config::CATCHUP_FILE_EXT
+            );
+            let filename = import_url
+                .path_segments()
+                .and_then(|x| x.last())
+                .map(|x| x.to_string())
+                .unwrap_or(default_filename);
```
limemloh, line 156: Sorry for not commenting this the first time around, but I don't think the URL should be used to construct the file name.
Always using the `default_filename` is more predictable.

concordium-node/src/bin/cli.rs
```diff
@@ -138,10 +140,49 @@ async fn main() -> anyhow::Result<()> {
         None
     };
 
-    if let Some(ref import_path) = conf.cli.baker.import_path {
-        info!("Starting out of band catch-up");
-        consensus.import_blocks(import_path.as_bytes());
-        info!("Completed out of band catch-up");
+    // Out-of-band catch-up
+    if let Some(ref import_blocks_from) = conf.cli.baker.import_blocks_from {
```
td202, line 144: This should probably be factored out in a separate function, possibly on `ConsensusContainer`. Errors should be handled gracefully, and we probably shouldn't stop the node if we can recover, even if download/import was not successful.

concordium-node/src/bin/cli.rs
```diff
@@ -138,10 +140,49 @@ async fn main() -> anyhow::Result<()> {
...
+            buffer.flush()?;
+
+            Ok(import_path.display().to_string())
+        } else {
+            Ok(import_blocks_from.to_string())
+        };
+        match from {
+            Ok(from) => {
+                info!("Starting out of band catch-up");
+                consensus.import_blocks(from.as_bytes());
```
td202, line 179: If we downloaded the catch-up file, then we should probably also delete it after catch-up is done.

limemloh (changes requested):

concordium-node/src/configuration.rs
```diff
@@ -76,6 +76,10 @@ pub const SOFT_BAN_DURATION_SECS: u64 = 300;
 pub const MAX_PEER_NETWORKS: usize = 20;
 /// Database subdirectory name
 pub const DATABASE_SUB_DIRECTORY_NAME: &str = "database-v4";
+/// Default out-of-band catch-up file name (excluding the extension)
+pub const CATCHUP_FILE_BASENAME: &str = "blocks_to_import";
+/// Default out-of-band catch-up file extension
+pub const CATCHUP_FILE_EXT: &str = ".mdb";
```
td202, line 82: I don't think we should use the extension "mdb", because it suggests it's an LMDB database or a Microsoft Access database, and it's neither of those things. Perhaps a more neutral "db".

concordium-node/src/bin/cli.rs
mkmks: Deferred to #437

concordium-node/src/bin/cli.rs
mkmks: Not sure if the downloader function should be part of `ConsensusContainer` but I factored it out for the sake of returning a `Result` instead of stopping the node.

concordium-node/src/bin/cli.rs
```diff
@@ -138,10 +140,23 @@ async fn main() -> anyhow::Result<()> {
...
+    // Out-of-band catch-up
+    if let Some(ref import_blocks_from) = conf.cli.baker.import_blocks_from {
+        let from: anyhow::Result<String> =
+            download_catchup_files(import_blocks_from, data_dir_path).await;
+        match from {
+            Ok(from) => {
+                info!("Starting out of band catch-up");
+                consensus.import_blocks(from.as_bytes());
+                info!("Completed out of band catch-up");
+                if let Err(e) = std::fs::remove_file(from) {
```
limemloh, line 152: We only want to delete the file when it was provided as a URL and downloaded by the node.

limemloh (changes requested):
It should not delete a user provided catchup file, only files downloaded by the node itself should be removed

limemloh (approved):

PR #394, opened 2022-06-15
Correct absolute height

### LINKED ISSUE #393: Incorrectly reported "absoluteBlockHeight"
**Bug Description**

The 4.2 node reports incorrect absolute block heights when more than 2 protocol updates have happened on the chain.

Reported by @smh1001 

**Versions**
- Software Version: 4.2

abizjak:
## Purpose

Fixes #393 

## Changes

After each protocol version we record the absolute height of the last finalized
block before the new genesis. This is so we can have a single set of heights.

During startup we did not record this height correctly.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
How is this related to the windows node installer?

abizjak:
> How is this related to the windows node installer?

What do you mean?

You mean updating the version? IN the interest of time I also bumped the version of the distribution. I believe for such a small change it is better to avoid the needless delays of a separate PR.

PR #399, opened 2022-06-20
Remove finalized blocks (apart from the last one) entirely from memory.

abizjak:
## Purpose

Remove memory use and startup time by not storing statuses in memory.

Memory use on mainnet (ubuntu distribution) drops to about 1.6GB a bit after startup.

## Changes

- Do not store finalized block statuses in the in-memory hashmap. This enables faster startup because we do not have to load them on startup. It could in principle make some lookups slower, but in general we do not need finalized blocks apart from the last one, which is already cached.
- Do not store dead blocks in the block table, apart from 1000 recent ones.This is a heuristic to perhaps deduplicate blocks in cases of a lot branching. Other approaches here would be possible, this just seemed a particularly simple one with minimal overhead.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Paired.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Paired.hs
@@ -1015,6 +1015,7 @@ instance (C.HasGlobalStateContext (PairGSContext lc rc) r,
             (RecentBlock (BlockPending pb1), RecentBlock (BlockPending pb2)) -> assertEq pb1 pb2 $ return $ RecentBlock (BlockPending pb1)
             (Unknown, RecentBlock BlockDead) -> return Unknown
             (RecentBlock BlockDead, Unknown) -> return Unknown
+            (RecentBlock BlockDead, RecentBlock BlockDead) -> return Unknown
             (OlderThanLastFinalized, OlderThanLastFinalized) -> return OlderThanLastFinalized
             _ -> error $ "getBlockStatus (Paired): block statuses do not match: " ++ show bs1 ++ ", " ++ show bs2
 
--- a/concordium-consensus/src/Concordium/GlobalState/Paired.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Paired.hs
@@ -1004,18 +1004,18 @@ instance (C.HasGlobalStateContext (PairGSContext lc rc) r,
             (Just BlockDead, Nothing) -> return Nothing
             _ -> error $ "getBlockStatus (Paired): block statuses do not match: " ++ show bs1 ++ ", " ++ show bs2
 
-    getBlockStatusOrOld bh = do
-        bs1 <- coerceGSML $ getBlockStatusOrOld bh
-        bs2 <- coerceGSMR $ getBlockStatusOrOld bh
+    getRecentBlockStatus bh = do
+        bs1 <- coerceGSML $ getRecentBlockStatus bh
+        bs2 <- coerceGSMR $ getRecentBlockStatus bh
         case (bs1, bs2) of
-            (Nothing, Nothing) -> return Nothing
-            (Just (Right (BlockAlive bp1)), Just (Right (BlockAlive bp2))) -> return $ Just (Right (BlockAlive (PairBlockData (bp1, bp2))))
-            (Just (Right (BlockFinalized bp1 fr1)), Just (Right (BlockFinalized bp2 fr2))) ->
-                assertEq fr1 fr2 $ return $ Just $ Right $ BlockFinalized (PairBlockData (bp1, bp2)) fr1
-            (Just (Right (BlockPending pb1)), Just (Right (BlockPending pb2))) -> assertEq pb1 pb2 $ return $ Just $ Right (BlockPending pb1)
-            (Nothing, Just (Right BlockDead)) -> return Nothing
-            (Just (Right BlockDead), Nothing) -> return Nothing
-            (Just (Left ()), Just (Left ())) -> return Nothing
+            (Unknown, Unknown) -> return Unknown
+            (RecentBlock (BlockAlive bp1), RecentBlock (BlockAlive bp2)) -> return $ RecentBlock (BlockAlive (PairBlockData (bp1, bp2)))
+            (RecentBlock (BlockFinalized bp1 fr1), RecentBlock (BlockFinalized bp2 fr2)) ->
+                assertEq fr1 fr2 $ return $ RecentBlock $ BlockFinalized (PairBlockData (bp1, bp2)) fr1
+            (RecentBlock (BlockPending pb1), RecentBlock (BlockPending pb2)) -> assertEq pb1 pb2 $ return $ RecentBlock (BlockPending pb1)
+            (Unknown, RecentBlock BlockDead) -> return Unknown
+            (RecentBlock BlockDead, Unknown) -> return Unknown
+            (OlderThanLastFinalized, OlderThanLastFinalized) -> return OlderThanLastFinalized
             _ -> error $ "getBlockStatus (Paired): block statuses do not match: " ++ show bs1 ++ ", " ++ show bs2
 
     makeLiveBlock pb (PairBlockData (parent1, parent2)) (PairBlockData (lf1, lf2)) (bs1, bs2) t e = do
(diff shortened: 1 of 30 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -490,37 +562,65 @@ instance (MonadLogger (PersistentTreeStateMonad bs m),
...
+    markDead bh = do
+            blockTable . liveMap . at' bh .=! Nothing
+            blockTable . deadCache %=! insertDeadCache bh
     type MarkFin (PersistentTreeStateMonad bs m) = Maybe (StoredBlock (MPV m) (BlockStatePointer bs))
-    markFinalized bh fr = use (blockTable . at' bh) >>= \case
+    markFinalized bh fr = use (blockTable . liveMap . at' bh) >>= \case
             Just (BlockAlive bp) -> do
               st <- saveBlockState (_bpState bp)
-              blockTable . at' bh ?=! BlockFinalized (finalizationIndex fr)
+              -- NB: This only makes sense if no block lookups are done between
```
EmilLa1, line 606: What does this mean?

concordium-node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "concordium_node"
-version = "4.2.1" # must be kept in sync with 'is_compatible_version' in 'src/configuration.rs'
+version = "4.2.2" # must be kept in sync with 'is_compatible_version' in 'src/configuration.rs'
```
EmilLa1, line 3: Why is this part of the PR?

EmilLa1 (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -490,37 +562,65 @@ instance (MonadLogger (PersistentTreeStateMonad bs m),
...
+              -- the call to this function and 'wrapUpFinalization'. This is
+              -- currently the case, and must remain the case in the future.
+              -- That is, finalization must remain atomic, or this handling, and
+              -- wrapUpFinalization must be changed.
```
td202, line 610: The documentation of `markFinalized` should be updated to reflect that the block is not properly finalized until `wrapupFinalization` is called, and thus looking up the block with `getBlockStatus` in the interim could give the wrong result.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -490,37 +562,65 @@ instance (MonadLogger (PersistentTreeStateMonad bs m),
...
-    markPending pb = blockTable . at' (getHash pb) ?=! BlockPending pb
-    markAllNonFinalizedDead = blockTable %=! fmap nonFinDead
-        where
-            nonFinDead BlockAlive{} = BlockDead
-            nonFinDead BlockPending{} = BlockDead
-            nonFinDead o = o
+    markPending pb = blockTable . liveMap . at' (getHash pb) ?=! BlockPending pb
+    markAllNonFinalizedDead = do
+      blockTable . liveMap .=! HM.empty
+      blockTable . deadCache .=! emptyDeadCache
```
td202, line 622: This behaviour seems not at all in the spirit of how the dead cache should operate. This is probably fine for the context the function is used in, but the documentation (and possibly the function name) should be updated to reflect this.

concordium-consensus/src/Concordium/GlobalState/TreeState.hs
```diff
@@ -85,6 +85,20 @@ data AddTransactionResult =
...
+-- |Status of a "recent" block. Recent here means that instead of looking up the
+-- full status of a block that is older than the last finalized one, we return
+-- only the fact that it is older than last finalized. This performs better in
+-- cases where this is the only information that is needed. It avoids loading
+-- finalized blocks from the database.
+data RecentBlockStatus bp pb  =
+  -- |The block is either pending, dead, or no older than the last finalized block.
+  RecentBlock !(BlockStatus bp pb)
+  -- |The block is known, and is strictly older than the last finalized block.
+  | OlderThanLastFinalized
```
td202, line 97: I think a better name like `OldFinalized` would be more appropriate, because this indicates that the block is finalized, which I think should be emphasised. (A dead block can fit the description "strictly older than the last finalized block".)

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -272,24 +272,30 @@ addBlock block txvers = do
...
-            parentStatus <- getBlockStatus parent
+            -- Look up the block parent, if any.
+            -- This is performance sensitive since it is before we managed to validate the block,
+            -- so we make sure to look up as little data as possible to determine block validity.
+            -- In particular we do not use getBlockStatus since that loads the entire block
+            -- from the database if the block is finalized.
+            parentStatus <- getRecentBlockStatus parent
             case parentStatus of
-                Nothing -> do
+                -- The block's parent is already beyond the last finalized one. So
```
td202, line 282:
```suggestion
                -- The block's parent is already older than the last finalized one. So
```

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -272,24 +272,30 @@ addBlock block txvers = do
         lfs <- getLastFinalizedSlot
         -- The block must be later than the last finalized block
         if lfs >= blockSlot block then deadBlock else do
-            parentStatus <- getBlockStatus parent
+            -- Look up the block parent, if any.
+            -- This is performance sensitive since it is before we managed to validate the block,
+            -- so we make sure to look up as little data as possible to determine block validity.
+            -- In particular we do not use getBlockStatus since that loads the entire block
+            -- from the database if the block is finalized.
```
td202, line 279: I don't think this commentary is useful here.

td202 (approved):

concordium-consensus/src/Concordium/GlobalState/TreeState.hs
```diff
@@ -120,6 +134,10 @@ class (Eq (BlockPointerType m),
     -- * Operations on the block table
     -- |Get the current status of a block.
     getBlockStatus :: BlockHash -> m (Maybe (BlockStatus (BlockPointerType m) PendingBlock))
+
+    -- |Get the current status of a block.
```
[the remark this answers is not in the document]
abizjak: I elaborated.

concordium-consensus/src/Concordium/GlobalState/Paired.hs
```diff
@@ -997,11 +997,27 @@ instance (C.HasGlobalStateContext (PairGSContext lc rc) r,
...
+    getRecentBlockStatus bh = do
+        bs1 <- coerceGSML $ getRecentBlockStatus bh
+        bs2 <- coerceGSMR $ getRecentBlockStatus bh
+        case (bs1, bs2) of
+            (Unknown, Unknown) -> return Unknown
+            (RecentBlock (BlockAlive bp1), RecentBlock (BlockAlive bp2)) -> return $ RecentBlock (BlockAlive (PairBlockData (bp1, bp2)))
+            (RecentBlock (BlockFinalized bp1 fr1), RecentBlock (BlockFinalized bp2 fr2)) ->
+                assertEq fr1 fr2 $ return $ RecentBlock $ BlockFinalized (PairBlockData (bp1, bp2)) fr1
+            (RecentBlock (BlockPending pb1), RecentBlock (BlockPending pb2)) -> assertEq pb1 pb2 $ return $ RecentBlock (BlockPending pb1)
+            (Unknown, RecentBlock BlockDead) -> return Unknown
```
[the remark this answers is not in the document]
abizjak: It seems like a reasonable choice.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -547,6 +548,17 @@ memberTransactionTable th = do
     $ transaction (dbh ^. storeEnv) True
     $ \txn -> isRecordPresent txn (dbh ^. transactionStatusStore) th
 
+-- |Check if a block with the given hash is stored in the LMDB block store.
+memberBlockTable ::
```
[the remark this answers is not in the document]
abizjak: Ok.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
abizjak: I updated the comment to reduce the scope of the function (and renamed it).

concordium-consensus/src/Concordium/GlobalState/TreeState.hs
abizjak: Ok.

concordium-consensus/src/Concordium/Skov/Update.hs
abizjak: I think it does bring value since it points out that specific issue that needs care and that that is why the functionality is implemented in a way that is slightly more cumbersome than what the reader might expect.

Issue #375, opened 2022-06-02
Incorrect RPC error response


abizjak:
**Bug Description**

If a transaction is submitted to the node such that the sender account has insufficient funds the error that is returned is "energy too low" which is confusing and incorrect.

**Steps to Reproduce**

Attempt to submit a transaction from an account that has insufficient funds to cover the costs.

**Expected Result**

Get an error like "insufficient funds"

**Actual Result**

Get "energy too low" as a response.

**Versions**
- Software Version: 4.*

abizjak:
These are the culprits
```
transactionVerificationResultToUpdateResult (TV.NotOk TV.NormalTransactionDepositInsufficient) = ResultTooLowEnergy
transactionVerificationResultToUpdateResult (TV.MaybeOk TV.NormalTransactionInsufficientFunds) = ResultTooLowEnergy
```
in Update.hs

PR #401, opened 2022-06-20
Fix incorrect rpc response for when user have insufficient funds.

EmilLa1:
## Purpose

Fixes https://github.com/Concordium/concordium-node/issues/375 by introducing a new `ResultInsufficientFunds` error when the user do not have enough funds to cover the costs of its transaction.

## Changes

Introduced a new FFI result `ResultInsufficientFunds` and vice versa a new rpc error which will result in a rpc `InvalidArgument` error that will yield the message: "The sender did not have enough funds to cover the costs".

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 - Decrease memory usage by not storing genesis blocks. This has the effect that
   the database produced by node versions >= 4.2.* cannot be used by node
   versions <= 4.1. The other direction works.
+- Fix issue #375 which resulted in a wrong rpc response when the sender of a 
+  transaction did not have enough funds to cover the costs.
 
 ## 4.0.11
 - The `SendTransaction` function exposed via the gRPC interface now provides the caller with detailed error messages if the 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Decrease memory usage by not storing genesis blocks. This has the effect that
   the database produced by node versions >= 4.2.* cannot be used by node
   versions <= 4.1. The other direction works.
-- Fix issue #375 which resulted in a wrong rpc response when the sender of a 
+- Fix an issue which resulted in a wrong rpc response when the sender of a
   transaction did not have enough funds to cover the costs.
 
 ## 4.0.11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Decrease memory usage by not storing genesis blocks. This has the effect that
   the database produced by node versions >= 4.2.* cannot be used by node
   versions <= 4.1. The other direction works.
-- The gRPC API reports correctly when the sender of a transaction did 
+- The gRPC API now reports correctly when the sender of a transaction did 
   not have enough funds to cover the transaction costs.
 
 ## 4.0.11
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Decrease memory usage by not storing genesis blocks. This has the effect that
   the database produced by node versions >= 4.2.* cannot be used by node
   versions <= 4.1. The other direction works.
+- Fix issue #375 which resulted in a wrong rpc response when the sender of a
```
abizjak, line 12: The changelog should not have issue numbers. Please explain what it gives to the user in plain terms.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -685,7 +685,7 @@ transactionVerificationResultToUpdateResult (TV.NotOk TV.CredentialDeploymentInv
 transactionVerificationResultToUpdateResult (TV.NotOk (TV.ChainUpdateSequenceNumberTooOld _)) = ResultChainUpdateSequenceNumberTooOld
 transactionVerificationResultToUpdateResult (TV.NotOk TV.ChainUpdateEffectiveTimeBeforeTimeout) = ResultChainUpdateInvalidEffectiveTime
 transactionVerificationResultToUpdateResult (TV.NotOk TV.CredentialDeploymentExpired) = ResultCredentialDeploymentExpired
-transactionVerificationResultToUpdateResult (TV.NotOk TV.NormalTransactionDepositInsufficient) = ResultTooLowEnergy
+transactionVerificationResultToUpdateResult (TV.NotOk TV.NormalTransactionDepositInsufficient) = ResultInsufficientFunds
```
abizjak, line 688: I don't think this is correct. This error is when the NRG amount is less than the minimal one based on transaction size https://github.com/Concordium/concordium-node/blob/1b088cb37b76f329975f73fab9523333e3fb38bc/concordium-consensus/src/Concordium/TransactionVerification.hs#L273

abizjak (changes requested):

abizjak (approved):

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -353,6 +357,9 @@ impl TryFrom<i64> for ConsensusFfiResponse {
             27 => Ok(ChainUpdateSequenceNumberTooOld),
             28 => Ok(ChainUpdateInvalidSignatures),
             29 => Ok(MaxBlockEnergyExceeded),
+            // 30 is used for`ResultImportStopped` which origins from a `SIGINT` or `SIGTERM`
+            // signal send to the node while catching up out of band.
+            31 => Ok(InsufficientFunds),
```
abizjak, line 362: Actually, looking at this now, the handling of StopImport is broken in the following ways.
- The error is not handled in the Rust side. This is generally fine since the node will stop at this point. But why return this status at all then? Why not just return "Ok".
- The return value of `importBlocks` is declared `u8` on the Rust side, but it is actually `u64` on the Haskell side.

I would recommend removing the "Import Stopped" message and returning Ok instead. And fixing the return value of `importBlocks` on the Rust side.

abizjak: I can do that in a followup PR if you prefer.

EmilLa1: I'll just fix it now I'm at it.

abizjak (approved):

PR #421, opened 2022-06-27
Account Cache 

EmilLa1:
## Purpose

Support caching for accounts. Once written to disk, accounts are not held in memory except for in a cache, which may hold a limited number of accounts. This reduces the amount of memory required per account to around 3685 bytes. (Without the cache, and with only initial accounts, the memory requirement per account is approximately 8240 bytes.) Note, this measurement is during normal operation, and does not account for the extra memory usage at initialisation or protocol updates.

## Changes

- Introduce an "accounts cache size" configuration parameter that determines the number of accounts that are cached in memory. This is configured by the `--accounts-cache-size` command-line argument or the `CONCORDIUM_NODE_CONSENSUS_ACCOUNTS_CACHE_SIZE` environment variable. In the Rust code, a `ConsensusRuntimeParameters` struct is introduced to manage the burgeoning number of parameters.
- Introduce an abstract `Cache` class and a `FIFOCache` implementation. The implementation is targeted to caching values that are keyed by `BlobRef`s - that is, pointers in the block state database.
- Introduce a `CachedRef`, that wraps a `BlobRef`, and will attempt to retrieve a value from a cache before loading it from the `BlobStore` if that fails. When a value is loaded, it is added to the cache. (A `CachedRef` can exist in memory before it is written to the disk, but the memory reference is only retained by the cache after it is written.)
- Introduce an `EagerlyHashedCachedRef` that retains the hash of the value.
- Introduce an `EagerBufferedRef` that always retains the value in memory.
- Revise the behaviour of `HashedBufferedRef` to compute the hash only on demand. This, in principle, reduces the amount of hashing required when, for instance, an account is updated multiple times in a single block.
- Introduce an `EagerlyHashedBufferedRef` that retains both the value and its hash in memory.
- Refactor the persistent LFMBTree implementation so that leaf values are not stored under the reference type. This allows using a different reference types at the leaves.
- Revise the account table to use `EagerlyHashedCachedRef`s at the leafs and `HashedBufferedRef`s at the branches. This ensures that accounts are only retained in memory by the cache, but their hashes are retained for updating the hash of the account table.
- Revise persistent accounts to use `EagerBufferedRef`s for the encrypted amount, release schedule and stake, and `EagerlyHashedBufferedRef`s for the persisting account data. This enforces loading these when the account is loaded, but is generally convenient since they are used if the hash needs to be recomputed (e.g. when the account is modified).
- Add the account cache to the `PersistentBlockStateContext`.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
```diff
@@ -36,9 +36,12 @@ import qualified Concordium.GlobalState.Basic.BlockState.Account as Transient
 import qualified Concordium.GlobalState.Basic.BlockState.AccountReleaseSchedule as Transient
 import Concordium.GlobalState.Persistent.BlobStore
 import Concordium.GlobalState.Persistent.BlockState.AccountReleaseSchedule
+import Concordium.GlobalState.Persistent.CachedRef
 import Concordium.GlobalState.Account hiding (addIncomingEncryptedAmount, addToSelfEncryptedAmount)
 
 -- | The persistent version of the encrypted amount structure per account.
+-- We use 'EagerBufferedRef's for the encrypted amounts so that when a 'PersistentAccount' is
+-- loaded, the entire encrypted amount will also be loaded.
```
abizjak, line 44: While this states the "what", the more important thing is the "why". It cannot be because of expected usage pattern, so it must be due to hashing. Please document intent.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
```diff
@@ -178,18 +177,16 @@ makeLenses ''PersistentExtraBakerInfo
...
-instance forall av m. (IsAccountVersion av, MonadBlobStore m) => Cacheable m (PersistentExtraBakerInfo av) where
-  cache = case accountVersion @av of
-    SAccountV0 -> pure
-    SAccountV1 -> theExtraBakerInfo cache
+instance forall av m. (Applicative m) => Cacheable m (PersistentExtraBakerInfo av)
 
 -- |A persistent version of 'BakerInfoEx'.
-
+-- IMPORTANT NOTE: The 'Cacheable' instance relies on this not requiring recursive caching.
+-- If this changes, the instance for 'BakerInfos' should also be updated.
```
abizjak, line 187: Similar to the other comment I have, the cacheable instance of this type is itself trivial. So is the comment talking about some other instances. If so, which ones?

concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
```diff
@@ -239,14 +236,16 @@ instance forall m av. (IsAccountVersion av, MonadBlobStore m) => BlobStorable m
...
 
-instance (IsAccountVersion av, MonadBlobStore m) => Cacheable m (PersistentBakerInfoEx av) where
-  cache PersistentBakerInfoEx{..} = 
-    PersistentBakerInfoEx <$> cache bakerInfoRef <*> cache bakerInfoExtra
+instance (Applicative m) => Cacheable m (PersistentBakerInfoEx av)
 
+-- |A baker associated with an account.
+-- IMPORTANT NOTE: The 'Cacheable' instance relies on the fact that no recursive caching is
+-- necessary (due to the use of 'EagerBufferedRef's). If this changes, the instance for
+-- 'PersistentAccountStake' will also need to be updated.
```
abizjak, line 244: Also here.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
```diff
@@ -344,10 +340,14 @@ putAccountBaker PersistentAccountBaker{..} = do
       put abie
       put _bakerPendingChange
 
+-- |Staking information associated with an account.
+-- IMPORTANT NOTE: The 'Cacheable' instance relies on the fact that no recursive caching is
+-- necessary (due to the use of 'EagerBufferedRef's). If this changes, the instance for
+-- 'PersistentAccount' will also need to be updated.
```
abizjak, line 346: And here.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
```diff
@@ -359,16 +359,15 @@ instance forall m av. (MonadBlobStore m, IsAccountVersion av) => BlobStorable m
         su0 :: PersistentAccountStake 'AccountV0 -> m (Put, PersistentAccountStake 'AccountV0)
         su0 pas@PersistentAccountStakeNone = return (put (refNull :: BlobRef (PersistentAccountBaker av)), pas)
         su0 (PersistentAccountStakeBaker bkrref) = do
-          (r, bkrref') <- storeUpdate bkrref
+          (!r, !bkrref') <- storeUpdate bkrref
```
abizjak, line 362: Is this based on profiling, or just a guess? We have many other instances like this without forcing evaluation. Should those also be updated?

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -325,14 +326,32 @@ instance HasBlobStore BlobStore where
...
+        (Functor, Applicative, Monad, MonadReader r, MonadIO, MonadFail, MonadLogger)
+        via (ReaderT r m)
+    deriving
+        (MonadTrans)
+        via (ReaderT r)
+
+instance (HasBlobStore r, MonadIO m) => MonadBlobStore (BlobStoreT r m)
+
+alterBlobStoreT :: (r1 -> r2) -> BlobStoreT r2 m a -> BlobStoreT r1 m a
+alterBlobStoreT f (BlobStoreT a) = BlobStoreT (a . f)
```
abizjak, line 342: Documentation.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -2623,24 +2626,34 @@ doGetInitialTransactionTable pbs = do
...
-instance HasBlobStore PersistentBlockStateContext where
+instance HasBlobStore (PersistentBlockStateContext av) where
     blobStore = bscBlobStore . pbscBlobStore
     blobLoadCallback = bscLoadCallback . pbscBlobStore
     blobStoreCallback = bscStoreCallback . pbscBlobStore
 
+instance AccountVersionFor pv ~ av => Cache.HasCache (Accounts.AccountCache av) (PersistentBlockStateContext pv) where
+  projectCache = pbscCache
+
+withNewAccountCache :: (MonadIO m) => Int -> BlobStoreT (PersistentBlockStateContext pv) m a -> BlobStoreT BlobStore m a
```
abizjak, line 2642: Documentation.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -59,26 +69,28 @@ data Accounts (pv :: ProtocolVersion) = Accounts {
     -- |Unique index of accounts by 'AccountAddress'
     accountMap :: !(AccountMap.PersistentAccountMap pv),
     -- |Hashed Merkle-tree of the accounts
-    accountTable :: !(LFMBTree AccountIndex HashedBufferedRef (PersistentAccount (AccountVersionFor pv))),
+    accountTable :: !(LFMBTree' AccountIndex HashedBufferedRef (LazilyHashedCachedRef (AccountCache (AccountVersionFor pv)) (PersistentAccount (AccountVersionFor pv)))),
```
abizjak, line 72: I would like to have a description of the caching approach written in the comment to this data structure.

In particular the use of BlobRef's for keys in the cache is not that great since we maintain many copies of the same account, only one of which is relevant. In some scenarios this is not an issue, in others it reduces the effective size of the cache a lot.

These tradeoffs and the reason for the choices need to be documented so that we may understand them in the future.

concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
```diff
@@ -0,0 +1,328 @@
...
+      CacheValue c ~ a
+    ) =>
+    Reference m (CachedRef c) a
+    where
+    refFlush cr@(CachedRef ioref) = do
+        mbr <- liftIO $ readIORef ioref
+        case mbr of
+            Disk br -> return (cr, br)
+            Mem val -> do
+                (!r, !val') <- {-# SCC "FLUSHCACHE" #-} storeUpdateRef val
```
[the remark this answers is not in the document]
td202: Will remove.

concordium-consensus/tests/globalstate/GlobalStateTests/Cache.hs
```diff
@@ -0,0 +1,84 @@
...
+import Concordium.GlobalState.Persistent.BlobStore (BlobRef (BlobRef))
+import Concordium.GlobalState.Persistent.Cache
+import Control.Monad.Reader
+import Data.Proxy
+import Test.Hspec
+
+-- |A reader monad transformer that is used to provied a 'MonadCache' instance.
+newtype CacheM c m a = CacheM { runCacheM :: c -> m a }
+  deriving (Functor, Applicative, Monad, MonadReader (CacheContext c), MonadIO) via (ReaderT (CacheContext c) m)
+  deriving (MonadTrans) via (ReaderT (CacheContext c))
```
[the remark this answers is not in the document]
td202: The point is that it should have a MonadCache instance that is not lifted from the underlying monad, as is the case with `ReaderT`.

concordium-consensus/src/Concordium/GlobalState/Persistent/Bakers.hs
```diff
@@ -36,6 +36,8 @@ import Concordium.Types.HashableTo
 import Concordium.Utils.Serialization.Put
 
 -- |A list of references to 'BakerInfo's, ordered by increasing 'BakerId'.
+-- IMPORTANT NOTE: The 'Cacheable' instance relies on this not requiring recursive caching through
+-- the use of 'EagerBufferedRef's.
```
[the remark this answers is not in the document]
td202: I've revised this comment. The point is that the datatype is always fully cached in memory, so it is fine for the 'Cacheable' instance to be trivial, and that if something changes this, then it may be necessary to reconsider how the 'Cacheable' instance is implemented.

concordium-consensus/src/Concordium/GlobalState/Persistent/Cache.hs
```diff
@@ -0,0 +1,201 @@
...
+    getCache = lift getCache
+    {-# INLINE getCache #-}
+
+instance (Monoid w, MonadCache c m) => MonadCache c (WriterT w m) where
+    getCache = lift getCache
+    {-# INLINE getCache #-}
+
+instance (MonadCache c m) => MonadCache c (ExceptT e m) where
+    getCache = lift getCache
+    {-# INLINE getCache #-}
```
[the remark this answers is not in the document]
td202: I am not sure what other behaviour would be considered natural. I assume you mean some kind of rollback of the cache in the error case, but I do not think this would be reasonable or natural.

concordium-consensus/src/Concordium/GlobalState/Persistent/Cache.hs
```diff
@@ -0,0 +1,201 @@
...
+
+    newCache = newDummyCache
+    collapseCache _ = return ()
+    putCachedValue _ _ = return
+    lookupCachedValue _ _ = return Nothing
+    getCacheSize _ = return 0
+
+-- |Construct a new 'DummyCache'. The size parameter is ignored.
+newDummyCache :: Int -> IO (DummyCache v)
+newDummyCache _ = pure DummyCache
```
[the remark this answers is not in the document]
td202: Renaming to NullCache.

concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
```diff
@@ -0,0 +1,328 @@
...
+      MHashableTo m h a
+    ) =>
+    MHashableTo m h (LazilyHashedCachedRef' h c a)
+    where
+    getHashM LazilyHashedCachedRef{..} =
+        liftIO (readIORef lhHash) >>= \case
+            Some h -> return h
+            Null -> do
+                h <- getHashM lhCachedRef
+                liftIO $ writeIORef lhHash (Some h)
```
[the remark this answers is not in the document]
td202: Apparently, it is not. I have updated the occurrences.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
td202: I think this will not be helpful in this case, so I've reverted.

td202 (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/Cache.hs
```diff
@@ -0,0 +1,205 @@
...
+    -- |Return the number of entries that can be stored in the cache.
+    getCacheSize :: (MonadCache c m) => Proxy c -> m Int
+
+-- | A context that simply wraps a cache, providing an instance @HasCache c (CacheContext c)@.
+newtype CacheContext c = CacheContext {theCacheContext :: c}
+
+instance HasCache c (CacheContext c) where
+    projectCache = theCacheContext
+
+-- |A dummy cache that does not store any values.
```
abizjak, line 94: Did this `dummy` remain on purpose?

concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
```diff
@@ -0,0 +1,462 @@
...
+    MHashableTo m h (CachedRef c a)
+    where
+    getHashM ref = getHashM =<< refLoad ref
+
+instance Show a => Show (CachedRef c a) where
+    show _ = "<CachedRef>"
+
+-- |We do nothing to cache a 'CachedRef'. Since 'cache' is generally used to cache the entire
+-- global state, it is generally undesirable to load every 'CachedRef' into the cache, as this
+-- can result in evictions and wasted effort if the chache size is insufficient.
```
abizjak, line 131:
```suggestion
-- can result in evictions and wasted effort if the cache size is insufficient.
```

abizjak (approved):

PR #434, opened 2022-07-12
Better trie

td202:
## Purpose

Improve performance and memory usage in the Trie implementation by using more efficient datastructures and greater strictness (to avoid unevaluated thunks).

Note, the on-disk representation of the Trie is unchanged. While the disk representation could be improved, this would require breaking changes or careful consideration.

## Changes

- Branching in the Trie is represented by boxed arrays of pairs of index and value.
- Stems in the Trie are represented by `ShortByteString`s, which are much more compact and have better strictness than the previous `[Word8]`.
- Strictness annotations are added in various places to guard against unevaluated thunks causing space leaks in the Trie.
- Where appropriate, `{-# UNPACK #-}` annotations are added to further reduce the size of datastructures in memory.
- The representation of the trie branches has been factored out of the trie structure itself to make changes easier, and unit tests have been added.
- A benchmark for the trie has been added.
- The `BufferedBlobbed` fixed-point combinator (and the associated `CachedBlobbed` and `Blobbed` combinators) has been replaced with a simpler `BufferedFix`, which uses `BufferedRef` directly.
- The `BufferedRef` includes a new `BRBoth` state that is similar to `BRMemory` except the disk reference cannot be null and is stored directly and not via an `IORef`. This change is to make the `BufferedFix` behave more similarly to the `BufferedBlobbed`, and it likely improves performance in other areas where `BufferedRef`s are used.

## Results

Memory usage for running a mainnet node (from a recent database) is reduced from around 1.7 GB typical to 0.9 GB. (Observed on Windows.)  Running `execute-chain` with a 180-block chain with 100,000 genesis accounts and many transactions showed reduced memory usage from 2.8 GB to 1.4 GB.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -710,8 +710,8 @@ instance MonadBlobStore m => BlobStorable m (CachedBlobbed BlobRef f)
 --
 -- It can contain either a CachedBlobbed value or both a Blobbed value and the recursive type.
 data BufferedBlobbed ref f
-    = LBMemory (IORef (Blobbed ref f)) (f (BufferedBlobbed ref f))
-    | LBCached (CachedBlobbed ref f)
+    = LBMemory {-# UNPACK #-} !(IORef (Blobbed ref f)) !(f (BufferedBlobbed ref f))
+    | LBCached !(CachedBlobbed ref f)
```
abizjak, line 714: I would probably prefer if we get rid of the intermediate "CachedBlobbed" since it both makes it harder to understand, and does not seem to be useful on its own.

It would also make profiling output much more useful, since using it would remove one of the silly fmaps during trie lookup/traversal that is conceptually just there to make the types work.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -57,14 +62,16 @@ class FixedTrieKey a where
 
 instance FixedTrieKey Word64
 instance FixedTrieKey Word32
-instance FixedTrieKey SHA256.Hash
+instance FixedTrieKey SHA256.Hash where
+    unpackKey (SHA256.Hash fbs) = FBS.unpack fbs
+    packKey = SHA256.Hash . FBS.pack
 deriving via SHA256.Hash instance FixedTrieKey ModuleRef
 instance FixedTrieKey AccountAddress
```
abizjak, line 69: This should also have a direct `FBS.unpack` / `FBS.pack` as in instance to remove the intermediate ByteString.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -76,16 +83,86 @@ instance OrdFixedTrieKey Word32
 deriving via Word64 instance OrdFixedTrieKey BakerId
 deriving via Word64 instance OrdFixedTrieKey DelegatorId
 
+-- |A pair of an index and a branch.
+data BranchEntry r = BranchEntry
+    { -- |The branch index.
```
abizjak, line 88: What is a branch index? Index into what?

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -76,16 +83,86 @@ instance OrdFixedTrieKey Word32
...
+        LT -> BranchEntry i v : l
+        EQ -> BranchEntry i v : r
+        GT -> p : updl r
+updateBranch i Null = Branches . Array.fromList . updl . Foldable.toList . theBranches
+  where
+    updl [] = []
+    updl l@(p@(BranchEntry k _) : r) = case compare i k of
+        LT -> l
+        EQ -> r
+        GT -> p : updl r
```
abizjak, line 140: It seems less than ideal to me to use intermediate lists. 

I don't really see how much is optimized by the compiler, and it might be fine, but it seems to me that it should be possible to do it better by doing binary search and updating the specific value if just the update is needed.

But it might not matter if most branches are small, then linear search is probably better anyhow.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -76,16 +83,86 @@ instance OrdFixedTrieKey Word32
...
+-- |Construct a 'Branches' with the given entries; all other entries are 'Null'.
+-- This is strict in the provided values.
+makeBranches :: [(Word8, Nullable r)] -> Branches r
+makeBranches = Branches . Array.fromList . collapse . sortOn fst
+  where
+    collapse [] = []
+    collapse [(i, Some v)] = [BranchEntry i v]
+    collapse [(_, Null)] = []
+    collapse ((i1, v1) : t@((i2, _) : _))
+        | i1 == i2 = collapse t
```
abizjak, line 151: Why is this special case needed? Why can't we make the assumption that keys are distinct?

Is there also an assumption that the list is ordered on the first component?

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -111,60 +188,60 @@ instance Bifunctor (TrieF k) where
...
                     return (fromIntegral (v - 3))
-            Stem <$> replicateM len getWord8 <*> get
+            Stem <$> getShortByteString len <*> get
 
 instance (BlobStorable m r, BlobStorable m (Nullable r), BlobStorable m v) => BlobStorable m (TrieF k v r) where
     storeUpdate (Branch vec) = do
-        pvec <- mapM storeUpdate vec
-        return $!! (putWord8 1 >> sequence_ (fst <$> pvec), Branch (snd <$> pvec))
+        pvec <- mapM storeUpdate (branchesToList vec)
+        return $!! (putWord8 1 >> sequence_ (fst <$> pvec), Branch (branchesFromList $ snd <$> pvec))
```
abizjak, line 216: What is the rationale for keeping full branches when store then on disk?

If most branches are not near full then it would save space to store a list of (k, v) pairs. And we'd only need one byte for the length.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -193,16 +270,18 @@ instance Functor (Trie k) where
...
 -- |Retrieve the value (if any) corresponding to a given key.
+-- TODO: With bytestring-0.11.3.0 and later, the operations on 'SBS.ShortByteString' can
+-- be used to implement this more efficiently without unpacking.
 lookupF :: (MRecursive m t, Base t ~ TrieF k v, FixedTrieKey k) => k -> t -> m (Maybe v)
 lookupF k = lu (unpackKey k) <=< mproject
     where
         lu [] (Tip v) = pure (Just v)
         lu _ (Tip _) = pure Nothing
-        lu key (Stem pref r) = case stripPrefix pref key of
+        lu key (Stem pref r) = case stripPrefix (SBS.unpack pref) key of
```
abizjak, line 280: We could implement a probably more efficient version now by modifying followStem to take a bytestring and a cursor.
It's not much more complicated than comparing two lists.

concordium-consensus/test-runners/execute-chain/Main.hs
```diff
@@ -43,11 +42,11 @@ main = do
     createDirectoryIfMissing True dataDir
     let config ::
             MultiVersionConfiguration
-                (PairGSConfig DiskTreeDiskBlockConfig MemoryTreeMemoryBlockConfig)
+                DiskTreeDiskBlockConfig
                 (NoFinalization ThreadTimer)
         config =
             MultiVersionConfiguration
-                { mvcStateConfig = (DiskStateConfig dataDir, ()),
+                { mvcStateConfig = DiskStateConfig dataDir,
```
abizjak, line 49: This does not seem like it belongs to this PR.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
abizjak: And instead just have the BufferedBlobbed with 3 constructors, which would remove an indirection as well.

If this indirection is somehow important, then it should be clearly documented.

abizjak:
Also, please document improvements (with numbers, e.g., mainnet node uses ?? before and after) as a result of this PR in its description.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -76,16 +83,86 @@ instance OrdFixedTrieKey Word32
...
+    deriving (Show, Functor, Foldable, Traversable)
+
+-- |Data structure representing the branches of the Trie.
+newtype Branches r = Branches
+    { -- |The branches, represented as an ordered array of pairs of indices and non-null values.
+      theBranches :: Array.Array (BranchEntry r)
+    }
+    deriving (Show, Functor, Foldable, Traversable)
+
+-- |Convert 'Branches' to a list. The list will have length 256.
```
mkmks, line 102: I'd put an explanation here that 256 is the alphabet set cardinality, and not some random number.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -473,9 +473,9 @@ class Monad m => Reference m ref a where
 -- ('BRMemory' with @brIORef = null@), or both in memory and on disk. When the
 -- value is both on disk and in memory the two values must match.
 data BufferedRef a
-    = BRBlobbed {brRef :: !(BlobRef a)}
+    = BRBlobbed {brRef :: {-# UNPACK #-} !(BlobRef a)}
```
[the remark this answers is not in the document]
td202: Probably. I have removed these.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -473,9 +473,9 @@ class Monad m => Reference m ref a where
...
     -- ^Value stored on disk
-    | BRMemory {brIORef :: !(IORef (BlobRef a)), brValue :: !a}
+    | BRMemory {brIORef :: {-# UNPACK #-} !(IORef (BlobRef a)), brValue :: !a}
```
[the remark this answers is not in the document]
td202: I did an experiment, and it seems not. Removed.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
td202: I have removed the `Blobbed`/`CachedBlobbed`/`BufferedBlobbed` types and replaced them with a simpler `BufferedFix`. This seems to be much more performant in the lookup benchmarks.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
td202: Done.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
td202: I have replaced this with a simpler function that is better suited to how this is used. There is no ordering assumption, but we can assume the keys are distinct - and in fact there are only two keys.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
td202: I looked into this. I found that it could save less than 100ns per look-up. That's not completely insignificant in the lookup benchmarks, but in the broader scheme of things it probably isn't significant. It was also not so trivial to implement, at least in my approach, and I think the added complexity of the change doesn't justify it.

concordium-consensus/benchmarks/trie/TrieBench.hs
```diff
@@ -0,0 +1,102 @@
...
+cleanupPersistent :: (BlobStore, AM.PersistentAccountMap 'P4) -> IO ()
+cleanupPersistent = destroyBlobStore . fst
+
+main :: IO ()
+main =
+    defaultMain
+        [ env (pure testAccountMap) $ \am0 ->
+            bgroup
+                "lookup"
+                [ env (pure (testAccountAddress 0)) $ \addr0 ->
```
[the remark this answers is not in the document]
td202: Done.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -630,185 +660,51 @@ instance (BlobStorable m a, BlobStorable m b, MHashableTo m H.Hash a) => BlobSto
...
-
--- | A type that is an instance of @HasNull@ has a distinguished value that is considered a Null value.
-class HasNull ref where
-    refNull :: ref
-    isNull :: ref -> Bool
-
-instance HasNull (BlobRef a) where
-    refNull = BlobRef maxBound
-    isNull = (== refNull)
+-- |'BufferedFix' is a fixed-point combinator that uses a 'BufferedRef'.
```
abizjak, line 663: This comment is as cryptic as it used to be. It would be good to expand on the intended uses instead of the generic comment.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -352,6 +352,14 @@ deriving via (LiftMonadBlobStore (StateT s) m)
 deriving via (LiftMonadBlobStore (ExceptT e) m)
     instance MonadBlobStore m => MonadBlobStore (ExceptT e m)
 
+-- |This instance lifts a 'MonadBlobStore' over a 'ReaderT' transformer.
+-- It is overlapped by the instance for @ReaderT BlobStore IO@ that uses
+-- the 'BlobStore' to provide the instance instead. This overlap should
+-- not be problematic, since there should be no 'MonadBlobStore' instance
+-- for 'IO'.
+deriving via (LiftMonadBlobStore (ReaderT r) m)
+    instance {-# OVERLAPPABLE #-} MonadBlobStore m => MonadBlobStore (ReaderT r m)
```
abizjak, line 361: Why is this actually needed? For the caching of instances? Why could it not have just stayed the way it was?

I think this is a bad instance due to the overlapping. It requires yet more understanding of yet more obscure Haskell stuff.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -111,63 +186,67 @@ instance Bifunctor (TrieF k) where
 instance (Serialize r, Serialize (Nullable r), Serialize v) => Serialize (TrieF k v r) where
     put (Branch vec) = do
         putWord8 1
-        forM_ vec put
+        forM_ (branchesToList vec) put
```
abizjak, line 189: Please provide documentation why we serialize the trie with branches in this representation, and not in the list of (k, v) pairs.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
abizjak: We could maintain that by making use of the fact that keys in the tries we use are never more than 96 bytes. So we could use tags, e.g. 200 for "branch in k, v representation".

We can do this separately, to test if it saves on disk space. But please provide a comment now as to why this disk layout is chosen for the trie.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
abizjak: Ok.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
td202: Yes, it is for caching of Instances. This instance allows us to avoid a complex lifting/unlifting of the `ReaderT` monad (as by `cacheBufferedBlobbed`.

I agree that overlapping instances are not ideal, but I think removing the instance for `ReaderT BlobStore IO` is a better solution.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
td202: Noted.

PR #439, opened 2022-08-02
Revise how the initial transaction table is generated

td202:
## Purpose

Revise how the transaction table is generated when initialising the Skov.

## Changes

- It is no longer necessary to call `activateSkovState`/`activateGlobalState` after `initialiseNewSkov`/`initialiseNewGlobalState` in order to use it for consensus (though it is after `initialiseExistingSkov`). This is because these functions now generate the transaction table based on the (re)genesis data. (And the only block is already cached.) This avoids potentially expensive additional traversals of the account table.
- The `genesisState` function returns an initial transaction table as well as the (in-memory) block state.
- Generating a transaction table from a block state is handled by a new `BlockStateQuery` function `getInitialTransactionTable`.
- The generated transaction tables omit entries for accounts with next nonce 1 (where they are not required).

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/concordium-consensus/src/Concordium/MultiVersion.hs
+++ b/concordium-consensus/src/Concordium/MultiVersion.hs
@@ -418,7 +418,9 @@ newGenesis (PVGenesisData (gd :: GenesisData pv)) vcGenesisHeight =
                 let newEConfig :: VersionedConfiguration gsconf finconf pv
                     newEConfig = VersionedConfiguration{..}
                 writeIORef mvVersions (oldVersions `Vec.snoc` newVersion newEConfig)
+                mvLog Runner LLTrace "Getting genesis configuration"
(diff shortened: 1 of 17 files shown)


concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -469,8 +471,8 @@ checkForProtocolUpdate = liftSkov body
                     oldTransactions <- terminateSkov
                     -- Transfer the non-finalized transactions to the new version.
                     lift $ do
-                        vvec <- liftIO . readIORef =<< asks mvVersions
-                        case Vec.last vvec of
+                        lastV <- fmap Vec.last . liftIO . readIORef =<< asks mvVersions
```
abizjak, line 474: It would probably be better to do the `fmap` inside IO instead.

abizjak (approved):

EmilLa1 (approved):

PR #440, opened 2022-08-02
Added support for exporting the block DB in chunks of fixed size

### LINKED ISSUE #383: Export the database in chunks
Instead of writing all blocks and finalization records for all protocol versions to a single file, one could write them to several files with names suffixed by the epoch number. This way it would be possible to begin catch-up starting from a specified epoch, and not from epoch 0.

Another choice of catch-up file granularity is to split it by slots, instead of epochs.

Currently, exportBlocks exports all blocks starting from the given block height, and exportFinRecs exports all finalization records starting from the last finalization index observed during block export. There's no one-to-one correspondence between between block heights and finalization indices. When exporting blocks from a range of block heights, finalization records matching those blocks would not necessarily have indices in an uninterrupted range.

To make the exporter implementation simpler, blocks and finalization records might be exported in separate files, without the expectation to have a correspondence between a file with blocks and a file with finalization records.

The database exporter should produce an index file listing the exported chunks together with block ranges present in the exported chunks.

# Acceptance criteria

It must possible to be import the chunked files exported by `database-exporter`. Possible circular dependency on #384.

mkmks:
## Purpose

Closes #383. `database-exporter` can now export blocks and finalization records as a collection of files (chunks) which size is set with the new obligatory `--chunksize` option.

## Changes

The export operation is implemented as a nest of four recursive functions (resembling how the previous database exporter version was structured): `exportSections` calls `writeChunks` which calls `exportBlocksToChunk` which calls `exportFinRecsForChunk`.

`exportSections` is left mostly unchanged from the previous version.

`writeChunks` creates exported chunks on disk and their accompanying index file listing chunk names and exported block ranges for each chunk.

`exportBlocksToChunk` writes a range of blocks into a chunk, while accumulating a list of finalization indexes corresponding to finalization records not already contained in the exported blocks.

Note that unlike the older `exportBlocks` function it doesn't return the number of blocks it wrote. It returns the _difference_ between the chosen chunk size and the number of exported blocks. For all chunks but the last one in a section it returns zero. A non-zero value is only returned for the last chunk in a section.

`exportFinRecsForChunk` writes the finalization records matching the list of finalization indexes produced by `exportBlocksToChunk`. Unlike the older `exportFinRecs`function, it doesn't require having the last exported finalization index in the exporter state. The exporter state `DBState` only carries the database handlers now.

Each exported chunk is meant to contain only finalization records matching its blocks. 

## Checklist

      hard-to-understand areas.

## CLA acceptance


concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
+    outDir indexHdl chunkSize = do
+        let chunkName = outDir </> "blocks-"
+                        ++ show sectionGenesisIndex ++ "-"
+                        ++ show sectionFirstBlockHeight
+                        ++ ".dat"
+        chunkHdl <- liftIO $ initialHandle chunkName
+
         (sectionStart, blocksStart) <- liftIO $ do
-            sectionStart <- hTell hdl
+            BS.hPut chunkHdl $ encode (3 :: Version)
```
abizjak, line 214: What is this 3 for? That was the version for the previous export. Why are we keeping the version the same? While the format seems to be the same, sections now don't start with first non-genesis blocks, so there is a difference.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
-            blocksStart <- hTell hdl
+            runPutH (liftPut $ putWord64be sectionHeaderLength >> put placeholderSectionHeader) chunkHdl
+            blocksStart <- hTell chunkHdl
             return (sectionStart, blocksStart)
-        sectionBlockCount <- writeBlocks
-        blocksEnd <- liftIO $ hTell hdl
-        sectionFinalizationCount <- writeFinRecs
+        -- how many blocks would it take to fill the exported chunk to its `chunkSize`
+        -- (non-zero means we're out of blocks to export in the current section)
+        (chunkUnused, finRecsMissing) <- exportBlocksToChunk chunkHdl sectionFirstBlockHeight chunkSize
```
abizjak, line 222: It seems it would be cleaner to return the number of blocks that were exported instead of this chunkUnused.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
     BlockHeight ->
-    m Word64
-exportBlocks hdl = eb 0
-  where
-    eb count height =
-        resizeOnResized (readFinalizedBlockAtHeight height) >>= \case
-            Nothing -> return count
+    -- |Number of blocks to export
+    Word64 ->
+    -- |(Block counter final value, finalization records not included in blocks)
```
abizjak, line 268: What does "block counter final value" mean? Please be clear.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
-                    case blockFinalizationData bf of
-                        BlockFinalizationData fr ->
-                            dbsLastFinIndex .= finalizationIndex fr
-                        _ -> return ()
-                eb (count + 1) (height + 1)
-            Just _ -> return count -- this branch should never be reachable, genesis blocks are not exported.
+                let finRecIndex :: Maybe FinalizationIndex = join . forM (blockFields (sbBlock sb)) $ \bf ->
+                      case blockFinalizationData bf of
+                        BlockFinalizationData _ -> Nothing
+                        _ -> Just (sbFinalizationIndex sb)
```
abizjak, line 283:
```suggestion
                let finRecIndex :: Maybe FinalizationIndex = (blockFields (sbBlock sb)) >>= \bf ->
                      case blockFinalizationData bf of
                        BlockFinalizationData _ -> Nothing
                        NoFinalizationData -> Just (sbFinalizationIndex sb)
```

abizjak: Please document the intent here. It is not at all obvious why we do this here, so this is not maintainable.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
+                if count > 1
+                  then (_2 %~ (maybeToList finRecIndex ++)) <$> exportBlocksToChunk hdl (height + 1) (count - 1)
```
abizjak, line 285: It would be better to make this function tail recursive instead of patching the response like this. That's easy to do with an accumulator.

concordium-consensus/tools/database-exporter/Main.hs
```diff
@@ -47,7 +47,8 @@ main = do
     Identity conf <- execParser opts
     case conf of
         Check file -> checkDatabase file
-        Export db filepath -> exportDatabaseV3 db filepath
+        Export db filepath n | n > 0 -> exportDatabaseV3 db filepath n
+        Export _ _ _ -> putStrLn "Chunk size should be larger than 0"
```
abizjak, line 51: This strictly speaking breaks compatibility. Our scripts for export will no longer work, nor is this compatible with existing nodes directly.

We can keep using the existing exporter, or presumably use `--chunk-size=2^63` to achieve this effect (but the file names will be different) but regardless, this needs to be planned so that we maintain out of band catchup also for existing 4.2.* nodes that have not upgraded immediately.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
-                    exportSections hdl dbDir (genIndex + 1)
-        else putStrLn ("The tree state database does not exist at " ++ treeStateDir)
+    -- |Finalization records to export
+    [FinalizationIndex] ->
+    -- |Number of exported finalization records
+    m Word64
+exportFinRecsForChunk hdl finRecIndexes = do
+  result <- forM finRecIndexes $ \finRecIndex ->
+    resizeOnResized (readFinalizationRecord finRecIndex) >>= \case
+      Nothing -> return 0
```
abizjak, line 302: When can this happen? Document.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -114,11 +115,74 @@ placeholderSectionHeader = SectionHeader 0 0 P1 (BlockHash minBound) 0 0 0 0
...
+                                    chunkSize
+                        )
+                        (DBState dbh)
+                    closeDatabase dbh
+                    exportSections dbDir outDir indexHdl chunkSize (genIndex + 1)
+        else putStrLn ("The tree state database does not exist at " ++ treeStateDir)
+
+-- |Write a database section as a collection of chunks in the specified directory. The exported
+-- chunks will be accompanied by an index file mapping chunk filenames to the block ranges stored in
+-- them. Each line in index file has format FILENAME,FIRSTBLOCK,LASTBLOCK\n
```
abizjak, line 183: There is missing documentation as to what is in each chunk. That is, how the finalization records in the chunk compare to the blocks in it.

abizjak (changes requested):

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,160 +191,123 @@ writeSection ::
...
                           ..
                         }
-            runPutH (liftPut $ putWord64be sectionHeaderLength >> put sectionHeader) hdl
-            hSeek hdl AbsoluteSeek sectionEnd
+            runPutH (liftPut $ putWord64be sectionHeaderLength >> put sectionHeader) chunkHdl
+            hClose chunkHdl
+            TIO.hPutStrLn indexHdl . T.pack $
+              takeFileName chunkName
+              ++ "," ++ show sectionFirstBlockHeight
+              ++ "," ++ show (sectionFirstBlockHeight + BlockHeight sectionBlockCount - 1)
```
td202, line 241: The index should probably give some indication of the genesis index/indices for chunks, or otherwise absolute block height should be used.

td202: Also, I don't recommend using the `Show` instance for `BlockHeight`/`GenesisIndex` etc. outside of diagnostics.

td202 (changes requested):
It would be good if the database exporter would take note of an existing index file and only export the missing blocks. To this end, it would probably be helpful if the index file included a header that includes the genesis block hash. This could be used by the database exporter, as well as any clients, to ensure compatibility early on.

concordium-consensus/tools/database-exporter/Main.hs
mkmks: I brought it up once or twice that removing the existing possibility to export the complete database without specifying a chunk size might break compatibility, and I was collectively assured it won't be an issue.

concordium-consensus/src/Concordium/ImportExport.hs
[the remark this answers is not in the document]
mkmks: I'm not sure I understand. The `Show` instance for `BlockHeight` is used here because the argument of `show` has type `BlockHeight`.

Also, in https://github.com/Concordium/concordium-node/pull/440/#discussion_r938796601 you advise against using the `Show` instance of `BlockHeight`, contrarily to the current comment. If that advice holds, what would you suggest here to turn `BlockHeight`s into strings that can be `Read`?

concordium-consensus/tools/database-exporter/Main.hs
td202: As you say, we can continue to use the existing database exporter. However, I think we would expect to deprecate the old exports once the current version of the node supports the new out-of-band catch-up. If someone needs to do out-of-band catch-up, it is because they are starting a node, in which case they should be starting the latest version.

concordium-consensus/src/Concordium/ImportExport.hs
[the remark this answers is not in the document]
td202: Sorry, this is a typo. I meant we should not rely on the `Show` instance for `BlockHeight`. i.e. convert them to a Word64 or similar and `show` that.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,76 +226,92 @@ writeSection ::
...
                     SectionHeader
                         { sectionLength = fromInteger (sectionEnd - sectionStart),
                           sectionBlocksLength = fromInteger (blocksEnd - blocksStart),
                           ..
                         }
-            runPutH (liftPut $ putWord64be sectionHeaderLength >> put sectionHeader) hdl
-            hSeek hdl AbsoluteSeek sectionEnd
+            runPutH (liftPut $ putWord64be sectionHeaderLength >> put sectionHeader) chunkHdl
+            hClose chunkHdl
+            TIO.hPutStrLn indexHdl . T.pack $
```
td202, line 277: What is the benefit of using `TIO.hPutStrLn` here, rather than `System.IO.hPutStrLn`? I think since we're using `String`s already there's nothing gained.

td202 (approved):
I would still like to have a header with the genesis hash in the export database. Otherwise, this seems to work fine.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
```
abizjak, line 4: This is a bad idea. Please remove it. If there is an impossible case please add it explicitly and document why it is impossible, e.g., `error "Cannot happen because I say so."`.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -114,11 +118,106 @@ placeholderSectionHeader = SectionHeader 0 0 P1 (BlockHash minBound) 0 0 0 0
 sectionHeaderLength :: Word64
 sectionHeaderLength = fromIntegral $ BS.length $ encode placeholderSectionHeader
 
--- |Write a section to the file handle.  It must be possible to write and seek the handle.
--- The section is written at the current seek location of the handle, and afterwards the seek
--- location is at the end of the the written section.
-writeSection ::
-    (MonadIO m) =>
+-- |Open a file handle for writing a chunk.
```
abizjak, line 121: The documentation is not adequate for what the function does. In particular this deletes an existing file.

concordium-consensus/tools/database-exporter/Main.hs
abizjak: @mkmks I am not sure where you brought this up, I am not aware of it. In this case it is also very easy to be backwards compatible.

While we can discuss amongst ourselves here, this needs to be coordinated with how the tools are used. So I expect a note to be written about this and circulated to at least Felice so that they may plan around it.

And we cannot expect that everybody "should" be starting the latest node, but that might not be the reality. People might be starting older nodes, since they don't have to upgrade, and the upgrade might require extra work for them.

In any case, I expect an explicit written plan on how migration is meant to work. If that already exists somewhere point me to it.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -127,76 +226,92 @@ writeSection ::
...
-    hdl
-    writeBlocks
-    writeFinRecs = do
+    sectionLastBlockHeight
+    outDir indexHdl chunkSize = do
+        let chunkName = outDir </> "blocks-"
+                        ++ show sectionGenesisIndex ++ "-"
+                        ++ show sectionFirstBlockHeight
+                        ++ ".dat"
+        chunkHdl <- liftIO $ initialHandle chunkName
```
td202, line 249: I mentioned this before, but I'm concerned that replacing existing chunks with new ones could cause problems for web caching/CDN. We could avoid this by picking a new name if a chunk already exists.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -205,82 +320,38 @@ exportBlocks hdl = eb 0
...
+                  then ebtc (height + 1) (count + 1)
+                  else return count
             Just _ -> return count -- this branch should never be reachable, genesis blocks are not exported.
 
--- |Export a series of finalization records, starting from the successor of 'dbsLastFinIndex'
--- as recorded in the state.
-exportFinRecs :: (MonadIO m, MonadState (DBState pv) m, MonadCatch m) => Handle -> m Word64
-exportFinRecs hdl = exportFinRecsFrom 0 . (+ 1) =<< use dbsLastFinIndex
+-- |Export all finalization records with indices above `dbsLastFinIndex` to a chunk
+exportFinRecsForChunk ::
```
td202, line 334: Maybe make the naming scheme more consistent wrt. ToChunk or ForChunk.

concordium-consensus/src/Concordium/ImportExport.hs
mkmks: HLS likes to add compiler options without being asked to...

concordium-consensus/tools/database-exporter/README.md
```diff
@@ -0,0 +1,27 @@
+# database-exporter
```
abizjak, line 1: There is already a section in the README in `concordium-cosnensus/README.md` That should be removed in light of this, and just a link to this document tshould be added.

concordium-consensus/tools/database-exporter/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Migration from earlier versions
+
+A version of `database-exporter` shipped with `concordium-node` <=4.3.0 did not implement the
+`--chunksize` option in its `export` command. It exported the node database as a single file. If you
+would like still to export the node database as a single file, run `database-exporter` with a very
+large `--chunksize` argument, for example, 1000000000000.
+
+After upgrading to `database-exporter` shipped with `concordium-node` >4.3.0, use the `--chunksize`
+option to every invocation of `database-exporter export`, as it is non-optional. A good first choice
```
abizjak, line 24:
```suggestion
option to every invocation of `database-exporter export`, as it is required. A good first choice
```

concordium-consensus/tools/database-exporter/README.md
```diff
@@ -0,0 +1,27 @@
...
+of the `--chunksize` argument is 10000, which exports block files with sizes on the order of tens of
+MiB.
```
abizjak, line 26: This number is not entirely right and is highly context dependent.

If blocks were close to their mainnet/testnet limit, i.e., 3MB, then 10000 blocks would be 30GB.

abizjak (approved):
Please cleanup the README before merging.

concordium-consensus/tools/database-exporter/README.md
mkmks: It'll take a much broader CCD adoption before blocks on mainnet reach this limit. What would be the number you'd recommend the node operators to choose?

PR #441, opened 2022-08-03
Fix consensus tests with corrected transaction costs.

td202:
## Purpose

The cost of dummy transfer transactions was previously incorrect. This is fixed in https://github.com/Concordium/concordium-base/pull/184. The change causes some consensus tests to fail, which this rectifies.

## Changes

- When checking the invariant for the transaction table, the account (and chain-update) non-finalized transactions should record the same transaction verification result as present in the transaction table hash map.

## Checklist

      hard-to-understand areas.


abizjak (approved):

EmilLa1 (approved):
Ok. That's better.

PR #447, opened 2022-08-07
Download catchup file into a robustly constructed temporary file.

### LINKED ISSUE #442: Catch-up downloader does not correctly identify Windows paths
**Bug Description**

When given a Windows path, the node attempts to parse it as a URL and download the catch-up file. This prevents using a file for catch-up.

**Steps to Reproduce**

Run the node with `--import-blocks-from c:\Users\Thomas\Downloads\mainnet-2022-08-03.db` (or a similar Windows path to a catch-up file).

**Expected Result**

The node should catch-up using the supplied file

**Actual Result**

The node fails to do so:
```
INFO concordium_node - Downloading the catch-up file from c:\Users\Thomas\Downloads\mainnet-2022-08-03.db to C:\ProgramData\Concordium\Node Runner\mainnet-cache\data\blocks_to_import_1659588325.dat
ERROR concordium_node - Downloading catch-up files failed: builder error for url (c:\Users\Thomas\Downloads\mainnet-2022-08-03.db): URL scheme is not allowed
```

**Versions**
- Software Version: Unreleased
- OS: Windows

abizjak:
## Purpose

Makes the downloading and storing of the file more robust.
Fixes #442 

## Changes

- Introduce a new configuration option/environment variable to explicitly request downloading of the block file.
- Minor cleanup. In particular the type of the function to import blocks did not match on Haskell and Rust sides.

## Checklist

      hard-to-understand areas.


mkmks (approved):
All fine with me.

concordium-node/src/bin/cli.rs
```diff
@@ -142,21 +145,42 @@ async fn main() -> anyhow::Result<()> {
...
+                        info!("Out of band catchup stopped.");
+                    } else {
+                        info!("Could not complete out of band catch-up due to: {:#}.", e);
+                    }
+                } else {
+                    info!("Completed out of band catch-up from {}.", download_url);
+                }
+                // attempt to properly clean up the downloaded file.
+                if let Err(e) = path.close() {
+                    error!("Could not delete the downloaded file: {}", e);
```
td202, line 178: I think this should be factored out to avoid having an overly long `main`. Possibly this should also move to consensus.rs.

td202 (approved):

concordium-node/Cargo.toml
```diff
@@ -80,8 +80,10 @@ mime = { version = "0.3", optional = true }
 serde_derive = { version = "1.0", optional = true }
 rmp-serde = { version = "0.15", optional = true }
 serde_json = { version = "1" }
-prometheus = { version = "0.13", default-features = false, features = ["gen", "push", "protobuf"], optional = true }
-tempfile = { version = "3.1", optional = true }
+num_cpus = { version = "1.13", optional = true }
+jsonwebtoken = { version = "7", optional = true }
+prometheus = { version = "0.12", default-features = false, features = ["gen", "push", "protobuf"], optional = true }
+tempfile = { version = "3.1" }
```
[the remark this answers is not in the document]
abizjak: This is here by accident. Will remove.

concordium-node/src/bin/cli.rs
abizjak: I do agree about factoring out. Not about consensus. The logic here is about getting the file as it is now. The actual import is already in consensus.

I think this should generally be done differently anyhow, and we should not be downloading and storing the file. We should just stream it if given a URL. But I don't have time to do that now.

Issue #120, opened 2021-08-04
Quicker verification of Network packets


EmilLa1:
Today most verification is only being carried out by the Scheduler i.e., after the transaction has already been added to the transactions table and it's about to get executed. 

In order to prevent congestion on the network we should carry out more verification of the incoming transactions even before adding the transactions to the transactions table. The verification results can - to some extent - be cached and reused by the scheduler and thus lowering the computational cost when executing transactions. 

Furthermore it could alleviate the current (lack of) rpc status' for send_transaction and thus help integrators with more detailed error messages. 

Note. Extensive testing has to be carried out as much of the transaction flow is rather subtle.

**Sub-tasks**
- [x] Implement an earlier verification process for the incoming transactions. As such a heuristic has to be created for when and when not to reject incoming transactions. 
- [x] Create a document which describes the transaction acceptance heuristic.

PR #449, opened 2022-08-08
Remove unused runtime parameter from the node.

abizjak:
## Purpose

The max expiry time used to be an option we had to limit damage from bad transactions. It is no longer used since we have introduced more transaction validation in #120 , but the command-line option remained.

## Changes

Remove unused option.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unrelease changes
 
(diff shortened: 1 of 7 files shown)


EmilLa1 (approved):
Nice catch - I totally forgot about this one.

PR #459, opened 2022-08-17
Modules caching

EmilLa1:
## Purpose

Addresses https://github.com/concordium/concordium-node/issues/461

The change revises the handling of smart contract modules in the state to reduce the resident memory used by them. Module artifacts (the pre-processed code of the modules) are no longer retained in memory at all times. Instead, they are loaded from the blob store as needed when executing smart contracts. Moreover, we introduce a cache for (the metadata of) smart contract modules. This serves to limit the number of contracts for which the metadata (including names of contracts init and receive methods) are retained in memory. This is a FIFO cache of default size 100.

## Changes

- Introduced a FIFO cache for the persistent modules. 
- Modules are now referenced via `HashedCachedReferences` in the `Modules` and `Instances` tables.

- Introduce a `BlobPtr` type for direct memory access in the blob store. A `BlobPtr` consists of an offset into the blob store together with a length. (This differs from a `BlobRef`, which does not directly incorporate the length of the data, but expects the length to be encoded in the blob store at the given offset.)  `BlobPtr`s can be accessed via the `withBlobPtr` and `loadBlobPtr` functions of `MonadBlobStore`.

- Introduce a new `DirectBlobStorable` class that gives more direct control over how a `BlobRef` is stored and loaded in the blob store. There is a blanket (overlappable) implementation of `DirectBlobStorable` based on `BlobStorable`. `DirectBlobStorable` is now used as a constraint in most places where previously `BlobStorable` was used (i.e. for `Reference` types in particular).

- Remove `store` from the signature of `BlobStorable`. This was already seldom used, but created a pitfall for implementations, as `storeUpdate` should typically be used instead.

- Revise `BufferedRef` so that `BRMem` has an `IORef` that will contain a `BRBoth` if the reference has already been written to the store.  This supports better sharing of references.  In particular, the change to the behaviour of `storeUpdate` on a `BRMem` where a copy of the reference has already been written will be to use the prior written (and hence updated) copy of the value. Not doing so caused an issue for the handling of modules, where references were kept from both the module table and from instances derived from the module: if both were written in a single pass, one reference would retain the in-memory
version of the module on the `storeUpdate`.

- Remove coercions on `BufferedRef`. This affects how smart contract modules are referenced from instances: `pinstanceModuleInterface` is a reference to a `Module` instead of a `ModuleV v`, thus losing the type-level version information around the module. The coercion is now handled explicitly when the contract instance is retrieved. The benefit of this is that module references can be shared between the module table (which were `Module` references) and contract instances.

- In the persistent state, introduce a `PersistentInstrumentedModuleV` type that stores the module in memory (before finalization) or on disk (with a `BlobPtr`). This is then used in the persistent `Module`, which implements `DirectBlobStorable` so that it can work with the `BlobPtr`.

- As module artifacts in the persistent state use `BlobPtr`s internally, modify the block state interface by introducing an abstract `InstrumentedModuleRef` type, and a class `ModuleQuery` with the operation `getModuleArtifact` that returns a Rust-side artifact that can be used by contract execution.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -378,14 +378,14 @@ data HashedPersistentBlockState pv = HashedPersistentBlockState {
 
 -- |Convert a 'PersistentBlockState' to a 'HashedPersistentBlockState' by computing
 -- the state hash.
-hashBlockState :: (SupportsPersistentAccount pv m) => PersistentBlockState pv -> m (HashedPersistentBlockState pv)
+hashBlockState :: (SupportsPersistentAccount pv m, Modules.SupportsPersistentModules m) => PersistentBlockState pv -> m (HashedPersistentBlockState pv)
```
td202, line 381: I recommend defining
```
type SupportsPersistentState pv m = (SupportsPersistentAccount pv m, Modules.SupportsPersistentModules m)
```
or perhaps
```
type SupportsPersistentState pv m =
    ( IsProtocolVersion pv,
      MonadBlobStore m,
      MonadCache (AccountCache (AccountVersionFor pv)) m,
      MonadCache ModuleCache m
    )
```

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
```diff
@@ -139,20 +160,33 @@ putModuleV0 (ModuleV1 ModuleV{..}) = sPut =<< loadRef moduleVSource
 
 --------------------------------------------------------------------------------
 
+-- |The cache retaining 'Module's
+type ModuleCache = FIFOCache Module
+
+-- |Construct a new `ModulesCache` with the given size.
```
td202, line 166:
```suggestion
-- |Construct a new 'ModuleCache' with the given size.
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Instances.hs
```diff
@@ -93,10 +94,11 @@ instance Applicative m => Cacheable m PersistentInstanceParameters
 data PersistentInstanceV (v :: Wasm.WasmVersion) = PersistentInstanceV {
     -- |The fixed parameters of the instance.
     pinstanceParameters :: !(BufferedRef PersistentInstanceParameters),
-    -- |The interface of 'pinstanceContractModule'. Note this is a BufferedRef to a Module as this
+    -- |The interface of 'pinstanceContractModule'. Note this is a 'HashedCachedRef' to a Module as this
     -- is how the data is stored in the Modules table. A 'Module' carries a BlobRef to the source
     -- but that reference should never be consulted in the scope of Instance operations.
-    pinstanceModuleInterface :: !(BufferedRef (Modules.ModuleV v)),
+    -- The module will always be of version @v@.
+    pinstanceModuleInterface :: !(HashedCachedRef Modules.ModuleCache Modules.Module),
```
td202, line 101: Wherever a `PersistentInstanceV` is constructed, I think it would be useful to have a brief argument that establishes why the module version will be `v`. (This would be the same as why the old `unsafeGetModuleReferenceV?` would be safe in the context.)

concordium-node/src/configuration.rs
```diff
@@ -287,6 +287,13 @@ pub struct BakerConfig {
         env = "CONCORDIUM_NODE_BAKER_DECRYPT_CREDENTIALS"
     )]
     pub decrypt_baker_credentials: bool,
+    #[structopt(
+        long = "modules-cache-size",
+        help = "The maximum number of artifacts that can be stored in the module cache",
+        default_value = "100",
```
td202, line 293: Is this actually a good default?

concordium-node/src/configuration.rs
```diff
@@ -287,6 +287,13 @@ pub struct BakerConfig {
         env = "CONCORDIUM_NODE_BAKER_DECRYPT_CREDENTIALS"
     )]
     pub decrypt_baker_credentials: bool,
+    #[structopt(
+        long = "modules-cache-size",
+        help = "The maximum number of artifacts that can be stored in the module cache",
```
td202, line 292: I think this could be more useful. Maybe `smart contract modules` instead of `artifacts`. I am not sure that users will know what `artifact` means.

CHANGELOG.md
```diff
@@ -25,6 +25,11 @@
 - Remove transaction logging functionality from the node. It is replaced by an
   external service. As a consequence the `transaction-outcome-logging` family of
   command-line options are removed from the node.
+- Smart contract modules are no longer being constantly retained in memory. Instead a 
+  limited number of smart contract modules are being retained in a cache. 
+  By default the node will retain at most 100 smart contract instances in memory, however this
+  is configurable via the `--modules-cache-size` command line argument or by using the 
+  `CONCORDIUM_NODE_CONSENSUS_MODULES_CACHE_SIZE` environment variable.
```
td202, line 32: Put this in `Unreleased changes`.

td202 (approved):

CHANGELOG.md
EmilLa1: Oops

concordium-node/src/configuration.rs
EmilLa1: There are currently 28 on `mainnet` so it is certainly sufficient for the current state. However we should revise this in the future with more modules being deployed to the chain.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
EmilLa1: Yes. I was a bit unsure if it was worth the hazzle of introducing a new abstraction for this pair of constraints. 

However thinking twice I actually lean towards your suggestion of adding it, so I will introduce it.

concordium-consensus/src/Concordium/GlobalState/Persistent/Instances.hs
EmilLa1: @td202 I've added some more documentation. Can you please check that it is in line with what you would've expected?

abizjak (commented):
It seems fine to me, but we are still relying on Haskell finalizers in a place where we should not in my view.

But I realize it might be more to change now so we need to see if it should go into this PR or not. 

Nevertheless, I would like to see updated timings based on this branch as discussed with @EmilLa1 .

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
```diff
@@ -139,20 +242,42 @@ putModuleV0 (ModuleV1 ModuleV{..}) = sPut =<< loadRef moduleVSource
 
 --------------------------------------------------------------------------------
 
+-- |A reference for obtaining a module in the 'LFMBTree'.
+-- The module is cached in the 'ModuleCache' while the actual artifact is
+-- loaded on demand.
+type ModuleReference = HashedCachedRef ModuleCache
```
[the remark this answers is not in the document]
EmilLa1: Renamed and put some more documentation which should make it clear we're not talking about 'ModuleRef' s.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -279,6 +278,41 @@ writeBlobBS BlobStoreAccess{..} bs = mask $ \restore -> do
     where
         size = encode (fromIntegral (BS.length bs) :: Word64)
 
+-- |A pointer into the blob store that can be used for direct memory access.
```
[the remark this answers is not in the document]
EmilLa1: Tried to elaborate a bit on this.

concordium-consensus/src/Concordium/GlobalState/Wasm.hs
```diff
@@ -135,10 +152,21 @@ instance Serialize (InstrumentedModuleV V1) where
...
+      SV0 -> return (InstrumentedWasmModuleV0 artifact)
+      SV1 -> return (InstrumentedWasmModuleV1 artifact)
+
 --------------------------------------------------------------------------------
 
--- |A Wasm module interface of a given version, specified via a type parameter.
-data ModuleInterfaceV v = ModuleInterface {
+-- |A Wasm module interface, parametrised by the type of the instrumented module.
+-- The instrumented module should be e.g. @InstrumentedModuleV v@.
+data ModuleInterfaceA instrumentedModule = ModuleInterface {
```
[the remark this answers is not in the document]
EmilLa1: Artifact

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -518,73 +622,59 @@ class Monad m => Reference m ref a where
...
 data BufferedRef a
     = BRBlobbed {brRef :: !(BlobRef a)}
     -- ^Value stored on disk
-    | BRMemory {brIORef :: !(IORef (BlobRef a)), brValue :: !a}
+    | BRMemory {brIORef :: !(IORef (Nullable (BufferedRef a))), brValue :: !a}
     -- ^Value stored in memory and possibly on disk.
-    -- When a new 'BRMemory' instance is created, we initialize 'brIORef' to 'refNull'.
-    -- When we store the instance in persistent storage, we update 'brIORef' with the corresponding pointer.
+    -- When a new 'BRMemory' instance is created, we initialize 'brIORef' to 'Null'.
+    -- When we store the instance in persistent storage, we update 'brIORef' to a @Some BRBoth{..}@.
```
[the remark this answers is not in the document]
EmilLa1: Added some more documentation for now.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
```diff
@@ -24,32 +33,62 @@ import Concordium.Crypto.SHA256
...
+
+-- |Load a 'PersistentInstrumentedModuleV', retrieving the artifact.
+-- If the artifact has been persisted to the blob store, this creates an (executable) copy of the
+-- artifact on the Rust side.  The Rust artifact is reference counted, and Haskell retains a
+-- reference with a 'ForeignPtr' that drops the reference when it is garbage collected.
+loadInstrumentedModuleV :: (MonadBlobStore m, IsWasmVersion v) => PersistentInstrumentedModuleV v -> m (GSWasm.InstrumentedModuleV v)
+loadInstrumentedModuleV (PIMVMem im) = return im
+loadInstrumentedModuleV (PIMVPtr ptr) = do
+  bs <- loadBlobPtr ptr
+  liftIO $ GSWasm.instrumentedModuleVFromBytes bs
```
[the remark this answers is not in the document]
EmilLa1: I'll see what I can manage to get ready before code freeze, but I'm unsure if there is time now.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -600,42 +702,41 @@ instance BlobStorable m a => BlobStorable m (Nullable (BufferedRef a)) where
...
-    if isNull r
-      then do
-        (!r' :: BlobRef a, !v') <- storeUpdateRef v
-        liftIO . writeIORef ref $! r'
+  refFlush (BRMemory ref v) = liftIO (readIORef ref) >>= \case
+    Null -> do
+        (!r' :: BlobRef a, !v') <- storeUpdateDirect v
+        let br' = BRBoth r' v'
+        liftIO . writeIORef ref $! Some br'
         return (BRBoth r' v', r')
```
abizjak, line 738: You have two copies of `br'` there.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -521,73 +636,60 @@ class Monad m => Reference m ref a where
...
-    -- When a new 'BRMemory' instance is created, we initialize 'brIORef' to 'refNull'.
-    -- When we store the instance in persistent storage, we update 'brIORef' with the corresponding pointer.
+    -- When a new 'BRMemory' instance is created, we initialize 'brIORef' to 'Null'.
+    -- When we store the instance in persistent storage, we update 'brIORef' to a @Some BRBoth{..}@.
     -- That way, when we store the same instance again on disk (this could be, e.g., a child block
     -- that inherited its parent's state) we can store the pointer to the 'brValue' data rather than
     -- storing all of the data again.
+    -- This definition makes it possible for shared references to share the
+    -- updated value after it's been 'storeUpdated'.
+    -- See implementation for 'storeUpdate'.
```
abizjak, line 653: This comment does not tell me (maybe it should, but I am not getting it) why, for example, I would not use 
```
    | BRMemory {brIORef :: !(IORef (Nullable (!a, !BlobRef a))), brValue :: !a}
```
seeing that the only variant that is ever put in is BRBoth (I realize the pair above should be strict and you can't put bangs on nested values, but you get my point). What would be wrong with this?

But also the fact that now BRMemory is not really that anymore is confusing. It sounds like there is some insight here that was gained through a lot of effort, but is not documented.

Please describe a specific example where sharing this IORef between multiple BRMemory values provides an improvement. Describe it in such a way that one can actually observe it in the future when any changes to this are needed.

abizjak (commented):
Looks good to me now, apart from that issue with BufferedRef.

As it is now I don't think it can be understood, so please try to improve the description there so that we know why it is the way it is also  in 2 months.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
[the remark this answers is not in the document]
EmilLa1: This goes into a new ticket! 
https://github.com/Concordium/concordium-node/issues/498

PR #460, opened 2022-08-18
Added support for importing catchup chunk files

### LINKED ISSUE #437: Fetch and import multiple catchup files
In #381, the `--import-blocks-from` option gained support for fetching catchup files from remote URLs. This functionality should be expanded to support fetching and processing catchup _index files_.

Catchup data exported from a node database in multiple chunks is accompanied by an index file where each line has a format "FILENAME,FIRSTBLOCK\n". Supplying a URL pointing to an index file to `--import-blocks-from` should result in downloading the index file, then using it to download only the catchup chunks containing the latest blocks missing from the local database.

Downloaded chunks should be deleted after import. 

# Acceptance criteria

It should be possible to import the catchup data exported using the `database-exporter` implementation from #383.

mkmks:
## Purpose

Closes #437. `--download-blocks-from` now takes a URL to the index file that lists block chunks to download and import. Only chunks containing blocks with a height greater than the best block height are downloaded.

## Changes

The implementation is contained in the `import_missing_blocks` function which streams the index file, parses it and uses the extracted chunk records to decide whether to download and import chunks.

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/concordium-consensus/src/Concordium/ImportExport.hs
+++ b/concordium-consensus/src/Concordium/ImportExport.hs
@@ -164,6 +164,7 @@ exportDatabaseV3 dbDir outDir chunkSize = do
       return (lastChunkGenesisIndex, lastChunkFirstBlock)
     else do
       createDirectoryIfMissing True outDir
+      writeFile indexFile "filename,genesis_index,first_block_height,last_block_height\n"
       -- we export blocks starting from height 1 because genesis blocks need not be exported
       return (0, 1)
   bracket (openFile indexFile AppendMode) hClose
@@ -220,7 +221,8 @@ exportSections dbDir outDir indexHdl chunkSize genIndex startHeight = do
 -- chunk (i.e. the one containing the block with the greatest height in the section) also contains
 -- finalization records finalizing all blocks after the last block containing a finalization
 -- record. The exported chunks will be accompanied by an index file mapping chunk filenames to the
--- block ranges stored in them. Each line in index file has format FILENAME,GENESISINDEX,FIRSTBLOCK,LASTBLOCK\n
+-- block ranges stored in them. Each line in index file has format
+-- 'filename,genesis_index,first_block_height,last_block_height\n'.
 writeChunks ::
     (IsProtocolVersion pv, MonadIO m, MonadState (DBState pv) m, MonadCatch m) =>
     -- |Genesis index
--- a/concordium-consensus/src/Concordium/ImportExport.hs
+++ b/concordium-consensus/src/Concordium/ImportExport.hs
@@ -164,7 +164,6 @@ exportDatabaseV3 dbDir outDir chunkSize = do
       return (lastChunkGenesisIndex, lastChunkFirstBlock)
     else do
       createDirectoryIfMissing True outDir
-      writeFile indexFile "filename,genesis_index,first_block_height,last_block_height\n"
       -- we export blocks starting from height 1 because genesis blocks need not be exported
       return (0, 1)
   bracket (openFile indexFile AppendMode) hClose
--- a/concordium-consensus/src/Concordium/ImportExport.hs
+++ b/concordium-consensus/src/Concordium/ImportExport.hs
@@ -221,7 +221,7 @@ exportSections dbDir outDir indexHdl chunkSize genIndex startHeight = do
 -- finalization records finalizing all blocks after the last block containing a finalization
 -- record. The exported chunks will be accompanied by an index file mapping chunk filenames to the
 -- block ranges stored in them. Each line in index file has format
--- 'filename,genesis_index,first_block_height,last_block_height\n'.
+-- @filename,genesis_index,first_block_height,last_block_height\n@.
 writeChunks ::
     (IsProtocolVersion pv, MonadIO m, MonadState (DBState pv) m, MonadCatch m) =>
     -- |Genesis index
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -543,7 +543,8 @@ async fn import_missing_blocks(
         .strip_prefix("# genesis hash ")
         .context(
             "The catchup index file does not begin with a line containing the genesis block hash. \
-             Please contact the catchup service administrator.",
+             Please verify that you specified a correct catchup service URL. If the specified URL \
+             is correct, contact the catchup service administrator.",
         )?
         .trim();
     let genesis_hash = genesis_block_hashes[0].to_string();
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -490,9 +490,14 @@ async fn maybe_do_out_of_band_catchup(
     } else if let Some(download_url) = download_blocks_from.as_ref().cloned() {
         info!("Starting out of band catch-up");
         let genesis_block_hashes = regenesis_arc.blocks.read().unwrap().clone();
-        if let Err(e) =
-            import_missing_blocks(consensus, &genesis_block_hashes, download_url, data_dir_path)
-                .await
+        if let Err(e) = import_missing_blocks(
+            consensus,
+            import_stopped.clone(),
+            &genesis_block_hashes,
+            download_url,
+            data_dir_path,
+        )
+        .await
         {
             if import_stopped.load(atomic::Ordering::Acquire) {
                 info!("Out of band catchup stopped.");
@@ -507,7 +512,6 @@ async fn maybe_do_out_of_band_catchup(
 
 // An index entry for a chunk of blocks. Its format must correspond to one
 // produced by `database-exporter`.
-#[allow(dead_code)]
 #[derive(serde::Deserialize)]
 struct BlockChunkData {
     // exported chunk of blocks' filename
@@ -515,6 +519,7 @@ struct BlockChunkData {
     // genesis block index from which relative heights of blocks in the chunk are counted
     genesis_index:      usize,
     // relative height of the oldest block stored in the chunk
+    #[allow(dead_code)]
     first_block_height: u64,
     // relative height of the newest block stored in the chunk
     last_block_height:  u64,
@@ -522,6 +527,7 @@ struct BlockChunkData {
 
 async fn import_missing_blocks(
     consensus: &ConsensusContainer,
+    import_stopped: Arc<atomic::AtomicBool>,
     genesis_block_hashes: &[HashBytes],
     index_url: &url::Url,
     data_dir_path: &Path,
@@ -573,6 +579,10 @@ async fn import_missing_blocks(
         .create_deserializer(index_reader)
         .into_deserialize();
     while let Some(result) = chunk_records.next().await {
+        if import_stopped.load(atomic::Ordering::Acquire) {
+            break;
+        }
+
         let block_chunk_data: BlockChunkData = result?;
         // no need to reimport blocks that are present in the database
         if block_chunk_data.genesis_index < current_genesis_index
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -525,7 +525,7 @@ async fn import_missing_blocks(
     index_url: &url::Url,
     data_dir_path: &Path,
 ) -> anyhow::Result<()> {
-    let best_block_height = consensus.get_best_block_height();
+    let last_finalized_block_height = consensus.get_last_finalized_block_height();
     let current_genesis_index = genesis_block_hashes.len() - 1;
     let current_genesis_block_hash = genesis_block_hashes[current_genesis_index].to_string();
     let json_genesis_block_value: serde_json::Value =
@@ -534,7 +534,7 @@ async fn import_missing_blocks(
 
     info!("Current genesis index: {}", current_genesis_index);
     info!("Current genesis block height: {}", current_genesis_block_height);
-    info!("Local best block height: {}", best_block_height);
+    info!("Local last finalized block height: {}", last_finalized_block_height);
 
     let comments_stream = reqwest::get(index_url.clone())
         .await?
@@ -573,9 +573,9 @@ async fn import_missing_blocks(
         // no need to reimport blocks that are present in the database
         if block_chunk_data.genesis_index < current_genesis_index
             || current_genesis_block_height + block_chunk_data.last_block_height
-                <= best_block_height
+                <= last_finalized_block_height
         {
-	    info!("Skipping chunk {}: no blocks above best block height", block_chunk_data.filename);
+	    info!("Skipping chunk {}: no blocks above last finalized block height", block_chunk_data.filename);
             continue;
         }
         let block_chunk_url = index_url.join(&block_chunk_data.filename)?;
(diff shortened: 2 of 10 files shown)


concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +485,70 @@ async fn maybe_do_out_of_band_catchup(
...
+    index_url: &url::Url,
     data_dir_path: &Path,
-) -> anyhow::Result<TempPath> {
+) -> anyhow::Result<()> {
+    let json_consensus_value: serde_json::Value =
+        serde_json::from_str(&consensus.get_consensus_status())?;
+    let best_block_height = json_consensus_value["bestBlockHeight"].as_u64().unwrap();
+    let current_genesis_index = json_consensus_value["genesisIndex"].as_u64().unwrap();
+    let current_genesis_block_hash =
+        json_consensus_value["currentEraGenesisBlock"].as_str().unwrap();
```
td202, line 519: The genesis index and block hash are available from the `regenesis_arc`.

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +485,70 @@ async fn maybe_do_out_of_band_catchup(
...
+
+async fn import_missing_blocks(
+    consensus: &ConsensusContainer,
+    index_url: &url::Url,
     data_dir_path: &Path,
-) -> anyhow::Result<TempPath> {
+) -> anyhow::Result<()> {
+    let json_consensus_value: serde_json::Value =
+        serde_json::from_str(&consensus.get_consensus_status())?;
+    let best_block_height = json_consensus_value["bestBlockHeight"].as_u64().unwrap();
```
td202, line 516: We should add an ffi call to get the (current era) block height of the last finalized block.

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +485,70 @@ async fn maybe_do_out_of_band_catchup(
...
+    let index_reader = reqwest::get(index_url.clone())
+        .await?
+        .bytes_stream()
+        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
+        .into_async_read();
+    let mut chunk_records = csv_async::AsyncReaderBuilder::new()
+        .comment(Some(b'#'))
+        .create_deserializer(index_reader)
+        .into_deserialize();
+    while let Some(result) = chunk_records.next().await {
```
td202, line 532: There does not seem to be any check that the genesis hash in the file matches ours. We should log an appropriate error message in that case, so the user can correct the issue.

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +485,70 @@ async fn maybe_do_out_of_band_catchup(
...
 
-async fn download_catchup_files(
-    download_url: url::Url,
+#[derive(serde::Deserialize)]
+struct BlockChunkData {
+    filename:           String,
+    genesis_index:      u64,
+    first_block_height: u64,
+    last_block_height:  u64,
+}
```
abizjak, line 507: Type and field documentation is missing. In particular the purpose of the type should be documented, as well as what its dependencies are, i.e., when it should change.

concordium-node/src/bin/cli.rs
mkmks: Since the (re)genesis hashes are stored in the index files as comments, I had to stream the index file twice, the first time to read the comments, the second time to parse the CSV rows.

Perhaps, it's not ideal because I could separate the comments from CSV rows while reading the download stream only once, but code readability would suffer, in my opinion. Also, the first stream is, in fact, only used to read the very first genesis hash comment, so the whole index file won't be downloaded twice. Almost certainly it won't matter in practice, because the index file will be much smaller than even one block chunk.

An alternative to storing the genesis hashes as comments would be to make them fields, but that would unnecessarily blow up the index file size (adding 64 ASCII characters needed to represent a 256-bit hash to each row will more than double the row length).

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +488,104 @@ async fn maybe_do_out_of_band_catchup(
...
     data_dir_path: &Path,
-) -> anyhow::Result<TempPath> {
+) -> anyhow::Result<()> {
+    let best_block_height = consensus.get_best_block_height();
+    let genesis_block_hashes = regenesis_arc.blocks.read().unwrap();
+    let current_genesis_index = genesis_block_hashes.len() - 1;
+    let current_genesis_block_hash = genesis_block_hashes[current_genesis_index].to_string();
+    let json_genesis_block_value: serde_json::Value =
+        serde_json::from_str(&consensus.get_block_info(&current_genesis_block_hash)?)?;
+    let current_genesis_block_height = json_genesis_block_value["blockHeight"].as_u64().unwrap();
```
td202, line 533: This wouldn't be necessary if you get the relative block height rather than absolute block height.

td202 (changes requested):

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +488,105 @@ async fn maybe_do_out_of_band_catchup(
...
+                    genesis_hash
+                ));
+            } else {
+                break;
+            }
+        }
+    }
+
+    let index_reader = reqwest::get(index_url.clone())
+        .await?
```
abizjak, line 558: I'm sorry, but I just cannot accept downloading the file twice. I understand that it is in practice not a lot of data, but it is also not free.

The only reason for doing this seems to be that the file format is not very good for not downloading the file twice.

It seems it would be easy to change the file format and define it as
- first line: hash of the genesis of the chain
- remaining lines, comma separated list as now. You could even include the headers if you wanted to as a first line here, but I don't think it is necessary.

That would allow you to parse the first line in a special way, and the remaining ones as CSV. Since the format is so simple you don't need a CSV parser, you can just parse it directly.

abizjak (changes requested):

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +488,105 @@ async fn maybe_do_out_of_band_catchup(
...
+    while let Some(result) = chunk_records.next().await {
+        let block_chunk_data: BlockChunkData = result?;
+        // no need to reimport blocks that are present in the database
+        if block_chunk_data.genesis_index < current_genesis_index
+            || block_chunk_data.last_block_height <= last_finalized_block_height
+        {
+            info!(
+                "Skipping chunk {}: no blocks above last finalized block height",
+                block_chunk_data.filename
+            );
```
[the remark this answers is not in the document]
mkmks: The catchup log is already a wall of messages (a paragraph for _each_ imported block) but I guess I could put these behind a higher verbosity level.

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +488,105 @@ async fn maybe_do_out_of_band_catchup(
...
+
+    let comments_stream = reqwest::get(index_url.clone())
+        .await?
+        .bytes_stream()
+        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e));
+    let mut comments_lines = tokio_util::io::StreamReader::new(comments_stream).lines();
+    while let Some(line) = comments_lines.next_line().await? {
+        if line.starts_with("# genesis hash ") {
+            let index_genesis_block_hash = line.strip_prefix("# genesis hash ").unwrap();
+            let genesis_hash = genesis_block_hashes[0].to_string();
```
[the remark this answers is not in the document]
mkmks: `genesis_block_hashes` isn't coming from the catchup file. It's coming from the consensus through `regenesis_arc`, I believe. I doubt it can ever be empty, as there has to be at least one genesis block in every chain.

concordium-node/src/bin/cli.rs
mkmks: A hand-written parser for a format even as simple as a custom dialect of CSV is going to put unnecessary mental strain the next person reading this code. I suggest we use the existing library.

Removing field headers from the file is, indeed, making the file more amenable to parsing in a single streaming. Done.

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +487,104 @@ async fn maybe_do_out_of_band_catchup(
...
+        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
+        .into_async_read();
+    let mut first_line: String = String::new();
+    index_reader.read_line(&mut first_line).await?;
+
+    let index_genesis_block_hash = first_line
+        .strip_prefix("# genesis hash ")
+        .context(
+            "The catchup index file does not begin with a line containing the genesis block hash. \
+             Please contact the catchup service administrator.",
```
td202, line 546: I think it might be better to suggest checking that the URL is correct, and probably including the URL in the details.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -220,7 +221,8 @@ exportSections dbDir outDir indexHdl chunkSize genIndex startHeight = do
 -- chunk (i.e. the one containing the block with the greatest height in the section) also contains
 -- finalization records finalizing all blocks after the last block containing a finalization
 -- record. The exported chunks will be accompanied by an index file mapping chunk filenames to the
--- block ranges stored in them. Each line in index file has format FILENAME,GENESISINDEX,FIRSTBLOCK,LASTBLOCK\n
+-- block ranges stored in them. Each line in index file has format
+-- 'filename,genesis_index,first_block_height,last_block_height\n'.
```
td202, line 224:
```suggestion
-- @filename,genesis_index,first_block_height,last_block_height\n@.
```

td202 (commented):
At least on Windows, Ctrl+C during out of band catch-up does not stop the node. (It does stop importing the blocks, however.)
```
2022-08-24T09:57:57.936335100Z: INFO: Signal received attempting to shutdown node cleanly
2022-08-24T09:57:57.937784500Z: INFO: Skov: Block d260ec31745fe63382689122ce2a5222c3c32c2032b8f658e558c743adc5611e (71572) arrived
2022-08-24T09:57:57.937944800Z: INFO: Skov: Arrive statistics: blocksVerifiedCount=4631 blockLastArrive=1.6613350779336128e9 blockArriveLatencyEMA=12907.278797180054 blockArriveLatencyEMSD=115.98234132540482 blockArrivePeriodEMA=Just 9.34619384502057e-3 blockArrivePeriodEMSD=Just 7.069726127947109e-3 transactionsPerBlockEMA=7.017519248332101e-9 transactionsPerBlockEMSD=8.377063446749446e-5
2022-08-24T09:57:57.939319000Z: INFO: Out of band catchup stopped.
2022-08-24T09:57:57.939444000Z: INFO: Starting the P2P layer
2022-08-24T09:57:57.939478600Z: INFO: Attempting to bootstrap
2022-08-24T09:57:57.939518200Z: ERROR: Can't bootstrap: "No bootstrap nodes specified."
2022-08-24T09:57:57.940848800Z: DEBUG: Closing the outbound consensus channel
2022-08-24T09:57:57.940946300Z: DEBUG: Stopping the outbound consensus requests thread
2022-08-24T09:57:58.049563000Z: DEBUG: Regenesis occurred; marking all peers as pending.
2022-08-24T09:57:58.050390100Z: INFO: Network layer has been shut down.
2022-08-24T09:58:43.390147800Z: INFO: Signal received to shutdown node cleanly, but an attempt to do so is already in progress.
```

mkmks:
I don't think that this problem 
```
2022-08-24T09:58:43.390147800Z: INFO: Signal received to shutdown node cleanly, but an attempt to do so is already in progress.
```
is caused by the changes introduced in the current PR. It exists on `main` too, and not only on Windows.

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +487,107 @@ async fn maybe_do_out_of_band_catchup(
...
+
+    trace!("Current genesis index: {}", current_genesis_index);
+    trace!("Local last finalized block height: {}", last_finalized_block_height);
+
+    let mut index_reader = reqwest::get(index_url.clone())
+        .await
+        .context("Unable to download the catchup index file.")?
+        .bytes_stream()
+        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
+        .into_async_read();
```
td202, line 539: This does not handle HTTP error codes appropriately. We should check the status code on the response and fail if it is not a success (e.g. if it is a 404) with an appropriate message. Also, given that we may be downloading multiple files, we should use `reqwest::Client` so that the connection can be reused. These also apply to `download_chunk`.

td202 (changes requested):

concordium-node/src/bin/cli.rs
```diff
@@ -485,36 +487,107 @@ async fn maybe_do_out_of_band_catchup(
...
+            info!("Completed out of band catch-up from {}.", download_url)
         }
     }
 }
 
-async fn download_catchup_files(
-    download_url: url::Url,
+// An index entry for a chunk of blocks. Its format must correspond to one
+// produced by `database-exporter`.
+#[allow(dead_code)]
```
abizjak, line 510: This should be applied to the field, not to the struct.

concordium-node/src/bin/cli.rs
```diff
@@ -485,33 +488,116 @@ async fn maybe_do_out_of_band_catchup(
...
+        genesis_hash
+    );
+
+    let mut chunk_records = csv_async::AsyncReaderBuilder::new()
+        .comment(Some(b'#'))
+        .has_headers(false)
+        .create_deserializer(index_reader)
+        .into_deserialize();
+    while let Some(result) = chunk_records.next().await {
+        let block_chunk_data: BlockChunkData = result?;
```
abizjak, line 576: You should also stop this loop if `import_stopped` is set so as to not download the remaining files.

concordium-node/src/bin/cli.rs
```diff
@@ -522,9 +618,18 @@ async fn download_catchup_files(
         tempfile::NamedTempFile::new_in(data_dir_path).context("Cannot create output file.")?;
     info!("Downloading the catch-up file from {} to {}", download_url, temp_file.path().display());
     {
+        let chunk_response = http_client.get(download_url.clone()).send().await?;
+        anyhow::ensure!(
+            chunk_response.status() == StatusCode::OK,
```
abizjak, line 623:
```suggestion
            chunk_response.status().is_success(),
```

concordium-node/src/bin/cli.rs
```diff
@@ -485,33 +488,126 @@ async fn maybe_do_out_of_band_catchup(
...
+    );
+
+    let mut chunk_records = csv_async::AsyncReaderBuilder::new()
+        .comment(Some(b'#'))
+        .has_headers(false)
+        .create_deserializer(index_reader)
+        .into_deserialize();
+    while let Some(result) = chunk_records.next().await {
+        if import_stopped.load(atomic::Ordering::Acquire) {
+            break;
```
abizjak, line 583: This is not entirely correct. In case the import is stopped while consensus has control the logs will say 
```
"Out of band catchup stopped."
```
but in case it is stopped while, e.g., parsing the index file, the logs will contain
```
"Completed out of band catch-up from ..."
```

It would be consistent to instead return an Err here so that the control flows are consistent.

```rust
ensure!(!import_stopped.load(atomic::Ordering::Acquire), "Import stopped by the user.")
```

abizjak (approved):

td202 (approved):
Looks good.

Issue #464, opened 2022-08-22
Performance regression in start-up time


td202:
**Bug Description**

The account caching changes introduced a performance regression in the startup time from an existing database.

**Steps to Reproduce**

Start the node from a recent mainnet database

**Expected Result**

Startup time should be on the order of 20 seconds.

**Actual Result**

Startup time is on the order of 80 seconds.

**Versions**
- Software Version 4.3.0

PR #465, opened 2022-08-22
Fix regression in start-up time

td202:
## Purpose

Addresses #464.

The account caching changes introduced a performance regression in the node start-up time with an existing database. This rectifies that. The most serious issue was that in the `Cacheable` instance for `HashedBufferedRef'`, the hash was computed on the uncached data, which lead to the data being loaded multiple times. A secondary issue was that many accounts were loaded twice because the accounts were traversed to construct the transaction table.

Anecdotally, start-up time is reduced to around 18 seconds for mainnet, compared with around 28 seconds for 4.2.3 and around 57 seconds for the previous 4.3.0.

## Changes

- Introduced a `Cacheable1` class that allows lifting a cache operation on an inner type to a cache operation on a container type.
- Implemented `Cacheable1` instances for reference types (`BufferedRef`, `HashedBufferedRef`, `HashedCachedRef`), for `LFMBTree'` and for `Accounts`.
- Introduced `cacheStateAndGetTransactionTable` for the persistent block state that amalgamates and replaces the previous functionality of `cacheBlockState` and `getInitialTransactionTable`. This makes use of the `Cacheable1` instance for `Accounts` to initialise the transaction table while the accounts are cached, avoiding a second pass over the transaction table, which would result in re-loading the accounts.
- Removed `cacheBlockState` and `getInitialTransactionTable` from the `BlockStateStorage` and `BlockStateQuery` interfaces respectively. As `getInitialTransactionTable` was still used in the basic (in-memory) block state for re-genesis, that version was retained in a modified form.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Basic/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Basic/BlockState.hs
@@ -2090,8 +2090,9 @@ genesisState gd = case protocolVersion @pv of
                             StateMigrationParametersP3ToP4{} -> False
                         genesisTT = getInitialTransactionTable hbs
 
--- |Construct a transaction table that is empty but records the next nonces/sequence numbers
--- for all accounts and chain updates.
+-- |Construct a transaction table that is empty but reflects the next nonces/sequence numbers
+-- for all accounts and chain updates. That is, if the next nonce/sequence number is not the minimal
+    -- one, it is recorded in the transaction table.
 getInitialTransactionTable ::
     ( HasBlockState s pv,
       IsChainParametersVersion (ChainParametersVersionFor pv)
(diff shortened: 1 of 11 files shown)


concordium-consensus/src/Concordium/GlobalState/Basic/BlockState.hs
```diff
@@ -2112,4 +2088,32 @@ genesisState gd = case protocolVersion @pv of
                         hashShouldMatch = case migration of
                             StateMigrationParametersTrivial{} -> True
                             StateMigrationParametersP3ToP4{} -> False
-                        genesisTT = runIdentity $ runPureBlockStateMonad $ BS.getInitialTransactionTable hbs
+                        genesisTT = getInitialTransactionTable hbs
+
+-- |Construct a transaction table that is empty but records the next nonces/sequence numbers
+-- for all accounts and chain updates.
```
abizjak, line 2094: Isn't this comment somewhat misleading? The transaction table only explicitly records the nonces for accounts which have non-initial nonces.

abizjak (approved):
Looks good.

EmilLa1 (approved):

PR #472, opened 2022-08-24
Fix "Out of memory" issue with block importing

### LINKED ISSUE #457: A suitably crafted catch-up file can cause the node to crash with "Out of memory"
**Bug Description**

An out-of-band catch-up file can be crafted which causes the node to crash with an "Out of memory" error.

This issue is caused by the use of [`Data.ByteString.hGet`](https://hackage.haskell.org/package/bytestring-0.11.3.1/docs/Data-ByteString.html#v:hGet) with a length previously read from the file. If the file is malformed, this length can be greater than the memory allocator will allow, leading to a crash. This can be mitigated, for instance, by using `Data.ByteString.Lazy.hGet` instead.

**Steps to Reproduce**

Start a node with the attached database specified as `--blocks-to-import`: [badDB3.zip](https://github.com/Concordium/concordium-node/files/9338199/badDB3.zip)

**Expected Result**

The node should report a serialization error with the catch-up file, but continue running normally.

**Actual Result**

The node crashes with output similar to the following:
```
2022-08-15T15:21:57.571876900Z: INFO: Starting out of band catch-up
2022-08-15T15:21:57.571554900Z: INFO: Runner: Transaction purging thread started.
concordium-node.exe: Out of memory

concordium-node.exe: warning: too many hs_exit()s
```

**Versions**
- Software Version
- OS
- Browser
- Mobile device, if applicable

td202:
## Purpose

Fixes #457.

This prevents a corrupted catch-up file from causing an out of memory error by checking that there are sufficient bytes remaining in the file to read a bytestring of a given length.

## Changes

- After deserializing the length of a bytestring, check that there are enough bytes left in the file before attempting to deserialize the contents of the bytestring.
- Remove unused code from a previous version of the block importer, which would be subject to the same issue.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -414,39 +415,41 @@ importBlocksV3 inFile firstGenIndex cbk = runExceptT $
...
                                 ImportFinalizationRecord sectionProtocolVersion sectionGenesisIndex
                             )
                     -- Move to the next section
                     liftIO $ hSeek hdl AbsoluteSeek (sectionStart + toInteger sectionLength)
-                    importSections hdl
-    importData :: Handle -> (BS.ByteString -> ImportData) -> ExceptT (ImportFailure a) m ()
-    importData hdl makeImport = do
-        blockBS <- getLengthByteString hdl
+                    importSections hdl fileSize
+    importData :: Handle -> Integer -> (BS.ByteString -> ImportData) -> ExceptT (ImportFailure a) m ()
```
abizjak, line 440: Can you please add a comment to this function documenting the meaning of the value and the reason for its existence.

abizjak (approved):

mkmks (approved):
LGTM

PR #478, opened 2022-08-30
Check pending blocks

eb-concordium:
## Purpose

To do a bit more validation of pending blocks when possible. 

## Changes

* A function `predictLeadershipElectionNonce` has been added. It can predict the leadership election nonce for the next epoch as long as the last finalized block is in the last 1/3 of the current epoch.
* A function `getDefiniteSlotBakers` has been added. It returns the bakers that apply to a future slot, if it is certain that it is correct. Otherwise it returns `Nothing`. 
* If `getDefiniteSlotBakers` returns Nothing, we only check the block signature. Otherwise, for a pending block we also check:
   ** Whether the claimed baker is valid in the committee
   ** If we can predict the leadership election nonce for a pending block, we do it and mark the block as dead if the proof is incorrect with the given nonce.
* Pending blocks will not be relayed anymore.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,8 @@
 
 ## Unrelease changes
 
-- A function `predictLeadershipElectionNonce` added. It can predict the leadership election nonce for the next epoch as long as the last finalized block is in the last 1/3 of the current epoch.
-- A function `getDefiniteSlotBakers` added. It returns the bakers that apply to a future slot, if it is certain that the bakers are correct. Otherwise it returns Nothing.
-- If `getDefiniteSlotBakers` returns Nothing, we only check the block signature. Otherwise, for a pending block we also check
-  - whether the claimed baker is valid in the committee
-  - if we can predict the leadership election nonce for a pending block, we do it and mark the block as dead if the proof is incorrect with the given nonce
-- Pending blocks will not be relayed anymore.
+- Verify pending blocks earlier when possible.
+- Do not relay pending blocks.
 
 
 ## 4.3.0


concordium-consensus/tests/consensus/ConcordiumTests/LeaderElectionTest.hs
```diff
@@ -0,0 +1,145 @@
...
+        (el :: Word64) <- (*9) <$> choose (1,10000)
+        epoch <- choose (0, 1000)
+        return SeedState {
+            epochLength = fromIntegral el,
+            currentLeadershipElectionNonce = H.hash "1",
+            updatedNonce = H.hash "2",
+            ..
+        }
+
+-- Tests that both Maybe List's are Just and that former contains the latter as a sublist
```
abizjak, line 71: You check `IsInfixOf`, not that it is a sublist. Perhaps update the documentation.

stack.yaml
```diff
@@ -13,3 +13,6 @@ extra-deps:
 extra-lib-dirs:
 - ./concordium-base/lib
 - ./concordium-base/smart-contracts/lib
+
+ghc-options:
+  "$locals": -fwrite-ide-info
```
abizjak, line 18: This does not belong to this PR.

concordium-consensus/src/Concordium/Skov/Update.hs
```diff
@@ -478,43 +498,97 @@ doStoreBlock pb@GB.PendingBlock{..} = unlessShutDown $ do
...
+                                                bkrPower
+                                                (blockProof pb)
+                                        if any verifyProofWithNonce nonceList
+                                        then do
+                                            continuePending
+                                        else do
+                                            logEvent Skov LLWarning $ "Block proof of pending block was not verifiable: " ++ show pb
+                                            processDead
+                        Nothing -> processDead
+                    Nothing -> continuePending
```
abizjak, line 571: Is it an explicit choice here to do the signature check at the very end of processing?

concordium-consensus/tests/consensus/ConcordiumTests/LeaderElectionTest.hs
```diff
@@ -0,0 +1,147 @@
...
+                        counterexample "Seedstate slot too early and prediction slot too late, parent in seedstate epoch" $ predictLeadershipElectionNonce ss seedStateSlotTooEarly (Just parentInThisEpoch) slotToPredictTooLate === Nothing,
+                        counterexample "Seedstate slot too early and prediction slot too late, parent in epoch after seedstate epoch" $ predictLeadershipElectionNonce ss seedStateSlotTooEarly (Just parentInNextEpoch) slotToPredictTooLate === Nothing
+                    ]
+
+testPredictionJust :: Property
+testPredictionJust = label "Prediction corrrect" makePropertyJust
+
+
+testPredictionNothing :: Property
+testPredictionNothing = label "No prediction possible" makePropertyNothing
```
td202, line 141: I am not sure that the `label`s add anything here. They're supposed to be for reporting test case distribution, but since there's no non-determinism in the choice, these will always just report 100% for whichever label.

concordium-consensus/tests/consensus/ConcordiumTests/LeaderElectionTest.hs
```diff
@@ -0,0 +1,147 @@
...
+            counterexample "Prediction invariant under seedstate update, with parent in same epoch as the last finalized block" $ predictLeadershipElectionNonce ss seedStateSlot (Just parentInThisEpoch) slotToPredictFor === predictLeadershipElectionNonce ss'' seedStateSlot'' (Just parentInThisEpoch) slotToPredictFor,
+            counterexample "Prediction invariant under seedstate update across an epoch, with parent in the same epoch as the target slot" $ predictLeadershipElectionNonce ss seedStateSlot (Just parentInNextEpoch) slotToPredictFor === predictLeadershipElectionNonce ss' seedStateSlot' (Just parentInNextEpoch) slotToPredictFor,
+            counterexample "Prediction invariant under seedstate update in same epoch, with parent in the same epoch as the target slot" $ predictLeadershipElectionNonce ss seedStateSlot (Just parentInNextEpoch) slotToPredictFor === predictLeadershipElectionNonce ss'' seedStateSlot'' (Just parentInNextEpoch) slotToPredictFor
+        ]
+
+-- Tests cases where the leadership election nonce cannot be predicted, either beceause the seedstate slot is too early (i.e. in the first 2/3 of the epoch), or
```
td202, line 109:
```suggestion
-- Tests cases where the leadership election nonce cannot be predicted, either because the seedstate slot is too early (i.e. in the first 2/3 of the epoch), or
```

concordium-consensus/tests/consensus/ConcordiumTests/LeaderElectionTest.hs
```diff
@@ -0,0 +1,147 @@
...
+-- on this seedstate and a target slot in the next epoch. It tests against @computeLeadershipElectionNonce@ and tests that
+-- @predictLeadershipElectionNonce@ gives the same prediction on an updated seedstate for the same target slot.
+makePropertyJust :: Property
+makePropertyJust = property $ do
+    ss@SeedState{..} <- genSeedstate
+    let epochSlot = epoch * fromIntegral epochLength
+    let nextEpoch = epochSlot + fromIntegral epochLength
+    let toThirds = 2 * fromIntegral epochLength `div` 3
+    -- Find a slot in the last 2/3 of the epoch
+    seedStateSlot :: Slot <- fromIntegral <$> choose (epochSlot + toThirds, nextEpoch-2) -- We substract 2 so that we can update the seedstate and still be in the same epoch.
```
td202, line 88:
```suggestion
    let twoThirds = 2 * fromIntegral epochLength `div` 3
    -- Find a slot in the last 2/3 of the epoch
    seedStateSlot :: Slot <- fromIntegral <$> choose (epochSlot + twoThirds, nextEpoch-2) -- We subtract 2 so that we can update the seedstate and still be in the same epoch.
```

td202 (approved):
Looks good. A couple more things to do before merging:
- Update the changelog.
- The change list in the PR should also mention the `getDefiniteSlotBakers`, and should spell out what is checked for a pending block.

abizjak (approved):

Issue #433, opened 2022-07-12
Implement the new grpc API in the node


abizjak:
**Task description**

Depends on
- #431 
- #432 (likely to be done in combination with it)

This task covers the implementation of the new API on the node.

This API should be exposed in addition to the existing one, and the node runner should be able to choose which one (or none or both) to run for the time being.
## Consensus queries
   - [x] GetAccountList
   - [x] GetAccountInfo
   - [x] GetInstanceList
   - [x] GetInstanceInfo
   - [x] GetAncestors
   - [x] GetModuleSource
   - [x] GetModuleList
   - [x] GetFinalizedBlocks
   - [x] GetNextAccountNonce
   - [x] GetConsensusInfo
   - [x] GetCryptographicParameters
   - [x] InvokeContract
   - [x] GetBlockInfo
   - [x] GetBakerList
   - [x] GetPoolStatus
   - [x] GetDelegators
   - [x] GetTokenomicsStatus
   - [x] GetBlocksAtHeight
   - [x] SendTransaction
      - [x] Header fully typed
      - The response should be the transaction hash.
      - For payloads we want to first model the common ones and have an extra
        "other" payload which just takes a serialized transaction payload. (More can be added later)
      The following should be modelled.
          - [x] Transfer (with and without memo)
          - [x] DeployModule
          - [x] InitContract
          - [x] UpdateContract
          - [x] RegisterData
   - [x] GetTransactionStatus
   - [x] GetBranches
   - [x] GetBirkParameters
   - [x] GetIdentityProviders
   - [x] GetAnonymityRevokers
   - [x] GetAccountNonFinalizedTransactions
   - [ ] GetBlockTransactions
   - [x] GetBlockEvents
   - [ ] GetPendingUpdates
   - [ ] GetChainParameters
   - [x] GetBlocks
   - [ ] Split `types.proto` into multiple files to help the compilers out (the generated file is quite large).
   
## Node queries   
   - [x] PeerConnect
   - [x] PeerDisconnect
   - [x] GetNodeStatus
   - [x] GetPeersInfo
   - [x] DumpStop
   - [x] DumpStart
   - [x] BanNode
   - [x] UnbanNode
   - [x] GetBannedPeers
   - [x] Shutdown

PR #482, opened 2022-09-07
First round of gRPC api v2 functions

limemloh:
## Purpose

Implement some of the new gRPC api v2 functions. https://github.com/Concordium/concordium-node/issues/433
Depends on https://github.com/Concordium/concordium-grpc-api/pull/12.

## Changes

- Add `GetModuleList`, `GetModuleSource`, `GetInstanceList`, `GetInstanceInfo`, `GetNextAccountSequenceNumber`, `GetConsensusInfo`, `GetAncestors` and `GetTransactionStatus`.

## Checklist

      hard-to-understand areas.


concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1046,6 +1135,43 @@ impl ConsensusContainer {
         Ok((out_hash, out_data))
     }
 
+    /// Get the best guess as to what the next account sequence number should
+    /// be. If all account transactions are finalized then this information
+    /// is reliable. Otherwise this is the best guess, assuming all other
+    /// transactions will be committed to blocks and eventually finalized.
```
Bargsteen, line 1141:
```suggestion
    /// Get the best guess as to what the next account sequence number should
    /// be. If all account transactions are finalized, then this information
    /// is reliable. Otherwise, this is the best guess, assuming all other
    /// transactions will be committed to blocks and eventually finalized.
```

concordium-node/src/grpc2.rs
```diff
@@ -403,12 +397,16 @@ pub mod server {
     impl service::queries_server::Queries for RpcServerImpl {
         type GetAccountListStream =
             futures::channel::mpsc::Receiver<Result<Vec<u8>, tonic::Status>>;
+        type GetAncestorsStream = futures::channel::mpsc::Receiver<Result<Vec<u8>, tonic::Status>>;
```
Bargsteen, line 400: We might consider documenting these. Although, I don't know if it is important.

concordium-node/src/grpc2.rs
```diff
@@ -480,10 +478,102 @@ pub mod server {
...
+            let mut response = tonic::Response::new(response);
+            add_hash(&mut response, hash)?;
+            Ok(response)
+        }
     }
 }
 
-/// Add hash to the provided response. The hash is added as part of metadata.
+/// Add a block hash to the metadata of a response. Used for returning the block
+/// hash for streaming responses.
```
Bargsteen, line 576:
```suggestion
/// hash.
```

Bargsteen (commented):
Looks good in general. Just minor findings. 
If we decide to change the module source to return `(version, wasmModule)` as Ales suggested, then we need to handle it here.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1046,6 +1126,43 @@ impl ConsensusContainer {
...
+            getNextAccountSequenceNumberV2(
+                consensus,
+                acc_type,
+                acc_id,
+                &mut out_data,
+                copy_to_vec_callback,
+            )
+            .try_into()?
+        };
+        response.ensure_ok("account")?;
```
abizjak, line 1152: account nonce?

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1046,6 +1126,43 @@ impl ConsensusContainer {
...
+        Ok(out_data)
+    }
+
+    /// Get information of the current state of consensus.
+    pub fn get_consensus_info_v2(&self) -> Result<Vec<u8>, tonic::Status> {
+        let consensus = self.consensus.load(Ordering::SeqCst);
+        let mut out_data: Vec<u8> = Vec::new();
+        let _response: ConsensusQueryResponse = unsafe {
+            getConsensusInfoV2(consensus, &mut out_data, copy_to_vec_callback).try_into()?
+        };
```
abizjak, line 1162: Why not check the response? I suppose it should never fail, but please document this.

concordium-node/src/grpc2.rs
abizjak: I think it is easy to do and will provide value for non-expert users of tonic. Just a line saying something like
"Return value for the `get_ancestors` endpoint."

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -46,8 +48,11 @@ import Concordium.Crypto.SHA256 (DigestSize, Hash (Hash))
 import Concordium.Crypto.SignatureScheme (VerifyKey (..))
 import Concordium.Types.Accounts.Releases
 import Concordium.Types.Execution
+import qualified Concordium.Wasm as Wasm
 import Data.Text (Text)
 import Data.Time (UTCTime)
+import Concordium.Wasm (InstanceInfo(iiSourceModule))
+import Concordium.Types.Block (AbsoluteBlockHeight(..))
```
abizjak, line 55: I'd prefer if you did not both include Wasm qualified, and also imported some things unqualified.

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -439,6 +562,9 @@ decodeAccountIdentifierInput 1 dta = do
 decodeAccountIdentifierInput 2 dta = AccIndex . AccountIndex <$> peek (castPtr dta)
 decodeAccountIdentifierInput n _ = error $ "Unknown account identifier tag: " ++ show n
 
+decodeModuleRefInput :: Ptr Word8 -> IO ModuleRef
```
abizjak, line 565: I don't think you are checking on the Rust end that the moduleRef is actually 32 bytes. So this function is problematic.

For block hashes and account addresses we check on the Rust end that they are of correct length, so we can then do this.

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -546,29 +826,120 @@ enqueueMessages callback = forkIO . go 0 . map encodeMsg
...
+
+foreign export ccall
+    getNextAccountSequenceNumberV2 ::
+        StablePtr Ext.ConsensusRunner ->
+        -- |Identifier type, 0 for account address, 1 for credential, 2 for account index
+        Word8 ->
+        -- |Serialized identifier. Length determined by the type.
+        Ptr Word8 ->
+        Ptr ReceiverVec ->
+        FunPtr (Ptr ReceiverVec -> Ptr Word8 -> Int64 -> IO ()) ->
```
abizjak, line 937: You also need to add these new functions to the `lib.def`  file. Otherwise Windows people will dislike you :)

abizjak (commented):
Looks pretty good overall. There is one issue that definitely needs fixing, and some minor ones.

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -497,6 +694,168 @@ getAccountListV2 cptr channel blockType blockHashPtr outHash cbk = do
...
+    -- |Callback to output data.
+    FunPtr CopyToVecCallback ->
+    IO Int64
+getModuleSourceV2 cptr blockType blockHashPtr moduleRefPtr outHash outVec copierCbk = do
+    Ext.ConsensusRunner mvr <- deRefStablePtr cptr
+    let copier = callCopyToVecCallback copierCbk
+    bhi <- decodeBlockHashInput blockType blockHashPtr
+    modRef <- decodeModuleRefInput moduleRefPtr
+    res <- runMVR (Q.getModuleSource bhi modRef) mvr
+    -- TODO: Consider splitting the module in chunks and streaming the results.
```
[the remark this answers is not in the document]
limemloh: It seems not to make sense for gRPC streams, to do this.

concordium-node/src/consensus_ffi/ffi.rs
limemloh: I have changed the sequence number to always succeed, as in gRPC api v1, so this check is not needed.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -500,16 +500,96 @@ extern "C" {
...
+        block_id_type: u8,
+        block_hash: *const u8,
+        acc_type: u8,
+        acc_id: *const u8,
+        out_hash: *mut u8,
+        out: *mut Vec<u8>,
+        copier: CopyToVecCallback,
+    ) -> i64;
+
+    pub fn getNextAccountSequenceNumberV2(
```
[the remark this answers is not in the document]
limemloh: I have added more doc comments here and for every grpc api v2 ffi call.

concordium-node/src/consensus_ffi/ffi.rs
limemloh: I have added a comment

concordium-consensus/src/Concordium/GRPC2.hs
limemloh: I have added a check on the rust side now

concordium-node/src/consensus_ffi/ffi.rs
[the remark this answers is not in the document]
abizjak: Looks good!

abizjak (approved):

Bargsteen (approved):

PR #483, opened 2022-09-07
Block removal until the last with a valid blockstate

### LINKED ISSUE #480: Treestate/blockstate rollback to the explicitly finalized block with a valid blockstate
Implement operations to delete blocks, finalization records and transaction records from LMDB. Then, recursively delete blocks together with their finalization records and transactions if their blockstate cannot be read from the blob store, starting from the last finalized block. Stop deleting at the block which both has a valid blockstate in the blob store and a finalization record. Truncate the blob store after that valid blockstate.

# Acceptance criteria

It should be possible to start Skov with after the rollback (#477).

mkmks:
## Purpose

Closes #480. 

## Changes

Adds `deleteFinalizedBlocksWhile` which traverses blocks in the treestate, starting with the last finalized block, while they match a predicate and deletes them. When used with the predicate checking for validity of the corresponding blockstate, it acts as a tree state rollback mechanism in case of the blob store corruption.
Adds `truncateBlobStore` which removes all content of the blob store following a blob specified by a reference. When given a reference to the last blob that can be successfully read, it performs the blob store cleanup after its corruption.

`loadSkovPersistentData` now attempts a rollback after the last finalized block is read from LMDB. If a rollback isn't needed, it doesn't happen.

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
@@ -36,8 +36,6 @@ module Concordium.GlobalState.LMDB.Helpers (
   movePrimitiveCursor,
   withPrimitiveCursor,
   PrimitiveCursor,
-  traverseTable,
-  popFromTableWhile,
   loadAll,
 
   -- * Low level operations.
@@ -428,54 +426,6 @@ movePrimitiveCursor target PrimitiveCursor{..} = do
       Just <$> peek pcValPtr
     else
       return Nothing
-
--- |Traverse a database from the first record to last, using a cursor internally.
--- 'MDB_val' values must not be retained as they will be reused.
-traverseTable :: MDBDatabase db => MDB_txn -> db -> ((MDB_val, MDB_val) -> a -> IO a) -> a -> IO a
-traverseTable = traverseTable' (CursorFirst, CursorNext) (const $ return True)
-
--- |Traverse a database backwards while a condition holds and collect results.
-popFromTableWhile :: forall db. MDBDatabase db => ((DBKey db, DBValue db) -> IO Bool) -> MDB_txn -> db -> IO (Maybe [(DBKey db, DBValue db)])
-popFromTableWhile p txn db = traverseTable' (CursorLast, CursorPrevious) p txn db step (Just [])
-  where step mdbKV (Just acc) = do
-          ekv <- decodeKV prox mdbKV
-          case ekv of
-            Left _ -> return Nothing
-            Right kv -> return (Just (kv : acc))
-        step _ Nothing = return Nothing
-        prox :: Proxy db
-        prox = Proxy
-
--- |Traverse a database, using a cursor internally, while a condition holds.  'MDB_val' values must
--- not be retained as they will be reused.  The `(first, next)` tuple indicates the beginning and
--- the direction of the traversal. Its possible values are `(CursorFirst, CursorNext)` and
--- `(CursorLast, CursorPrevious)`.
-traverseTable' :: forall db a.  MDBDatabase db
-  => (CursorMove, CursorMove)
-  -> ((DBKey db, DBValue db) -> IO Bool)
-  -> MDB_txn
-  -> db
-  -> ((MDB_val, MDB_val) -> a -> IO a)
-  -> a
-  -> IO a
-traverseTable' (first, next) p txn db step initialAcc =
-    withPrimitiveCursor txn (mdbDatabase db) $ \cursor -> do
-      let
-          loop (Just mdbKV) acc = decodeKV prox mdbKV >>= \ekv  -> case ekv of
-            Left e -> error $ "traverseTable': " <> e
-            Right kv  -> do
-              isP <- p kv
-              if isP then do
-                nxt <- step mdbKV acc
-                nxtRes <- getPrimitiveCursor next cursor
-                loop nxtRes nxt
-              else return acc
-          loop Nothing acc = return acc
-      fstRes <- getPrimitiveCursor first cursor
-      loop fstRes initialAcc
-  where
-    prox :: Proxy db
-    prox = Proxy
 
 -- |A type-safe cursor over a database.
 newtype Cursor db = Cursor PrimitiveCursor
--- a/concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
@@ -25,6 +25,7 @@ module Concordium.GlobalState.LMDB.Helpers (
   storeReplaceRecord,
   storeReplaceBytes,
   loadRecord,
+  deleteRecord,
   databaseSize,
 
   -- * Traversing the database.
@@ -36,6 +37,7 @@ module Concordium.GlobalState.LMDB.Helpers (
   withPrimitiveCursor,
   PrimitiveCursor,
   traverseTable,
+  popFromTableWhile,
   loadAll,
 
   -- * Low level operations.
@@ -427,19 +429,53 @@ movePrimitiveCursor target PrimitiveCursor{..} = do
     else
       return Nothing
 
--- |Traverse a database, using a cursor internally.
+-- |Traverse a database from the first record to last, using a cursor internally.
 -- 'MDB_val' values must not be retained as they will be reused.
-traverseTable :: MDB_txn -> MDB_dbi' -> (a -> MDB_val -> MDB_val -> IO a) -> a -> IO a
-traverseTable txn db step start =
-    withPrimitiveCursor txn db $ \cursor -> do
+traverseTable :: MDBDatabase db => MDB_txn -> db -> ((MDB_val, MDB_val) -> a -> IO a) -> a -> IO a
+traverseTable = traverseTable' (CursorFirst, CursorNext) (const $ return True)
+
+-- |Traverse a database backwards while a condition holds and collect results.
+popFromTableWhile :: forall db. MDBDatabase db => ((DBKey db, DBValue db) -> IO Bool) -> MDB_txn -> db -> IO (Maybe [(DBKey db, DBValue db)])
+popFromTableWhile p txn db = traverseTable' (CursorLast, CursorPrevious) p txn db step (Just [])
+  where step mdbKV (Just acc) = do
+          ekv <- decodeKV prox mdbKV
+          case ekv of
+            Left _ -> return Nothing
+            Right kv -> return (Just (kv : acc))
+        step _ Nothing = return Nothing
+        prox :: Proxy db
+        prox = Proxy
+
+-- |Traverse a database, using a cursor internally, while a condition holds.  'MDB_val' values must
+-- not be retained as they will be reused.  The `(first, next)` tuple indicates the beginning and
+-- the direction of the traversal. Its possible values are `(CursorFirst, CursorNext)` and
+-- `(CursorLast, CursorPrevious)`.
+traverseTable' :: forall db a.  MDBDatabase db
+  => (CursorMove, CursorMove)
+  -> ((DBKey db, DBValue db) -> IO Bool)
+  -> MDB_txn
+  -> db
+  -> ((MDB_val, MDB_val) -> a -> IO a)
+  -> a
+  -> IO a
+traverseTable' (first, next) p txn db step initialAcc =
+    withPrimitiveCursor txn (mdbDatabase db) $ \cursor -> do
       let
-          loop Nothing cur = return cur
-          loop (Just (key, val)) cur = do
-            nxt <- step cur key val
-            nxtRes <- getPrimitiveCursor CursorNext cursor
-            loop nxtRes nxt
-      fstRes <- getPrimitiveCursor CursorFirst cursor
-      loop fstRes start
+          loop (Just mdbKV) acc = decodeKV prox mdbKV >>= \ekv  -> case ekv of
+            Left e -> error $ "traverseTable': " <> e
+            Right kv  -> do
+              isP <- p kv
+              if isP then do
+                nxt <- step mdbKV acc
+                nxtRes <- getPrimitiveCursor next cursor
+                loop nxtRes nxt
+              else return acc
+          loop Nothing acc = return acc
+      fstRes <- getPrimitiveCursor first cursor
+      loop fstRes initialAcc
+  where
+    prox :: Proxy db
+    prox = Proxy
 
 -- |A type-safe cursor over a database.
 newtype Cursor db = Cursor PrimitiveCursor
@@ -451,14 +487,17 @@ withCursor txn db op = withPrimitiveCursor txn (mdbDatabase db) (op . Cursor)
 
 -- |Move a cursor and read the key and value at the new location.
 getCursor :: forall db. (MDBDatabase db) => CursorMove -> Cursor db -> IO (Maybe (Either String (DBKey db, DBValue db)))
-getCursor movement (Cursor pc) = getPrimitiveCursor movement pc >>= mapM decodeKV
+getCursor movement (Cursor pc) = getPrimitiveCursor movement pc >>= mapM (decodeKV prox)
   where
-    decodeKV (keyv, valv) = runExceptT $ do
-      key <- ExceptT $ decodeKey prox keyv
-      val <- ExceptT $ decodeValue prox key valv
-      return (key, val)
     prox :: Proxy db
     prox = Proxy
+
+-- |Decode a key-value pair from the database internal representation.
+decodeKV :: MDBDatabase db => Proxy db -> (MDB_val, MDB_val) -> IO (Either String (DBKey db, DBValue db))
+decodeKV prox (keyv, valv) = runExceptT $ do
+  key <- ExceptT $ decodeKey prox keyv
+  val <- ExceptT $ decodeValue prox key valv
+  return (key, val)
 
 -- |Load all key value pairs from a database. Raises an error on a serialization failure.
 loadAll :: forall db. (MDBDatabase db) => MDB_txn -> db -> IO [(DBKey db, DBValue db)]
@@ -545,6 +584,22 @@ loadRecord txn dbi key = do
       Just bval -> decodeValue prox key bval >>= \case
         Left _ -> return Nothing
         Right val -> return (Just val)
+  where
+    prox :: Proxy db
+    prox = Proxy
+
+-- |Delete a record at the specified key. Returns True on success, or False if no value exists for
+-- the specified key.
+deleteRecord :: forall db. (MDBDatabase db)
+  => MDB_txn
+  -- ^Transaction
+  -> db
+  -- ^Table
+  -> DBKey db
+  -- ^Key
+  -> IO Bool
+deleteRecord txn dbi key = do
+  withMDB_val (encodeKey prox key) $ \val -> mdb_del' txn (mdbDatabase dbi) val Nothing
   where
     prox :: Proxy db
     prox = Proxy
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +694,51 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
+  DatabaseHandlers pv st -> (StoredBlock pv st -> IO Bool) -> IO ()
+deleteBlocksBy dbh p = transaction (dbh ^. storeEnv) False $ \txn -> do
+  popFromTableWhile (p . snd) txn (dbh ^. blockStore) >>=
+    \case
+      Just hb -> do
+        let (blockHashes, storedBlocks) = unzip hb
+        mapM_ (deleteRecord txn (dbh ^. blockStore)) blockHashes
+        mapM_ (deleteRecord txn (dbh ^. finalizationRecordStore)) (sbFinalizationIndex <$> storedBlocks)
+        mapM_ (deleteRecord txn (dbh ^. transactionStatusStore)) (concatMap ((wmdHash <$>) . blockTransactions . sbBlock) storedBlocks)
+      Nothing -> return ()
```
abizjak, line 710: This is not used, it is also wrong since it iterates the block table by block hashes (lexicographically). So essentially in a random order.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +694,51 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
+                  -- parent is not finalized by the same finalization record.
+                  Just sbParent -> do
+                    delFinRecOK <- deleteRecord txn (h ^. finalizationRecordStore) finIndex
+                    if delFinRecOK
+                      then loopBlocks (sbFinalizationIndex sbParent) sbParent
+                      else return . Left $ "Could not delete finalization record with index "
+                           <> show finIndex <> " finalizing block " <> show bh
+                  Nothing -> return . Left $ "Could not read parent of block " <> show bh
+            else return . Left $ "Could not delete block " <> show bh
+          else return $ Right bh
```
abizjak, line 739: I don't see you deleting the index of blocks by height.

abizjak: This means that you are breaking invariants in the DB.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +694,51 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
+      -- genesis blocks won't be deleted
```
abizjak, line 740: I don't see you deleting anything from the `finalizedByHeightStore` which means you are corrupting the LMDB database.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +694,51 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
+                mapM_ (deleteRecord txn (h ^. transactionStatusStore)) ((wmdHash <$>) . blockTransactions $ block)
+                loadRecord txn (h ^. blockStore) (blockPointer block) >>= \case
+                  Just sbParent | finIndex == sbFinalizationIndex sbParent -> loopBlocks finIndex sbParent
+                  -- Before deleting a finalization record for a block, we must ensure that its
+                  -- parent is not finalized by the same finalization record.
+                  Just sbParent -> do
+                    delFinRecOK <- deleteRecord txn (h ^. finalizationRecordStore) finIndex
+                    if delFinRecOK
+                      then loopBlocks (sbFinalizationIndex sbParent) sbParent
+                      else return . Left $ "Could not delete finalization record with index "
```
abizjak, line 735: It's a bit non-obvious why this would also work for when we reach genesis. It relies on the finalization index of the genesis block being 0, the only one. So please provide documentation for the future understanding of this code.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -378,10 +379,17 @@ loadSkovPersistentData rp _treeStateDirectory pbsc = do
     GenesisBlock gd' -> return gd'
     _ -> logExceptionAndThrowTS (DatabaseInvariantViolation "Block at height 0 is not a genesis block.")
 
+  -- Unroll the treestate if the last finalized blockstate is corrupted. If the last finalized
+  -- blockstate is not corrupted, the treestate is unchanged.
+  deleteFinalizedBlocksWhile _db isBlockStateCorrupted >>= \case
+    Left e -> logExceptionAndThrowTS . DatabaseInvariantViolation $ "Database recovery failed: " <> e
+    Right bh -> logEvent GlobalState LLInfo $ "Local treestate and blockstate unrolled. Last finalized block hash is now " <> show bh
```
abizjak, line 386: This always logs a message even if nothing was removed. That is confusing, so please do not log "... unrolled ..." if nothing was removed.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -390,8 +398,11 @@ loadSkovPersistentData rp _treeStateDirectory pbsc = do
     readBlobBSFromHandle (bscBlobStore . PBS.pbscBlobStore $ pbsc) (sbState lfStoredBlock)
   case elfBlob :: Either IOError BS.ByteString of
     Left _ -> logExceptionAndThrowTS $
-      DatabaseInvariantViolation "Last finalized block cannot be read. Blockstate database recovery required."
+        DatabaseInvariantViolation "Last finalized block cannot be read. Blockstate database recovery required."
```
abizjak, line 401: This check  does not seem apt anymore.

The only time we'd see this now, it appears to me, is if the state for the genesis block was not written properly. So we could have a more precise error message. Recovery in this case is simple (and cost-free).

abizjak (changes requested):

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -427,19 +429,53 @@ movePrimitiveCursor target PrimitiveCursor{..} = do
...
+            Right kv -> return (Just (kv : acc))
+        step _ Nothing = return Nothing
+        prox :: Proxy db
+        prox = Proxy
+
+-- |Traverse a database, using a cursor internally, while a condition holds.  'MDB_val' values must
+-- not be retained as they will be reused.  The `(first, next)` tuple indicates the beginning and
+-- the direction of the traversal. Its possible values are `(CursorFirst, CursorNext)` and
+-- `(CursorLast, CursorPrevious)`.
+traverseTable' :: forall db a.  MDBDatabase db
```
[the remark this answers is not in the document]
mkmks: The original `traverseTable` is not used either and it also relied on the order of keys as returned by the table cursor. Would it make sense to delete the original altogether?

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
mkmks: The intention was not to specifically handle the genesis block state corruption but to indicate that the recovery attempt failed. I cleaned that up now.

The genesis block state recovery will be explicitly handled under the recovery from global state initialisation failure because this is the exact occasion when it can happen.

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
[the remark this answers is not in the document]
mkmks: I deleted the original `traverseTable` too because it's unused and its order of iteration was no clearer.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +693,44 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
     blocksSize = sum . map (\b -> 2*digestSize + fromIntegral (LBS.length . fst . fst $ b)) $ serializedBlocksTss
     tssSize = (* (2 * digestSize + 16)) . sum . map (Prelude.length . snd) $ blocktss
+
+-- If an explicitly finalized block does not satisfy a predicate, delete it, together with its
+-- finalization record, all blocks implicitly finalized by its finalization records and all
+-- associated transaction statuses. Returns the last finalized block hash available in the unrolled
+-- treestate/blockstate on success.
+unrollTreeStateWhile :: (MonadLogger m, MonadIO m, IsProtocolVersion pv, FixedSizeSerialization st) =>
+  DatabaseHandlers pv st -> (StoredBlock pv st -> IO Bool) -> m (Either String (FinalizationRecord, StoredBlock pv st))
+unrollTreeStateWhile dbh p = resizeOnFullInternal 4096 dbh $ \h -> transaction (h ^. storeEnv) False $ \txn ->
```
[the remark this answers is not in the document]
abizjak: Seeing that the predicate is effectful, please explain in documentation how `p` is used. I'd also recommend calling the `p` a more descriptive name, such as `shouldDelete`

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +693,44 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
           digestSize + 64 + digestSize + (32 * Prelude.length vs) + 48 + 64
     blocksSize = sum . map (\b -> 2*digestSize + fromIntegral (LBS.length . fst . fst $ b)) $ serializedBlocksTss
     tssSize = (* (2 * digestSize + 16)) . sum . map (Prelude.length . snd) $ blocktss
+
+-- If an explicitly finalized block does not satisfy a predicate, delete it, together with its
+-- finalization record, all blocks implicitly finalized by its finalization records and all
+-- associated transaction statuses. Returns the last finalized block hash available in the unrolled
+-- treestate/blockstate on success.
+unrollTreeStateWhile :: (MonadLogger m, MonadIO m, IsProtocolVersion pv, FixedSizeSerialization st) =>
```
abizjak, line 701: We have to be somewhat careful how we use the lmdb. We use a reader-writer lock to make sure that there is only the one active transactions when the db is being resized (see https://github.com/Concordium/concordium-node/pull/182 )

The way this function is currently used I think it's all good, since we are still in the setup stage. But please document the assumption here that nobody else is using the database while this function is working.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +693,44 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
+                Just sbParent | finIndex == sbFinalizationIndex sbParent -> loopBlocks finIndex sbParent
+                Just sbParent -> do
+                  delFinRecOK <- deleteRecord txn (h ^. finalizationRecordStore) finIndex
+                  if delFinRecOK then loopBlocks (sbFinalizationIndex sbParent) sbParent
+                    else return . Left $ "Could not delete the finalization record with index "
+                    <> show finIndex <> " finalizing block " <> show bh
+                Nothing -> return . Left $ "Could not read the parent of the block " <> show bh
+             else return . Left $ "Could not rollback block " <> show bh
+          else returnBlock finIndex sb
+      -- genesis blocks won't be deleted
```
abizjak, line 728: Perhaps rationale as to why?

To be clear,I think it is acceptable to not handle this specific case where the genesis state is corrupted since it is a bit more tricky to handle, but just insert a comment as to why we chose it.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -688,3 +693,44 @@ writeFinalizationComposite finRec blocktss = resizeOnFull (finRecSize + blocksSi
...
+unrollTreeStateWhile dbh p = resizeOnFullInternal 4096 dbh $ \h -> transaction (h ^. storeEnv) False $ \txn ->
+  let loopBlocks finIndex sb | NormalBlock block <- sbBlock sb =
+        do
+          let bh = _bpHash . sbInfo $ sb
+          isP <- p sb
+          if isP then do
+            delBlockOK <- deleteRecord txn (h ^. blockStore) bh
+            delBlockHashOK <- deleteRecord txn (h ^. finalizedByHeightStore) (_bpHeight . sbInfo $ sb)
+            delTxsOK <- mapM (deleteRecord txn (h ^. transactionStatusStore)) ((wmdHash <$>) . blockTransactions $ block)
+            if delBlockOK && delBlockHashOK && and delTxsOK then do
```
abizjak, line 712: What is the reasoning for failing if one of these keys does not exist? Is it because that should not happen?

Can you please document rationale.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
abizjak, line 701: Where do you use the MonadLogger here? I can't see any logging.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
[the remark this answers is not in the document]
abizjak: Would it also make more sense to do the deletion over multiple transactions? As it is now we are doing everything in one transaction, including using the predicate. Would it not be better to not try to do IO inside the transaction?

So in each DB tranasction we delete the section finalized by the last finalization record. Then check if we need to delete more. And delete the next batch in the transaction. This way also the unit we are deleting here corresponds exactly to the unit that is being written in a single transaction.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -378,20 +379,15 @@ loadSkovPersistentData rp _treeStateDirectory pbsc = do
...
-  case elfBlob :: Either IOError BS.ByteString of
-    Left _ -> logExceptionAndThrowTS $
-      DatabaseInvariantViolation "Last finalized block cannot be read. Blockstate database recovery required."
-    Right _ -> do
+  -- Unroll the treestate if the last finalized blockstate is corrupted. If the last finalized
+  -- blockstate is not corrupted, the treestate is unchanged.
+  unrollTreeStateWhile _db isBlockStateCorrupted >>= \case
+    Left e -> logExceptionAndThrowTS . DatabaseInvariantViolation $
+              "The block state database is corrupt. Recovery attempt failed: " <> e
+    Right (_lastFinalizationRecord, lfStoredBlock) -> do
```
abizjak, line 387: I can see that now you are not logging anything at all if you delete stuff from the database.

That is not OK and it will make support harder. It should be very easy to return a list of block hashes and finalization records that were deleted, so you could just log those here. If the list was empty then nothing would be logged.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -607,16 +609,19 @@ resizeOnFull :: (MonadLogger m, MonadIO m, MonadState s m, HasDatabaseHandlers p
   -> m a
 resizeOnFull addSize a = do
     dbh <- use dbHandlers
-    inner dbh
+    resizeOnFullInternal addSize dbh a
+
+resizeOnFullInternal :: (MonadIO m, MonadLogger m) => Int -> DatabaseHandlers pv st -> (DatabaseHandlers pv st -> IO b) -> m b
```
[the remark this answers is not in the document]
mkmks: There is a `resizeOnResizedInternal` function which is different from `resizeOnResized` in that it takes the database handlers as a parameter, instead of reading them from the monadic state.

Since we are performing recovery during Skov initialization, we're not working in a monad that `HasDatabaseHandlers`. Using `resizeOnFullInternal` here is similar to using `resizeOnResizedInternal` (through `openReadOnlyDatabase`) in the database exporter (which also doesn't rely on a Skov state).

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
mkmks: I was using it during debugging, and I think I'll still keep using it to log the deleted blocks and finalization records without having to return them to `loadSkovPersistentState`.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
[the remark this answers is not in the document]
abizjak: But can you please add documentation?

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
abizjak: Sounds reasonable.

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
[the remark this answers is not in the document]
mkmks: > What is the rationale here for the resizing? Seeing you are not adding values to the database, is it necessary?
> 
> (I don't think it harms having it, but I would like to understand why it is needed, if it is)

It turns out that it is necessary. Without resizing, I'm getting `MDB_MAP_FULL` on deletion. I found [a relevant LMDB issue](https://www.mail-archive.com/openldap-bugs@openldap.org/msg04006.html) which could be resolved by resizing the store by the size of one memory page.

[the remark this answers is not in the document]
mkmks: > Would it also make more sense to do the deletion over multiple transactions? As it is now we are doing everything in one transaction, including using the predicate. Would it not be better to not try to do IO inside the transaction?

Good points, will do.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -215,6 +215,24 @@ runBlobStoreTemp dir a = bracket openf closef usef
...
+    esize <- decode <$> BS.hGet bhHandle 8
+    case esize :: Either String Word64 of
+      Right size -> do
+        let newSize = offset + 8 + size
+        hSetFileSize bhHandle $ fromIntegral newSize
+        putMVar blobStoreFile bh{bhSize = fromIntegral newSize, bhAtEnd=False}
+      _ -> throwIO $ userError "Cannot truncate the blob store: cannot obtain the last blob size"
+  case eres :: Either SomeException () of
+    Left e -> throwIO e
+    Right () -> return ()
```
td202, line 234: The typical pattern that we use `mask` for is to ensure that we restore the `MVar` in the event of an exception, including an async exception. That is not happening in this case. I think it's probably reasonable to remove the `mask`ing in this case, because if there is some issue from `hSetFileSize` then we can't really expect to recover. (Alternatively, it might make sense to but an error value into the `MVar`, but I'm not sure that would be better.)

As a second point, I think the `blobStoreMMap` should be remapped on truncation. Otherwise, there could potentially be an issue of incorrect data being read.

td202 (commented):
This looks good. Just a few things to address.

abizjak (approved):

PR #486, opened 2022-09-08
gRPC api v2: Transaction Status

Bargsteen:
## Purpose

Add the transaction status endpoint to the new gRPC v2 API.
Depends on: https://github.com/Concordium/concordium-grpc-api/pull/13 and https://github.com/Concordium/concordium-base/pull/216

## Changes

- Add the transaction status endpoint.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -770,6 +1506,29 @@ getAncestorsV2 cptr channel blockType blockHashPtr depth outHash cbk = do
...
+getTransactionStatusV2 cptr trxHashPtr outVec copierCbk = do
+    Ext.ConsensusRunner mvr <- deRefStablePtr cptr
+    let copier = callCopyToVecCallback copierCbk
+    trxHash <- decodeTransactionHashInput trxHashPtr
+    res <- runMVR (Q.getTransactionStatus trxHash) mvr
+    case res of
+        Nothing -> do
+          return $ queryResultCode QRNotFound
+        Just ts -> case toTransactionStatus ts of
+              Left e -> Exception.throw e -- TODO: Review question: What should be returned if the if it is CEFailed* and not CEInvalid*?
```
limemloh, line 1526: You could extend `QueryResult` with the error and define an error code for it.
We probably also want to log a warning, since these are internal errors.

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -535,11 +563,714 @@ instance ToProto AccountInfo where
...
+    Updates.Level1KeysRootUpdate{..} -> Proto.make $ ProtoFields.level1KeysUpdate .= toProtoKeys l1kruKeys
+    Updates.Level2KeysRootUpdate{..} -> Proto.make $ ProtoFields.level2KeysUpdateV0 .= toProtoAuthorizationsV0 l2kruAuthorizations
+    Updates.Level2KeysRootUpdateV1{..} -> Proto.make $ ProtoFields.level2KeysUpdateV1 .= Proto.make (do
+        ProtoFields.v0 .= toProtoAuthorizationsV0 l2kruAuthorizationsV1
+        case Updates.asCooldownParameters l2kruAuthorizationsV1 of
+            JustForCPV1 as -> ProtoFields.parameterCooldown .= toProto as
+        case Updates.asTimeParameters l2kruAuthorizationsV1 of
+            JustForCPV1 as -> ProtoFields.parameterTime .= toProto as)
+
+-- |Is not needed if we can make a ToProto instance for HigherLevelKeys.
```
abizjak, line 855: I don't quite understand why we are not doing that? Is it because there is a type parameter to the higher-level keys?

abizjak: I recommend just having an instance.

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -535,11 +563,714 @@ instance ToProto AccountInfo where
...
+  type Output BakerAggregationVerifyKey = Proto.BakerAggregationVerifyKey
+  toProto = mkSerialize
+
+instance ToProto Memo where
+  type Output Memo = Proto.Memo
+  toProto = mkSerialize
+
+instance ToProto RegisteredData where
+  type Output RegisteredData = Proto.RegisteredData
+  toProto = mkSerialize
```
abizjak, line 950: Since registeredData is already a byte array, would we rather not just return that, rather than serialized with a length upfront?

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -535,11 +563,714 @@ instance ToProto AccountInfo where
...
+  type Output BakerElectionVerifyKey = Proto.BakerElectionVerifyKey
+  toProto = mkSerialize
+
+instance ToProto BakerAggregationVerifyKey where
+  type Output BakerAggregationVerifyKey = Proto.BakerAggregationVerifyKey
+  toProto = mkSerialize
+
+instance ToProto Memo where
+  type Output Memo = Proto.Memo
+  toProto = mkSerialize
```
abizjak, line 946: Since Memo is already a byte array, would we rather not just return that, rather than serialized with a length upfront?

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -535,11 +563,714 @@ instance ToProto AccountInfo where
...
+                    _ -> Left CEInvalidTransactionResult
+                Right . Proto.make $ ProtoFields.bakerAdded .= v
+            TTRemoveBaker -> mkSuccess <$> do
+                v <- case events of
+                    [BakerRemoved{..}] -> Right $ toProto ebrBakerId
+                    _ -> Left CEInvalidTransactionResult
+                Right . Proto.make $ ProtoFields.bakerRemoved .= v
+            TTUpdateBakerStake -> mkSuccess <$> do
+                v <- case events of
+                    [] -> Right Proto.defMessage -- TODO: Review question: Is it correct that if the stake is unchanged, no events occur?
```
abizjak, line 1042: Yes, it is unfortunately correct  and requires this special case. See line 1845 in the Scheduler.

concordium-consensus/src/Concordium/GRPC2.hs
abizjak: I think what Emil suggests is the correct approach. Return a special code here, and log an internal server error.
The response to the client is likewise an internal server error and "report bug".

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -535,11 +563,714 @@ instance ToProto AccountInfo where
...
+    ProtoFields.exponent .= fromIntegral mrExponent
+
+instance ToProto Wasm.WasmVersion where
+  type Output Wasm.WasmVersion = Proto.ContractVersion
+  toProto Wasm.V0 = Proto.V0
+  toProto Wasm.V1 = Proto.V1
+
+instance ToProto Wasm.ContractEvent where
+  type Output Wasm.ContractEvent = Proto.ContractEvent
+  toProto = mkSerialize
```
abizjak, line 906: I think this should be 
```
toProto (ContractEvent short) = Proto.make $ Proto.value .= BSS.toByteString short
```
since we don't want to be duplicating the length information.

abizjak (commented):
I have some comments that should be simple to resolve.

But overall looks very good. And better than what I expected actually. I expected there would be more code.

concordium-consensus/src/Concordium/GRPC2.hs
Bargsteen: I've extended the QueryResult now. I just logged with `hPutStrLn` which probably isn't ideal. Let me know what to use ;)

abizjak (approved):

PR #491, opened 2022-09-13
Revise module handling to pass module artifacts as bytes

EmilLa1:
## Purpose

Revise module handling so that module artifacts are passed between Rust and Haskell as byte strings.

## Changes

- Change `InstrumentedModuleV` to wrap a `ByteString` (via `ModuleArtifactBytes`).
- Change `processModule` to wrap the `ByteString`s returned by Rust as `InstrumentedModuleV`s. (The `ByteString` memory is allocated and freed on the Rust side, so it is given a finalizer on the Haskell side to manage this.)
- Change `applyInitFun` and `applyReceiveFun` to pass `ByteString`s to Rust.

Note: as before, module artifacts are not retained in memory (by the persistent state) once the block state is written to disk. `BlobPtr`s are used to read `ByteString`s directly from the blob store as needed. These `ByteString`s are direct references into the memory map (and not copies - at least in the typical case), so it is important that they are not retained unduly. The Rust side may make a copy of the artifact if it is needed for resuming contract execution.

## Checklist

      hard-to-understand areas.


td202 (approved):

concordium-consensus/src/Concordium/GlobalState/Wasm.hs
```diff
@@ -39,96 +32,32 @@ import Data.Serialize
...
--- garbage collected on the Haskell side.
-moduleArtifactFromBytes :: forall v. (IsWasmVersion v) => BS.ByteString -> IO (ModuleArtifact v)
-moduleArtifactFromBytes bs = do
-  let martifact = case getWasmVersion @v of
-        SV0 -> fromBytesHelper freeArtifactV0 fromBytesArtifactV0 bs
-        SV1 -> fromBytesHelper freeArtifactV1 fromBytesArtifactV1 bs
-  case martifact of
-    Nothing -> error "Cannot decode module artifact."
-    Just maArtifact -> return ModuleArtifact{..}
+-- |A processed module artifact as a 'BS.ByteString'.
```
abizjak, line 39: Please elaborate a bit more here about the provenance, and when this is meant to be used.

abizjak (approved):
Looks good. Please add a bit more documentation about the "module artifact" so there is more context for future maintainers.

PR #500, opened 2022-09-15
Expose contract state in GRPC query.

### LINKED ISSUE #489: Expose entire contract state via a GRPC2 query
**Task description**

Add an endpoint to return the entire contract state. For V1 contracts the endpoint should stream a list of (key, value) pairs.

For V0 the stream will consist of a single value, with empty key. The version information should be in the metadata.

This feature was requested by some users in order to be able to simulate contract operations off-chain.

abizjak:
## Purpose

Closes #489 

## Changes

Depends on 
- https://github.com/Concordium/concordium-base/pull/219
- https://github.com/Concordium/concordium-grpc-api/pull/15

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -283,13 +283,25 @@ type family UpdatableContractState (v :: Wasm.WasmVersion) = ty | ty -> v where
   UpdatableContractState GSWasm.V0 = Wasm.ContractState
   UpdatableContractState GSWasm.V1 = StateV1.MutableState
 
+-- |An external representation of the persistent (i.e., frozen) contract state.
+-- This is used to pass this state through FFI for queries and should not be
+-- used for other purposes.
```
Bargsteen, line 288: Consider adding a brief explanation of _why_ it shouldn't be used for other purposes.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -664,6 +679,38 @@ extern "C" {
...
+        consensus: *mut consensus_runner,
+        block_id_type: u8,
+        block_id: *const u8,
+        contract_address_index: u64,
+        contract_address_subindex: u64,
+        out_hash: *mut u8,
+        out_v0: *mut Vec<u8>,
+        copier_v0: CopyToVecCallback,
+        out_v0: *mut V1ContractStateReceiver,
+        copier_v0: CopyV1ContractStateCallback,
```
Bargsteen, line 711:
```suggestion
        out_v1: *mut V1ContractStateReceiver,
        copier_v1: CopyV1ContractStateCallback,
```

concordium-node/src/grpc2.rs
```diff
@@ -610,6 +617,102 @@ pub mod server {
...
+                    let (sender, receiver) = tokio::sync::mpsc::channel(1);
+                    let _sender = tokio::spawn(async move {
+                        let msg = types::InstanceStateKvPair {
+                            key:   Vec::new(),
+                            value: state,
+                        };
+                        let _ = sender.send(Ok(msg)).await;
+                        // The error only happens if the receiver has been
+                        // dropped already (e.g., connection closed),
+                        // so we do not have to handle it. We just stop
```
Bargsteen, line 644: `stop.`? Or is there a line missing?

concordium-node/src/grpc2.rs
```diff
@@ -610,6 +617,102 @@ pub mod server {
...
+                    let mut response =
+                        tonic::Response::new(tokio_stream::wrappers::ReceiverStream::new(receiver));
+                    add_hash(&mut response, hash)?;
+                    Ok(response)
+                }
+                ContractStateResponse::V1 {
+                    state,
+                    mut loader,
+                } => {
+                    let (sender, receiver) = tokio::sync::mpsc::channel(10);
```
Bargsteen, line 655: Why was 10 chosen?

Bargsteen (approved):
Looks good in general 😊 👍

concordium-consensus/src/Concordium/GRPC2.hs
```diff
@@ -1574,6 +1575,59 @@ getInstanceInfoV2 cptr blockType blockHashPtr addrIndex addrSubindex outHash out
...
+    -- |Smart contract instance subindex.
+    Word64 ->
+    -- |Out pointer for writing the block hash that was used.
+    Ptr Word8 ->
+    -- |If the instance is a V0 instance its instance state is written to this vector.
+    -- Otherwise this is unused.
+    Ptr ReceiverVec ->
+    -- |Callback to output data in case it is a V0 instance.
+    FunPtr CopyToVecCallback ->
+    -- |If the instance is a V1 instance its instance state and callbacks are written to this vector.
```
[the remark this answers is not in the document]
abizjak: I suppose we could do that. But then we can't use the CopyToVecCallback since the types don't match.

We could just have the one callback with an additional argument and use the null pointers to signal which version we are writing. but I think that is more complicated than this.

concordium-node/src/grpc2.rs
abizjak: It is `stop.` :)

concordium-node/src/grpc2.rs
abizjak: I added a comment now. Because it's a big small number and provides some buffer, but not too much so that we'd have to keep too much data in memory in case the consumer is slow.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
abizjak: I tried to do that.

concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -605,6 +606,25 @@ getInstanceInfo bhi caddr = do
               Wasm.iiSourceModule = GSWasm.miModuleRef (instanceModuleInterface iiParameters)
               }))
 
+
+-- |Get the details of a smart contract instance in the block state.
```
limemloh, line 610: This description needs to be updated

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1596,6 +1643,57 @@ impl ConsensusContainer {
...
+        let (block_id_type, block_hash) =
+            crate::grpc2::types::block_hash_input_to_ffi(block_hash).require()?;
+        let addr_index = address.index;
+        let addr_subindex = address.subindex;
+        let consensus = self.consensus.load(Ordering::SeqCst);
+        let mut out_v0_data: Vec<u8> = Vec::new();
+        let mut out_v1_data = V1ContractStateReceiver {
+            state:  None,
+            loader: None,
+        };
```
limemloh, line 1662: Would it makes sense to lift the Option to the entire struct? 
It would remove some of the cases when matching.
If not, maybe add a comment on why at the definition of the struct.

limemloh (approved):
Looks great, only minor comments

concordium-node/src/consensus_ffi/ffi.rs
abizjak: That also works. Changed it.

Issue #418, opened 2022-06-23
Windows installer licenses are out of date.


abizjak:
**Bug Description**

The licenses included in the windows installer are out of date. New dependencies have been added, some have been removed.

PR #502, opened 2022-09-16
Update the license file for the windows installer.

td202:
## Purpose

Addresses #418

Update the license file that ships with the Windows installer to account for changes in dependencies.

## Changes

License file was regenerated using the license tool based on the components that ship with the Windows installer.

## Checklist

      hard-to-understand areas.


abizjak:
Could we not instead remove these files? What exactly is needed with the installer?

And if these really **must** be shipped with the installer, then they should be generated at the time the installer is created. Is that not possible?

This kind of PR is impossible to review and having these files in is just a lot of extra data in the repository, which makes every operation with the repository just a bit slower.

EmilLa1 (approved):
GJ.
While I agree with @abizjak it is best to have this being produced as part of the build step, I can also accept the solution at hand if we want to have the licenses upgraded for the upcoming release. Thus postponing the former mentioned task to next Q is also fine for me.

abizjak:
These licenses have not been updated since they were generated. And the main reason is that it is impractical to do so, nor is it documented that this is part of the release process. I should have cared more about this when these were first added, but I failed.

The licenses really must be removed from the repository.

It is the same in the documentation repository, where we have a copy of these licenses, probably in a different version than here because they have probably been updated at some other point in time, and maybe they also cover other software.

PR #504, opened 2022-09-17
Fix space leak in nonce computation.

abizjak:
## Purpose

Previously starting the chain with a year old genesis would lead to 1.7GB of memory use that did not go away for a long time. Now the memory is <100MB with the same genesis.

Most of the memory that is leakedin that example is actually retained in foreign code, in modules and in contract state.

This also reduces execution time on the specific example from ~900s to ~600s. It must be that GC time is less, or that there are fewer finalizers for foreign objects that need to be maintained that causes this improvement.

## Changes

Update with epoch is strict now.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
Good catch!

PR #506, opened 2022-09-19
Fix typo in environment variable

bisgardo:
## Purpose

Spell cewrectly.

## Changes

Fix typo in environment variable `CONCORDIUM_NODE_PROMETHEUS_LISTEN_ADDRESSS` (remove trailing `S`).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix typo in environment variable `CONCORDIUM_NODE_PROMETHEUS_LISTEN_ADDRESSS` (remove trailing `S`).
+
 ## 4.4.2
 
 - Speed up and reduce memory overhead during protocol updates.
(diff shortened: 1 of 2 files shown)


EmilLa1 (approved):

abizjak:
What is your position about this being both a typo fix, but also a breaking change?

EmilLa1:
1. I think the typo fix is good to solve.
2. Ideally I would also think we should wait and merge this for an appropriate time where we have more emerging breaking changes as well.

abizjak:
> 1. I think the typo fix is good to solve.
> 
>     2. Ideally I would also think we should wait and merge this for an appropriate time where we have more emerging breaking changes as well.

I just wanted to bring up the fact that this is a breaking change technically.

However since we don't advertise, nor is it documented, the fact that we have the prometheus interface, nor is this interface enabled in any of our node distributions, we can just merge this.

abizjak (approved):

PR #509, opened 2022-09-22
Ubuntu pipeline

Thahara:
## Purpose

Add Jenkins pipeline for building the Ubuntu distribution of the node.

## Changes

Added Jenkinsfile. 

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/jenkinsfiles/ubuntu.Jenkinsfile
+++ b/jenkinsfiles/ubuntu.Jenkinsfile
@@ -42,6 +42,7 @@ pipeline {
         DATA_DIR = "./scripts/distribution/ubuntu-packages/template/data/"
         RPC_PORT = "${rpc_port[environment]}"
         LISTEN_PORT = "${listen_port[environment]}"
+        STATIC_BINARIES_IMAGE_TAG = "${BUILD_TAG}"
     }
     stages {
         stage('Precheck') {
@@ -83,6 +84,7 @@ pipeline {
                 sh '''
                     docker build\
                         --build-arg ubuntu_version=${UBUNTU_VERSION}\
+                        --build-arg static_binaries_image_tag=${STATIC_BINARIES_IMAGE_TAG}\
                         --build-arg build_env_name=${ENVIRONMENT_CAP}\
                         --build-arg build_env_name_lower=${ENVIRONMENT}\
                         --build-arg build_genesis_hash=$(cat ${GENESIS_HASH_PATH} | tr -cd "[:alnum:]")\
--- a/jenkinsfiles/ubuntu.Jenkinsfile
+++ b/jenkinsfiles/ubuntu.Jenkinsfile
@@ -7,6 +7,18 @@
 
 @Library('concordium-pipelines') _
 
+Map rpc_port = [
+            mainnet: "10000",
+            testnet: "10001",
+            stagenet: "10500"
+        ]
+
+Map listen_port = [
+            mainnet: "8888",
+            testnet: "8889",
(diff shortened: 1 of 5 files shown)


jenkinsfiles/ubuntu.Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+        stage('Publish') {
+            steps {
+                // Copy out the build artifacts. We create a temporary container and use docker
+                // cp. This makes the output artifacts have correct file permissions (they are
+                // owned by the user who ran the script).
+                sh '''
+                    id=$(docker create ${ENVIRONMENT}-deb)
+                    docker cp $id:/out ${ENVIRONMENT}-build
+                    docker rm $id
+                    aws s3 cp ${ENVIRONMENT}-build/${BUILD_FILE} ${OUTFILE} --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers
```
abizjak, line 108: In some other scripts we check for existence of the file and fail early if the file with the given name already exists. Would this make sense here as well?

jenkinsfiles/ubuntu.Jenkinsfile
```diff
@@ -0,0 +1,113 @@
...
+                    // Build it if forced_rebuild is checked
+                    expression {
+                        return params.FORCE_REBUILD
+                    }
+
+                    // OR
+                    // Check if static-node-binaries image is present, if not build it.
+                    equals expected: "1",
+                    actual: "${sh script:'docker inspect --type=image static-node-binaries > /dev/null 2> /dev/null', returnStatus:true}"
+                }
```
abizjak, line 60: Seeing that this is a build script, will it ever make sense not to rebuild the static libraries? 

In the shell script that I have in the repository I added this because it makes it faster to build stagenet/testnet/mainnet versions at the same time, since it only has to build the static libraries once. However it is a bit tricky to use correctly since you have to remember to use the right version of the image :)

My guess would be that here we always want to rebuild, so I would simplify this and remove the "FORCE_REBUILD". Alternatively if reuse is useful, I would tag the image with the commit hash, not just `static-node-binaries`.

abizjak (approved):
Looks good in general.

I would recommend we always fully build though, since that is less error prone.

jenkinsfiles/ubuntu.Jenkinsfile
Thahara: Yes it would, added the check.

jenkinsfiles/ubuntu.Jenkinsfile
Thahara: Yeah, it has very limited usage, as the build agent, and any docker images on it, is killed after 10 minutes of inactivity. So I'll remove it to avoid any potential confusion.

abizjak: Will this be a problem if, e.g., I start two builds at the same time (one stagenet, one mainnet)? The fact that the intermediate image is tagged with the same name?

Thahara: yeah it probably could, I added a new param to the build scripts/dockerfile `STATIC_BINARIES_IMAGE_TAG`, in the "manual" builds for test-/mainnet it defaults to latest, and in jenkins it is set to `BUILD_TAG`, which is of the format `jenkins-concordium-node-ubuntu-10`, so that ties it to the specific jenkins build, meaning it should basically never be reused.

PR #510, opened 2022-09-22
Fix check for database corruption.

abizjak:
## Purpose

Sometimes when the node crashes or is killed the blockstate.dat ends up of the correct size, but with trailing zeros.
This PR fixes the check for "database is good" by attempting to load the block state at the given location.

## Changes

- Make the check for "is database corrupted" more strict.
- Bump version number for next release.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
@@ -345,8 +345,16 @@ readBlobBS bs@BlobStoreAccess{..} br@(BlobRef offset) = do
                             $ BS.take (fromIntegral size)
                             $ BS.drop dataOffset mmap
 
--- |Check if a given 'BlobRef' is valid in the blob store. This does not attempt to deserialize
--- the value, but will check that the offset and length of the 'BlobRef' are valid.
+-- |Check if a given 'BlobRef' is valid in the blob store. This attempts to
+-- load the value from the given location and handles the following two failures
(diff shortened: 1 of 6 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -347,8 +347,8 @@ readBlobBS bs@BlobStoreAccess{..} br@(BlobRef offset) = do
 
 -- |Check if a given 'BlobRef' is valid in the blob store. This does not attempt to deserialize
```
mkmks, line 348: I guess, this comment is not accurate after having changed `readBlobBS` to `loadRef`. The former does not deserialize the blob, but the latter does.

mkmks (approved):
Just out of curiosity, why would a check based on `readBlobBS` fail where a check based on `loadRef` succeed? The difference between the two is deserialization. Are we expecting that it would not be possible to deserialize non-truncated blobs?

EmilLa1 (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
abizjak: I fixed the comment.

abizjak:
> Just out of curiosity, why would a check based on `readBlobBS` fail where a check based on `loadRef` succeed? The difference between the two is deserialization. Are we expecting that it would not be possible to deserialize non-truncated blobs?

I tried to explain in the PR description. The blobstore ends with enough trailing zeros.

PR #512, opened 2022-09-23
Fix migration of modules.

### LINKED ISSUE #511: Node crashes when processing block
**Bug description**
On mainnet
```
2022-09-22T20:50:09.240123723Z: DEBUG: Skov: Blocks d8f22d7752b182a7e66c39decf2385dd74660feef3e793b95115b6ace1f0e2ed marked finalized
2022-09-22T20:50:09.240195206Z: DEBUG: Skov: Processed finalization in 0.003993748s
2022-09-22T20:50:09.240281124Z: DEBUG: Afgjort: Starting passive finalization round: height=311256 delta=0
concordium-node: Cannot decode module artifact.
CallStack (from HasCallStack):
  error, called at src/Concordium/GlobalState/Wasm.hs:95:16 in concordium-consensus-0.1.0.0-FfYjXKiH7BB1kKTGkCZsbY:Concordium.GlobalState.Wasm

�PV: warning: too many hs_exit()s
```

The node was started from scratch.

**Versions**
- Software Version: 4.4.2

abizjak:
## Purpose

Fixes #511 

## Changes

Migrate inner references when migrating modules.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
@@ -199,6 +199,8 @@ instance MonadBlobStore m => DirectBlobStorable m Module where
         ModuleV0 mv0 -> sudV SV0 mv0
         ModuleV1 mv1 -> sudV SV1 mv1
     where
+      -- NB: If this is updated the @storeV@ function in 'migrateModules' below
+      -- must also be updated if any changes in the representation are made.
       sudV :: SWasmVersion v -> ModuleV v -> m (BlobRef Module, Module)
       sudV ver ModuleV{moduleVInterface = GSWasm.ModuleInterface{..}, ..} = do
           !instrumentedModuleBytes <- case miModule of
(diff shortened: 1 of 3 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
```diff
@@ -398,19 +398,69 @@ migrateModules ::
...
+            let !miModule' = PIMVPtr BlobPtr {
+                    -- Pointer is blob ref + 8 bytes (length of blob) + header length +
+                    -- 4 bytes for version + 4 bytes for length of instrumented module
+                    theBlobPtr = theBlobRef br + 8 + headerLen + 8,
+                    -- Length is the length of the serialized instrumented module -
+                    -- 4 bytes for version - 4 bytes for length
+                    blobPtrLen =  imLen - 8
+                  }
+            let mv' = ModuleV{moduleVInterface = GSWasm.ModuleInterface{miModule = miModule',..}, moduleVSource = newSource,..}
+            return $! mkModule ver mv'
```
td202, line 463: Both `sudV` and `refFlush` will end up writing the module to the blob store, which is redundant.  Perhaps something like the following would do:

```suggestion
      newModuleVInterface <- lift $ mapM (fmap PIMVMem . loadInstrumentedModuleV) moduleVInterface
      (newModuleBlobRef, newModule) <- storeUpdateDirect $
          mkModule (getWasmVersion @v) $ ModuleV {
            moduleVInterface = newModuleVInterface,
            moduleVSource = newModuleVSource
          }
      let !h :: Hash = getHash newModule
      return $! HCRFlushed newModuleBlobRef h
```

For a cleaner abstraction, the `storeUpdateDirect`, `getHash(M)` and `HCRFlushed` could be moved to a `makeFlushedHashedCachedRef`.

td202 (changes requested):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
abizjak: Yes, I realized that. I do not like deserializing the artifact which is what `loadInstrumentedModuleV` does, so there is a bit of duplication with `storeDirect/loadDirect` in the current version.

but I added documentation to all these functions so that they will hopefully be updated at the same time.

EmilLa1 (approved):
Looks good!

I would probably factor out the bits and pieces that are in common with the loadDirect instance decl. But I'm fine either way.

abizjak:
> Looks good!
> 
> I would probably factor out the bits and pieces that are in common with the loadDirect instance decl. But I'm fine either way.

I factor that out now. I don't think it makes it better. The function that is factored out has awkward semantics. It is just internal, but still.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
td202: `loadInstrumentedModuleV` no longer deserializes the artifact.

abizjak: It does on the branch that this PR is based on, i.e., main.

abizjak:
```
-- |Load an instrumented module from a 'BS.ByteString'. This creates a reference counted copy of
-- the artifact on the Rust side and attaches a finalizer to drop the reference when it is
-- garbage collected on the Haskell side.
instrumentedModuleVFromBytes :: forall v. (IsWasmVersion v) => BS.ByteString -> IO (InstrumentedModuleV v)
instrumentedModuleVFromBytes bs = do
    artifact <- moduleArtifactFromBytes @v bs
    case getWasmVersion @v of
      SV0 -> return (InstrumentedWasmModuleV0 artifact)
      SV1 -> return (InstrumentedWasmModuleV1 artifact)
```

abizjak:
I can wait with merging this to main after the refactoring FFI thing is merged, to make the migration simpler, but I am not putting that in the next release since only bugfixes are meant to be merged there.

EmilLa1:
@abizjak I believe this is ready for a merge now as https://github.com/Concordium/concordium-node/pull/491 is merged into `main`.

abizjak:
I need to update it to the simpler solution that is now possible. Will try to do that today.

abizjak:
I merged in main now, and made the changes discussed above.

concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
```diff
@@ -418,10 +418,20 @@ instance
 
 
 -- |Construct a 'HashedCachedRef'' given the value and hash.
+-- The value is in memory, and is __not__ stored to disk.
 makeHashedCachedRef :: (MonadIO m) => a -> h -> m (HashedCachedRef' h c a)
 makeHashedCachedRef val hsh = liftIO $
     HCRUnflushed <$!> (newIORef $! HCRMemHashed val hsh)
 
+-- |Construct a 'HashedCachedRef'' given the value. THe value is hashed and then
```
EmilLa1, line 426:
```suggestion
-- |Construct a 'HashedCachedRef'' given the value. The value is hashed and then
```

EmilLa1 (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
```diff
@@ -390,24 +390,47 @@ putModulesV0 mods = do
...
+    where 
+      migrateCachedModule :: CachedModule -> t m CachedModule
+      migrateCachedModule cm = do
+        existingModule <- lift (refLoad cm)
+        case existingModule of
+          ModuleV0 v0 -> migrateModuleV v0
+          ModuleV1 v1 -> migrateModuleV v1
+
+      migrateModuleV :: forall v . (IsWasmVersion v) => ModuleV v -> t m CachedModule
+      migrateModuleV ModuleV{..} = do
```
td202, line 414: It might be a little simpler to remove `mkModule` and change as follows:
```suggestion
          ModuleV0 v0 -> migrateModuleV ModuleV0 v0
          ModuleV1 v1 -> migrateModuleV ModuleV1 v1

      migrateModuleV :: (ModuleV v -> Module) -> ModuleV v -> t m CachedModule
      migrateModuleV mkModule ModuleV{..} = do
```

td202 (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Modules.hs
abizjak: Not sure. I dislike the need for this embedding as well, but having to provide a callback is to me somewhat less clear than the current situation. So I prefer to keep it as is.

PR #514, opened 2022-09-26
Dead block cache test failure

td202:
## Purpose

A test on the dead block cache can fail because the assumption that the last-inserted 1000 blocks are in the cache is only guaranteed if none of them were already in the cache.

## Changes

- Remove the 1000 block check from the existing test.
- Add a test with the check where the generator guarantees distinctness of block hashes.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
lgtm

abizjak (approved):

Issue #515, opened 2022-09-26
Truncating blob store fails on Windows


td202:
**Bug Description**

When shrinking the block state database on start-up, if a non-trivial change is made then the node hangs on Windows. This is because Windows [does not support resizing a file that is memory-mapped](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setendoffile#remarks). The exception handling expects that the blob store will be left in a state where it can be closed, but that is not guaranteed by the truncation code.

**Steps to Reproduce**

Run the node on Windows from an existing state where the block state database file is longer than would be expected. (This can be achieved by e.g. appending bytes to a valid database.)

**Expected Result**

The node should successfully start up, truncating the database as necessary.

**Actual Result**

The node hangs.

**Versions**
- Software Version: 4.4.3
- OS: Windows

PR #516, opened 2022-09-26
Fix for block state truncation issue on Windows.

td202:
## Purpose

Addresses #515

This fixes a problem with block state truncation failing on Windows.  This issue is caused by attempting to truncate a file that is currently memory mapped.  As releasing the memory maps is dependent on the garbage collector, this revises the process so that truncation is performed before the file is ever memory mapped.

## Changes

- Introduce a `ProvisionalBlobStore` that supports loading, but not storing, and does not make use of the memory map.
- Use the `ProvisionalBlobStore` while testing block states for validity.
- Redefine truncation only on `ProvisionalBlobStore`.
- Allow upgrading a `ProvisionalBlobStore` to a regular `BlobStore` once the truncation is completed.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState.hs
@@ -32,7 +32,7 @@ import qualified Concordium.GlobalState.Basic.TreeState as Basic
 import Concordium.GlobalState.BlockMonads
 import Concordium.GlobalState.BlockState
 import Concordium.GlobalState.Parameters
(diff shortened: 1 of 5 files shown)


abizjak (commented):
This is rather complicated, but if necessary then it is necessary.

Have you considered a simpler solution with unmapping the file explicitly before remapping, e.g., https://github.com/Concordium/concordium-node/compare/main...unmap-before-truncating 

Alternatively with inserting `performGC` instead of `finalizeForeignPtr`, that should also work.

This seems like it should also work the way we use the blob store, and seems like a simpler solution.

EmilLa1 (approved):
I agree with @abizjak that this seems rather complex to solve an issue on some specific platform. 

However I find it manageable and as @abizjak says if it's necessary then it's necessary.

PR #517, opened 2022-09-27
Unmap before truncating

td202:
## Purpose

Addresses #515.

This unmaps the blob store memory map before truncating the blob store, as truncating the file while it is memory mapped will fail on Windows.

## Changes

- Unmap the memory map in `truncateBlobStore`.
- Load the genesis block after truncation, to reduce the risk of any residual bad pointers. (It is unlikely that loading the genesis block earlier would actually retain bad pointers, but this mitigates the risk from future changes to block storage.)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased changes
 
-- Fix a bug in database recovery where truncating the block state database would hang on Windows.
+- Fix a bug in database recovery where the node would hang when truncating the block state database
+  on Windows.
 
 ## 4.4.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix a bug in database recovery where truncating the block state database would hang on Windows.
+
 ## 4.4.3
 
 - Fix a bug in database recovery where corruption was not always correctly detected.
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
@@ -284,8 +284,8 @@ runBlobStoreTemp dir a = bracket openf closef usef
 -- | Truncate the blob store after the blob stored at the given offset. The blob should not be
 -- corrupted (i.e., its size header should be readable, and its size should match the size header).
 --
--- **Note**: This function must not be made after any writes to the blob store. It assumes that the
--- current 'blobStoreMMap' is the only mapping of the file (which can be relied on if no writes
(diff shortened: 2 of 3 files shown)


abizjak (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix a bug in database recovery where truncating the block state database would hang on Windows.
```
EmilLa1, line 5: Maybe a bit more context as to what would `hang`.
```suggestion
- Fix a bug in database recovery where truncating the block state database would hang the node process on Windows.
```

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -281,6 +283,11 @@ runBlobStoreTemp dir a = bracket openf closef usef
 
 -- | Truncate the blob store after the blob stored at the given offset. The blob should not be
 -- corrupted (i.e., its size header should be readable, and its size should match the size header).
+--
+-- **Note**: This function must not be made after any writes to the blob store. It assumes that the
```
EmilLa1, line 287:
```suggestion
-- **Note**: This function must not be invoked after any writes to the blob store. It assumes that the
```

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -290,12 +297,25 @@ truncateBlobStore BlobStoreAccess{..} (BlobRef offset) = do
...
-        hSetFileSize bhHandle $ fromIntegral newSize
+        unless (bhSize == fromIntegral newSize) $ do
+          -- unmap the current memory mapped file since on some platforms the file cannot be
+          -- truncated if it is memory mapped.
+          oldMmap <- readIORef blobStoreMMap
+          -- write a temporary empty string.
+          writeIORef blobStoreMMap BS.empty
+          let (fp, _, _) = BSInternal.toForeignPtr oldMmap
+          -- unmap the old mapping.
+          finalizeForeignPtr fp
```
EmilLa1, line 308: A matter of taste and no deal breaker, but I think it could be nice to have this factored out in a separate `where` func decl. `unmap` in order to make keep the abstraction level the same throughout the internals of the function.

EmilLa1 (approved):
Looks good!  A few remarks you can do with as you want :-)

mkmks (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
td202: Fair comment, but I think I prefer the way it is.

PR #531, opened 2022-09-30
Node queries

EmilLa1:
## Purpose

- NodeInfo
- PeersInfo
- Network dumps
- Banning
- Peer connect/disconnect
- Shutdown

## Changes

Introduced types and endpoints.

## Checklist

      hard-to-understand areas.


concordium-node/src/grpc2.rs
```diff
@@ -264,7 +265,10 @@ pub mod server {
     use std::sync::{Arc, Mutex};
     use tonic::{async_trait, transport::ServerTlsConfig};
 
-    use super::*;
+    use super::{
+        types::node_info::{self, baker_consensus_info},
```
abizjak, line 269: Probably just qualify these. That's what we do with others.

concordium-node/src/grpc2.rs
```diff
@@ -985,6 +988,132 @@ pub mod server {
...
             Ok(response)
         }
+
+        async fn get_node_info(
+            &self,
+            _request: tonic::Request<crate::grpc2::types::Empty>,
+        ) -> Result<tonic::Response<types::NodeInfo>, tonic::Status> {
+            let peer_version = self.node.get_version();
+            let node_type = match self.node.peer_type() {
+                crate::common::PeerType::Bootstrapper => node_info::NodeType::Bootstrapper as i32,
```
abizjak, line 998: Isn't there an `into` instead?  `as` is  not great for maintainability.

concordium-node/src/grpc2.rs
```diff
@@ -1367,6 +1370,374 @@ pub mod server {
...
+                Err(tonic::Status::internal("Could not load banned peers."))
+            }
+        }
+
+        async fn ban_peer(
+            &self,
+            request: tonic::Request<crate::grpc2::types::PeerToBan>,
+        ) -> Result<tonic::Response<crate::grpc2::types::Empty>, tonic::Status> {
+            let ip = request.into_inner().ip_address.require()?;
+            match <std::net::IpAddr as std::str::FromStr>::from_str(&ip.value) {
```
abizjak, line 1465: What is the reason for this rather verbose style? Why can't we use

```
ip.value.parse::<std::net::IpAddr>()
```

concordium-node/src/grpc2.rs
```diff
@@ -1367,6 +1370,374 @@ pub mod server {
...
+                    e
+                ))),
+            }
+        }
+
+        async fn unban_peer(
+            &self,
+            request: tonic::Request<crate::grpc2::types::BannedPeer>,
+        ) -> Result<tonic::Response<crate::grpc2::types::Empty>, tonic::Status> {
+            match <std::net::IpAddr as std::str::FromStr>::from_str(
```
abizjak, line 1481: Likewise here

concordium-node/src/grpc2.rs
```diff
@@ -1367,6 +1370,374 @@ pub mod server {
...
+            }
+        }
+
+        #[cfg(feature = "network_dump")]
+        async fn dump_start(
+            &self,
+            request: tonic::Request<crate::grpc2::types::DumpRequest>,
+        ) -> Result<tonic::Response<crate::grpc2::types::Empty>, tonic::Status> {
+            let file_path = request.get_ref().file.to_owned();
+            // todo: fail with invalid argument if file path was empty.
```
abizjak, line 1505: TODO, please handle as described.

concordium-node/src/grpc2.rs
```diff
@@ -1367,6 +1370,374 @@ pub mod server {
...
+                .activate_dump(
+                    if file_path.is_empty() {
+                        "dump"
+                    } else {
+                        &file_path
+                    },
+                    request.get_ref().raw,
+                )
+                .is_ok();
+            Ok(tonic::Response::new(crate::grpc2::types::Empty {}))
```
abizjak, line 1517: Would it make sense to be more defensive here and only allow relative paths inside, e.g., a data directory?

abizjak (approved):
This looks good to me, modulo the changes that are going to be needed as a result of the changes in the types.proto definition.

concordium-node/src/grpc2.rs
EmilLa1: I'm not sure. We assume the operator has exclusive access to his node anyhow - why not let him dump the file where ever he wants to? 
On the other hand it also sounds fine to somehow sanitize the path. 

I'll let that decision be up to the reviewer :) ...

abizjak: I guess as long as it doesn't crash the node it should be fine. So let's keep it simple.

PR #533, opened 2022-10-01
GRPC2 network dumps

EmilLa1:
## Purpose

Expose start and stop network dump in the new grpc API. 

## Changes

## Checklist

      hard-to-understand areas.


concordium-node/build.rs
```diff
@@ -455,6 +455,24 @@ fn build_grpc2(proto_root_input: &str) -> std::io::Result<()> {
                 .server_streaming()
                 .build(),
         )
+        .method(
+            tonic_build::manual::Method::builder()
+                .name("dump_start")
+                .route_name("DumpStart")
+                .input_type("crate::grpc2::types::DumpRequest")
+                .output_type("crate::grpc2::types::BoolResponse")
```
abizjak, line 463: Make sure to change this as well.

concordium-node/src/grpc2.rs
```diff
@@ -985,6 +985,54 @@ pub mod server {
...
+                .node
+                .activate_dump(
+                    if file_path.is_empty() {
+                        "dump"
+                    } else {
+                        &file_path
+                    },
+                    request.get_ref().raw,
+                )
+                .is_ok();
```
abizjak, line 1005: I recommend that you just return "internal server error" here if you cannot start the dump.

If the string is not a valid path then return `invallid_argument`.

Issue #490, opened 2022-09-13
Remove redundant `accountRegIds`


td202:
**Task description**

The (persistent) accounts table is currently defined as follows:

```
data Accounts (pv :: ProtocolVersion) = Accounts {
    -- |Unique index of accounts by 'AccountAddress'
    accountMap :: !(AccountMap.PersistentAccountMap pv),
    -- |Hashed Merkle-tree of the accounts
    accountTable :: !(LFMBTree' AccountIndex HashedBufferedRef (HashedCachedRef (AccountCache (AccountVersionFor pv)) (PersistentAccount (AccountVersionFor pv)))),
    -- |Optional cached set of used 'ID.CredentialRegistrationID's
    accountRegIds :: !(Nullable (Map.Map ID.RawCredentialRegistrationID AccountIndex)),
    -- |Persisted representation of the map from registration ids to account indices.
    accountRegIdHistory :: !(Trie.TrieN BufferedFix ID.RawCredentialRegistrationID AccountIndex)
}
```

`accountRegIds` should be removed as `accountRegIdHistory` can be used in all cases instead.

abizjak:
Additionally, we should try not to cache the regIdHistory to lessen the memory load. The conjecture is that it should perform well enough. This will also speed up startup a bit.

PR #536, opened 2022-10-03
Remove memory credential map

td202:
## Purpose

Addresses #490.

This removes the in-memory map from credential registration IDs to account indexes. The persisted map is modified to be always on disk only.

Based on benchmarking, look-ups in the persisted map are ~1000 times slower than having the map in memory (around 100 microseconds). But the context in which this is used (primarily in credential deployments) this cost is acceptable.

## Changes

- Introduce `UnbufferedRef` reference type that is not cached in memory after it is flushed to the blob store.
- Introduce `UnbufferedFix`, a fixpoint combinator analogous to `BufferedFix` but using `UnbufferedRef` instead of `BufferedRef`.
- Support for migrating Tries that use `UnbufferedFix`.
- Remove the `accountRegIds` map, and handle all queries and additions in the Trie only.
- Add a benchmark for comparing the performance of querying by credential registration IDs with different indexes.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/benchmarks/credential-lookup/CredentialLookup.hs
+++ b/concordium-consensus/benchmarks/credential-lookup/CredentialLookup.hs
@@ -50,7 +50,9 @@ genRegID :: Int -> RawCredentialRegistrationID
 genRegID seed = RawCredentialRegistrationID $! FBS.pack (randoms (mkStdGen seed))
 
 -- |Create a benchmark environment by populating the map and tries with the given number of
--- generated credentials.
+-- generated credentials. The trie is written out after each insertion. This decreases the locality
+-- of the trie in the blob store and gives a more realistic scenario than if the trie is only
+-- written a single time.
 setup :: Int -> IO Setup
 setup size = do
     tempBlobStoreFile <- emptySystemTempFile "blb.dat"
--- a/concordium-consensus/benchmarks/credential-lookup/CredentialLookup.hs
+++ b/concordium-consensus/benchmarks/credential-lookup/CredentialLookup.hs
@@ -1,0 +1,138 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- |A benchmark for credential lookup. This compares three approaches:
+--
+-- 1. A map stored in memory.
+--
+-- 2. A trie that is cached in the memory.
+--
+-- 3. A trie that is not cached in the memory and always accessed from the disk.
+module Main where
+
+import Control.DeepSeq
+import Control.Monad
+import Criterion
+import Criterion.Main
+import qualified Data.Map.Strict as Map
+import Data.Maybe
+import System.IO.Temp
+import System.Random
+
+import Concordium.ID.Types
+import Concordium.Types
+import qualified Data.FixedByteString as FBS
+
+import Concordium.GlobalState.Persistent.BlobStore
+import qualified Concordium.GlobalState.Persistent.Trie as Trie
+
+-- |The starting state from which the benchmarks are run.
+-- This consists of the blob store, and indexes in the three forms required for testing.
+data Setup = Setup
+    { setupBlobStore :: !BlobStore,
+      setupMemMap :: !(Map.Map RawCredentialRegistrationID AccountIndex),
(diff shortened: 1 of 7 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -1047,6 +1053,84 @@ instance DirectBlobStorable m a => BlobStorable m (Nullable (EagerBufferedRef a)
...
+-- This is essentially a simplified version of 'BufferedRef', where @BRBoth ref v@ is simply
+-- replaced with @URBlobbed ref@.
+data UnbufferedRef a
+    = URBlobbed !(BlobRef a)
+    -- ^A reference that is already on disk
+    | URMemory {urIORef :: !(IORef (Nullable (UnbufferedRef a))), urValue :: !a}
+    -- ^A reference that is in memory and may have been written to disk.
+    -- If the reference has not been written, the 'urIORef' will contain 'Null'.
+    -- If it has been written, it will contain @Some (URBlobbed b)@ for a 'BlobRef' @b@ that
+    -- holds the stored value.
```
abizjak, line 1067: The need for the recursive type here is unclear and should be made clear.

It seems again that it would be a lot clearer if instead it was just `BlobRef a` instead of `UnbufferedRef a`, so at the very least it needs to be clearly explained why this is not.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -1047,6 +1053,84 @@ instance DirectBlobStorable m a => BlobStorable m (Nullable (EagerBufferedRef a)
 instance (Applicative m, Cacheable m a) => Cacheable m (EagerBufferedRef a) where
     cache (EagerBufferedRef ioref v) = EagerBufferedRef ioref <$> cache v
 
+-- |A reference that is generally not retained in memory once it has been written to disk.
```
abizjak, line 1056: What does "generally" mean here?

abizjak (approved):

concordium-consensus/benchmarks/credential-lookup/CredentialLookup.hs
```diff
@@ -0,0 +1,138 @@
...
+
+-- |Benchmark looking up random registration IDs. There should be a 50% chance of each occurring
+-- in the map.
+benchesRandom :: Int -> Benchmark
+benchesRandom size = envWithCleanup (setup size) cleanup $ \cnf ->
+    bgroup
+        ("Map size " ++ show size)
+        [ bench "map" $
+            perRunEnv
+                (genRegID <$> randomRIO (1, 2 * size))
```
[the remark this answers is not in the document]
td202: The problem with making it deterministic is that it is liable to warm up cache lines etc. on repeated runs. Using random inputs seems fine with large numbers of runs.

concordium-consensus/benchmarks/credential-lookup/CredentialLookup.hs
```diff
@@ -0,0 +1,138 @@
...
+            Trie.EmptyTrieN -> return Trie.EmptyTrieN
+            Trie.TrieN sz (BufferedFix br) -> do
+                (_, blobRef) <- refFlush br
+                return $! Trie.TrieN sz (UnbufferedFix (blobRefToUnbufferedRef (BlobRef (theBlobRef blobRef))))
+        return $! Setup{..}
+  where
+    toInsert = [(genRegID x, fromIntegral x) | x <- [1 .. size]]
+    add (m, t) (regid, acctix) = do
+        (_, t') <- storeUpdate =<< Trie.insert regid acctix t
+        return (Map.insert regid acctix m, t')
```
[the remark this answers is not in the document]
td202: Done.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
td202: I've replaced it with a `BlobRef` instead.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
td202: Elaborated.

eb-concordium (approved):
Looks good. I did not understand all the changes in detail, but overall looks good.

PR #541, opened 2022-10-09
Implement the health check endpoint.

abizjak:
## Purpose

Implement a basic health check service. The health check service has reflection enabled since that is required by automatic health checks in, e.g., AWS load balancers.

## Changes

Add a new health check endpoint that mainly checks whether we have recent enough finalized blocks.

If it becomes necessary we can configure additional checks later.

Depends on
- [ ] https://github.com/Concordium/concordium-grpc-api/pull/28

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/configuration.rs
+++ b/concordium-node/src/configuration.rs
@@ -340,7 +340,7 @@ pub struct GRPC2Config {
         help = "Address the GRPC server should listen on, e.g., 127.0.0.1",
         env = "CONCORDIUM_NODE_GRPC2_LISTEN_ADDRESS"
     )]
-    pub listen_addr:      Option<std::net::IpAddr>,
+    pub listen_addr:                Option<std::net::IpAddr>,
     #[structopt(
         name = "grpc2-listen-port",
         long = "grpc2-listen-port",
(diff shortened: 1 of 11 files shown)


concordium-node/src/configuration.rs
```diff
@@ -380,7 +380,15 @@ pub struct GRPC2Config {
...
         requires = "grpc2-listen-addr"
     )]
-    pub endpoint_config:  Option<PathBuf>,
+    pub endpoint_config:            Option<PathBuf>,
+    #[structopt(
+        long = "grpc2-health-max-finalized-delay",
+        help = "Maximum amount of seconds that the time of the last finalized block can be behind \
+                present before the health check fails.",
+        env = "CONCORDIUM_NODE_GRPC2_HEALTH_MAX_FINALIZED_DELAY",
+        default_value = "300"
```
EmilLa1, line 389: Is this a sensible default? 5 min seems too much?

concordium-node/src/health.rs
```diff
@@ -0,0 +1,47 @@
+//! Implementation of the healt check service that is part of the GRPC2
```
EmilLa1, line 1:
```suggestion
//! Implementation of the health check service that is part of the GRPC2
```

EmilLa1 (approved):
Looks good

concordium-node/src/configuration.rs
abizjak: I think it is somewhat sensible. 2min will have a reasonable amount of false positives which is not ideal on mainnet.

Maybe 3min would be good enough. Don't know, but that is also why I made it a configuration option since we have to experiment a bit.

abizjak: So I will keep the default as 5min for now.

PR #544, opened 2022-10-10
Account refactor

td202:
## Purpose

This refactors account handling to support a new `AccountV2` account version that is currently essentially the same as `AccountV1`. It also introduces protocol version `P5`, which uses `AccountV2`.

## Changes

- Types that previously used `'AccountV1` are generalised to support any account version that satisfies `AVSupportsDelegation`.
  - In many instances, patterning matching is done on `delegationSupport @av` rather than `accountVersion @av` (or in some cases on `protocolVersion @pv`). This prevents duplicating the `AccountV1` case for `AccountV2`, since both are wrapped into a single case where we know `AVSupportsDelegation av`.
  - In `ActiveDelegatorInfo` and `ActiveBakerInfo`, `StakePendingChange 'AccountV1` is replaced by `StakePendingChange' Timestamp` in order to make these types independent of the account version. Likewise `bsoProcessPendingChanges` uses `Timestamp` instead of `PendingChangeEffective (AccountVersionFor (MPV m))`. (Both `PendingChangeEffective 'AccountV1` and `PendingChangeEffective 'AccountV2` are effectively wrappers for `Timestamp`, but they are not the same type. Thus, it makes sense to convert to/from `Timestamp` at appropriate interfaces.)
- State migration from `P4` to `P5` is implemented. This does not make any changes to the state at present.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -335,36 +330,38 @@ dispatch (msg, mVerRes) = do
...
+        (
+          ChainParametersVersionFor (MPV m) ~ 'ChainParametersV0,
+          AccountVersionFor (MPV m) ~ 'AccountV0
+        ) => a)
+        -> a
+    onlyWithoutDelegation c = case protocolVersion @(MPV m) of
+      SP1 -> c
+      SP2 -> c
+      SP3 -> c
+      _ -> error "Operation unsupported at this prtocol version."
```
limemloh, line 358: Could this be done by casing on `delegationSupport @(AccountVersionFor (MPV m))` like the function just below?

limemloh, line 358:
```suggestion
    onlyWithoutDelegation c = case protocolVersion @(MPV m) of
      SP1 -> c
      SP2 -> c
      SP3 -> c
      _ -> error "Operation unsupported at this protocol version."
```

concordium-consensus/src/Concordium/ProtocolUpdate.hs
```diff
@@ -33,7 +33,7 @@ checkUpdate = case protocolVersion @pv of
     SP1 -> fmap UpdateP1 . P1.checkUpdate
     SP2 -> fmap UpdateP2 . P2.checkUpdate
     SP3 -> fmap UpdateP3 . P3.checkUpdate
-    SP4 -> const $ Left "Unsupported update."
+    _ -> const $ Left "Unsupported update."
```
limemloh, line 36: Is it intensional not to support updating yet?

limemloh (approved):
No major comments, LGTM.
The `AVSupportsDelegation` constraint is a great addition.

concordium-consensus/src/Concordium/Scheduler.hs
td202: Unfortunately not. The constraint on the chain parameters version cannot be inferred just from knowing that the account version is `AccountV0`.

concordium-consensus/src/Concordium/ProtocolUpdate.hs
td202: I have added an update to P5 now.

EmilLa1 (approved):

PR #546, opened 2022-10-11
Add  build pipeline for database exporter

Thahara:
## Purpose

Add a standalone build for the database exporter.

## Changes

Added Jenkinsfile, and related scripts.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/db-exporter/build.sh
+++ b/scripts/db-exporter/build.sh
@@ -5,9 +5,9 @@ set -euxo pipefail
 # The script must be invoked with the concordium-node source directory as working directory.
 
 # Expected parameters:
-# - BUILD_VERSION
+# - VERSION
 
-build_version="${BUILD_VERSION}"
+version="${VERSION}"
 
 flatbuffers_tag=v2.0.0
 rust_toolchain_version=1.62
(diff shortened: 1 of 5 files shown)


scripts/db-exporter/build.sh
```diff
@@ -0,0 +1,183 @@
...
+CHUNK_SIZE=
+EOF
+
+cat <<EOF > DEBIAN/control
+Package: database-exporter
+Version: $version
+Section: extra
+Priority: optional
+Architecture: amd64
+Depends: debconf ( >= 1.5.73 ), debhelper ( >= 10 ), libunbound2 ( >= 1.6.7 ) | libunbound8 ( >= 1.9.4 ), openssl ( >= 1.1.1 ), libgmp10 ( >= 6.1.2 ), libpq-dev ( >= 9.5.19)
```
abizjak, line 86: I think unbound should not be necessary, and hasn't been for some time now. Neither should openssl.

scripts/db-exporter/build.sh
```diff
@@ -0,0 +1,183 @@
...
+rustup default "$rust_toolchain_version"
+
+# - 'database-exporter'
+
+stack build \
+	--stack-yaml=./concordium-consensus/stack.yaml \
+	--flag=concordium-consensus:-dynamic \
+	--copy-bins \
+	--local-bin-path=. \
+	concordium-consensus:database-exporter
```
abizjak, line 56: An alternative to building the exporter here would be to reuse the `build-static-libraries.sh` script which also builds this binary and puts it into `/binaries` in the resulting archive.

Then you also would not have to copy libraries from `concordium-base/rust-src/target` and wasm-chain integration.

abizjak: Have you considered that?

scripts/db-exporter/build.sh
Thahara: Thank you, they have been removed

abizjak (approved):
This looks good to me, but I have not tested it so I might have missed something.

scripts/db-exporter/build.sh
Thahara: Changed the build to use the `build-static-libraries.sh`  script, it makes the build take a bit longer, but I think that's fine as it should reduce the amount of maintenance required to keep various builds in line.

Issue #545, opened 2022-10-11
Revise storage and hashing of AccountV2 accounts


td202:
Revise the structure of `AccountV2` accounts and the corresponding hashing in accordance with the feature proposal. (This does not include revising the structure of release schedules, which will be handled as a separate task.)

PR #559, opened 2022-10-20
Account restructuring

td202:
## Purpose

Addresses #545.

This introduces the new account structure for `AccountV2`. To do so, persistent accounts are refactored with:
- `Concordium.GlobalState.Persistent.Account` providing the interface to accounts;
- `Concordium.GlobalState.Persistent.Account.StructureV0` providing the implementation with the old structure (supporting `AccountV0` and `AccountV1`); and
- `Concordium.GlobalState.Persistent.Account.StructureV1` providing the implementation with the new structure (supporting `AccountV2`).

## Changes

- `Concordium.GlobalState.Persistent.Account` provides a common interface to functions on accounts that dispatches to `StructureV0` or `StructureV1` depending on the account version.
  - Much of the functionality previously in `Account` is moved to `StructureV0`.
  - `updateAccount` is moved from `Accounts` to `Account(.StructureV0)`.
  - Much of `Concordium.GlobalState.Persistent.BlockState` is rewritten to use the new interface.
- `Concordium.GlobalState.Persistent.BlockState` is refactored to use the interface, rather than working directly on the account representation.
- Account hashing uses a new scheme for `AccountV2`.
  - The old `AccountHashInputs` is renamed `AccountHashInputsV0`.
  - An `AccountHashInputsV2` is introduced for `AccountV2`, which also introduces `AccountMerkleHash` as an intermediate hash.
  - A GADT `AccountHashInputs`, with constructors for each account version, is defined and used as the input for `makeAccountHash`.
- Encrypted amounts:
  - Introduce `EncryptedAmountHash`, which is computed just from serializing the encrypted amount and hashing.
  - Persistent encrypted amounts are moved to `Concordium.GlobalState.Persistent.Account.EncryptedAmount` (out of `Account`).
- The basic state implementation is unchanged in `AccountV3` except for the hashing.
- Add `eagerBufferedDeref` for pure loading of an `EagerBufferedRef`, and conversion from `BufferedRef` to `EagerBufferedRef`.
- The `_auAddress` field is removed from `AccountUpdate`, as it is unused. This also removes unnecessary lookups of account addresses in the scheduler.
- GlobalState tests are updated to use `P5`.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+        PersistentAccountStakeEnduring av
+    PersistentAccountStakeEnduringDelegator ::
+        { paseDelegatorId :: !DelegatorId,
+          paseDelegatorRestakeEarnings :: !Bool,
+          paseDelegatorTarget :: !DelegationTarget,
+          paseDelegatorPendingChange :: !(StakePendingChange' Timestamp)
+        } ->
+        PersistentAccountStakeEnduring av
+
+-- |Convert a 'PersistentAccountStakeEnduring' to an 'AccountStake' given the amount of the stake.
```
abizjak, line 125: Please add why this is a thing we need.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+-- The stake is not stored under a reference (excepting the baker info, as per the definition of
+-- 'PersistentAccountStakeEnduring'). This is since the information is relatively succinct.
+data PersistentAccountEnduringData (av :: AccountVersion) = PersistentAccountEnduringData
+    { -- |The Merkle hash computed from the other fields.
+      paedHash :: !(AccountMerkleHash av),
+      -- |A reference to the persisting account data.
+      paedPersistingData :: !(EagerBufferedRef PersistingAccountData),
+      -- |The encrypted amount. Invariant: if this is present, it will not satisfy
+      -- 'isZeroAccountEncryptedAmount'.
+      paedEncryptedAmount :: !(Nullable (LazyBufferedRef PersistentAccountEncryptedAmount)),
```
abizjak, line 204: Check that this is properly guaranteed. An account can end in this state even after having used encrypted amounts.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+        Some e -> getHash <$> (loadPersistentAccountEncryptedAmount =<< refLoad e)
+    amhi2AccountReleaseScheduleHash <- case paedReleaseSchedule of
+        Null -> return Transient.emptyAccountReleaseScheduleHash
+        Some (rs, _) -> getHashM rs
+    let hashInputs :: AccountMerkleHashInputs 'AccountV2
+        hashInputs = AccountMerkleHashInputsV2{..}
+        !paedHash = getHash hashInputs
+    return $! PersistentAccountEnduringData{..}
+
+-- |[For internal use.] Recompute the Merkle hash of the enduring account data.
```
abizjak, line 244: Please elaborate what "for internal use" means. What is the scope of "internal"? And why?

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+-- |Load a 'PendingChangeFlags' from the low-order 2 bits of a 'Word8'.
+-- All other bits must be 0.
+pendingChangeFlagsFromBits :: Word8 -> Either String PendingChangeFlags
+pendingChangeFlagsFromBits 0b00 = return PendingChangeNone
+pendingChangeFlagsFromBits 0b01 = return PendingChangeReduce
+pendingChangeFlagsFromBits 0b10 = return PendingChangeRemove
+pendingChangeFlagsFromBits _ = Left "Invalid pending change type"
+
+-- |Flags that represent the nature of the stake on an account.
+-- These are stored as the low-order 6 bits of a 'Word8'.
```
abizjak, line 298: This sentence does not really explain how they are stored and what those 6 bits mean.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+
+instance HashableTo (AccountHash 'AccountV2) (PersistentAccount 'AccountV2) where
+    getHash PersistentAccount{..} =
+        makeAccountHash $
+            AHIV2 $
+                AccountHashInputsV2
+                    { ahi2NextNonce = accountNonce,
+                      ahi2AccountBalance = accountAmount,
+                      ahi2StakedBalance = accountStakedAmount,
+                      ahi2MerkleHash = getHash accountEnduringData
```
abizjak, line 504: If I understand correctly this will load the entire enduring data to compute the hash (which is then just looked up).

It would be better to not have to load all that data and that seems possible since the storage format puts the hash at the start.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+isAllowed :: Monad m => PersistentAccount av -> AccountAllowance -> m Bool
+isAllowed acc AllowedEncryptedTransfers = do
+    creds <- getCredentials acc
+    return $! Map.size creds == 1
+isAllowed acc AllowedMultipleCredentials = do
+    let ed = enduringData acc
+    -- We use the invariant that if the encrypted amount is present then it will be non-empty.
+    return $! case paedEncryptedAmount ed of
+      Null -> True
+      Some _ -> False
```
abizjak, line 596: Since we are changing the protocol then we don't have to keep the semantics exactly, but it seems to me that the semantics here has changed a bit. Previously if you ended up with 0 (encrypted in a specific way) you could then add multiple credentials on the account.

In this version it seems we never go back to `Null` once we do anything with encrypted transfers. This is probably OK, but it is a change in behaviour that was not planned.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+
+-- |Get the list of credentials deployed on the account, ordered from most
+-- recently deployed.  The list should be non-empty.
+getCredentials :: (Monad m) => PersistentAccount av -> m (Map.Map CredentialIndex RawAccountCredential)
```
abizjak, line 600: This comment does not appear to be accurate.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1314 @@
...
+-- This is because we typically require ready access to the locked balance so that we can compute
+-- the available balance.
+--
+-- The stake is not stored under a reference (excepting the baker info, as per the definition of
+-- 'PersistentAccountStakeEnduring'). This is since the information is relatively succinct.
+data PersistentAccountEnduringData (av :: AccountVersion) = PersistentAccountEnduringData
+    { -- |The Merkle hash computed from the other fields.
+      paedHash :: !(AccountMerkleHash av),
+      -- |A reference to the persisting account data.
+      paedPersistingData :: !(EagerBufferedRef PersistingAccountData),
```
abizjak, line 201: PersistingAccountData has two copies of account keys. Would it not be better to also have a new version of that with a more refined representation of the accountinformation and credentials to remove duplication? RawAccountCredential does not have much more data, and we could probably just have that and not the account keys.

Having just account keys would be good as well, but then we'd have to have the ability to load them without loading everything else.

abizjak (changes requested):

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/Structure.hs
```diff
@@ -0,0 +1,61 @@
+module Concordium.GlobalState.Persistent.Account.Structure where
```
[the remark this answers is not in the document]
td202: Since there is not so much here, I have moved the definitions to `Concordium.GlobalState.Account`.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account.hs
```diff
@@ -1,928 +1,544 @@
...
-        when asfHasRemovedCredentials $ put (_accountRemovedCredentials ^. unhashed)
-        put _accountNonce
-        put _accountAmount
-        putEA
-        when asfExplicitReleaseSchedule $ put arSched
-        when asfHasBakerOrDelegation $ serializeAccountStake stake
+serializeAccount :: (MonadBlobStore m, MonadPut m) => GlobalContext -> PersistentAccount av -> m ()
+serializeAccount gc (PAV0 acc) = V0.serializeAccount gc acc
+serializeAccount gc (PAV1 acc) = V0.serializeAccount gc acc
+serializeAccount gc (PAV2 acc) = V1.serializeAccount gc acc
```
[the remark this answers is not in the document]
td202: I am not removing it since it is still used for migration in the basic block state, and it is simpler to retain it than remove everything that is involved at present.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV0.hs
```diff
@@ -0,0 +1,1215 @@
...
+--
+-- * For 'AllowedMultipleCredentials' the account must have the empty encrypted balance.
+isAllowed :: MonadBlobStore m => PersistentAccount av -> AccountAllowance -> m Bool
+isAllowed acc AllowedEncryptedTransfers = do
+    creds <- getCredentials acc
+    return $! Map.size creds == 1
+isAllowed acc AllowedMultipleCredentials = do
+    PersistentAccountEncryptedAmount{..} <- refLoad (acc ^. accountEncryptedAmount)
+    if null _incomingEncryptedAmounts && isNothing _aggregatedAmount
+        then isZeroEncryptedAmount <$> refLoad _selfAmount
```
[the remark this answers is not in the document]
td202: This test is different because it does not check `_startIndex`, and I don't want to change the semantics.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
td202: I have revised this to `isInitialPersistentAccountEncryptedAmount` and it is checked and preserved.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
td202: That is not correct. The enduring data is stored under an `EagerBufferedRef`, which means that it is always loaded when the account is loaded.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
td202: I have changed this back so it will do the `isZero` check now.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -1356,11 +1340,10 @@ delegationCheckTargetOpen
     -> m ()
 delegationCheckTargetOpen _ Transactions.DelegatePassive = return ()
 delegationCheckTargetOpen bsp (Transactions.DelegateToBaker bid@(BakerId baid)) = do
-    Accounts.indexedAccount baid (bspAccounts bsp) >>= \case
-        Just PersistentAccount{_accountStake = PersistentAccountStakeBaker abr} -> do
-            baker <- refLoad abr
-            poolInfo <- refLoad (baker ^. bakerPoolInfoRef)
-            case poolInfo ^. BaseAccounts.poolOpenStatus of
+    -- This may load more of the baker information than is actually required
```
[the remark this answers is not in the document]
td202: Removed

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -808,47 +805,45 @@ doTransitionEpochBakers pbs newEpoch = do
...
+                        -- Reduction takes effect next epoch, so apply it in the generated list
+                        | redEpoch == newEpoch + 1 -> do
+                            return (bs0, (binfoRef, newAmt) : bkrs0)
+                        -- Reduction complete, so update the account as well
+                        | redEpoch <= newEpoch -> do
+                            -- Reduce the baker's stake on the account by applying the change
+                            newAccounts <- Accounts.updateAccountsAtIndex' applyPendingStakeChange aid (bspAccounts bs0)
+                            -- The baker is included with the revised stake
+                            return (bs0 {bspAccounts = newAccounts}, (binfoRef, newAmt) : bkrs0)
+                    _ -> return (bs0, (binfoRef, _stakedAmount acctBkr) : bkrs0)
```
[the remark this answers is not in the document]
td202: I have reinstated much of the whitespace as it was. You can disable whitespace in the diff.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
abizjak: I do not like this very much and I thought that was in scope.

But it is too late now, so OK.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1393 @@
...
+        let newInfo =
+                oldInfo
+                    & bieBakerInfo
+                        %~ ( (bakerAggregationVerifyKey .~ bkuAggregationKey upd)
+                                . (bakerSignatureVerifyKey .~ bkuSignKey upd)
+                                . (bakerElectionVerifyKey .~ bkuElectionKey upd)
+                           )
+        newInfoRef <- refMake $! newInfo
+        return $! baker{paseBakerInfo = newInfoRef}
+    _ -> error "setBakerKeys invariant violation: account is not a baker"
```
abizjak, line 1022: In principle it would be better for maintainability not to have these catch-all cases and instead list all the current cases.

That makes it harder to screw up if we are adding cases.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -0,0 +1,1393 @@
...
+    PersistentAccount 'AccountV2 ->
+    m (PersistentAccount 'AccountV2)
+setRestakeEarnings newRestake =
+    updateStake $
+        return . \case
+            baker@PersistentAccountStakeEnduringBaker{} ->
+                baker{paseBakerRestakeEarnings = newRestake}
+            del@PersistentAccountStakeEnduringDelegator{} ->
+                del{paseDelegatorRestakeEarnings = newRestake}
+            _ -> error "setRestakeEarnings invariant violation: account is not a baker or delegator"
```
abizjak, line 1048: Likewise here and in the functions below.

abizjak (approved):
Looks ok. Only a few issues that in principle can be postponed.

Let's get this merged as soon as possible since it affects other PRs.

eb-concordium (approved):

Issue #561, opened 2022-10-24
Revise top-level release schedule map


td202:
The block state maintains a global index of the accounts that have scheduled releases, recording when their next release is scheduled. The current implementation has a number of issues:

-    It maps account addresses to the next release time. This makes it inefficient to determine the which accounts have expired releases (it’s linear in the number of accounts with releases). The keys of the map should instead be the release times, so the lookup is logarithmic.

-    It would be more efficient to refer to accounts by index than account address.

-    If there are accounts with scheduled releases, the map is rewritten for every block, regardless of whether releases occur.

-    The index is written in the state as a flat serialized map. This means that any access to it is going to be linear in the number of accounts with release schedules.

This should be changed to use a trie map from release times to (lists of) account indexes. We can also cache the minimum release time to avoid the lookup in the common case where there are no releases to process.

For earlier protocol versions, we can use a map from release time to account addresses that is compatible with the old serialization.

PR #570, opened 2022-10-26
Release schedule top level changes

td202:
## Purpose

Addresses #561

This changes the handling of the top-level release schedule index to improve how it is handled.

## Changes

- Remove `bsoAddReleaseSchedule` and change the semantics of `bsoModifyAccount` to update the top-level release schedule index as necessary when adding releases to an account.
- Add an operation on accounts to get the next scheduled release timestamp (if any).
- For the basic (in-memory) block state, revise the release schedule to be a map from timestamps to sets of account indexes. `Concordium.GlobalState.Basic.BlockState.ReleaseSchedule` implements this and provides an interface for the basic operations on the release schedule.
- For the persistent block state, there are two implementations of the release schedule, found in `Concordium.GlobalState.Persistent.ReleaseSchedule`:
  - `LegacyReleaseSchedule` uses a map from timestamps to sets of account addresses. This is implemented to have a backwards compatible serialisation with the former implementation that was a map from account addresses to timestamps. (Indexing by timestamp is preferable, as it is relatively efficient to use as a priority queue.)
  - `NewReleaseSchedule` uses a trie from timestamps to account indexes. This is not backwards compatible, so is only used for `P5`.
  - The actual persistent `ReleaseSchedule` type is a GADT indexed by the protocol version.
- Block state migration for the release schedules is implemented.
- Conversion from in-memory to persistent release schedules is implemented.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
```diff
@@ -0,0 +1,415 @@
...
+
+-- |A release schedule for the P5 protocol version.  This uses a 'Trie.Trie' mapping 'Timestamp's
+-- to sets of accounts.
+data NewReleaseSchedule = NewReleaseSchedule
+    { -- |The first timestamp at which a release is scheduled (or the maximum possible timestamp if
+      -- no releases are scheduled).
+      rs5FirstTimestamp :: !Timestamp,
+      -- |A map recording the first release time for each account with a pending release.
+      -- An account should occur at most once in the map.
+      rs5Map :: !(Trie.TrieN BufferedFix Timestamp AccountSet)
```
abizjak, line 151: I'm guessing this relies quite crucially on the fact timestamps are serialized in big-endian. Could we please record this explicitly?

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
```diff
@@ -0,0 +1,415 @@
...
+
+instance MonadBlobStore m => BlobStorable m AccountSet
+instance Applicative m => Cacheable m AccountSet
+
+-- |A release schedule for the P5 protocol version.  This uses a 'Trie.Trie' mapping 'Timestamp's
+-- to sets of accounts.
+data NewReleaseSchedule = NewReleaseSchedule
+    { -- |The first timestamp at which a release is scheduled (or the maximum possible timestamp if
+      -- no releases are scheduled).
+      rs5FirstTimestamp :: !Timestamp,
```
abizjak, line 148: Isn't it technically possible somebody schedules a release at that time?

From that perspective it would be better to use 0 as a sentinel value or really a Maybe, since that cannot be used by a valid transaction.

I see that would require a few changes down below but it does not seem that much would be affected.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
```diff
@@ -0,0 +1,415 @@
...
+            NewReleaseSchedule
+                { rs5FirstTimestamp = min newts (rs5FirstTimestamp rs),
+                  ..
+                }
+      where
+        remAcc Nothing = error "updateAccountRelease: no entry at expected release time"
+        remAcc (Just (AccountSet accs)) =
+            return $!
+                let accs' = Set.delete ai accs
+                 in if Set.null accs' then ((), Trie.Remove) else ((), Trie.Insert (AccountSet accs'))
```
abizjak, line 201: That's a funny single space indent.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
abizjak, line 148: Also our usual naming convention would be `nrsFirstTimestamp`.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
```diff
@@ -0,0 +1,415 @@
...
+    rsP1 = do
+        rs5Map <- Trie.fromList (second AccountSet <$> Map.toList (Transient.rsMap tRS))
+        return $!
+            ReleaseScheduleP5
+                NewReleaseSchedule
+                    { rs5FirstTimestamp = Transient.rsFirstTimestamp tRS,
+                      ..
+                    }
+
+-- |(For testing purposes) get the map of the earliest scheduled releases of each account.
```
abizjak, line 404: Maybe add a warning pragma like we do for similar functions.

concordium-consensus/tests/globalstate/GlobalStateTests/AccountReleaseScheduleTest.hs
```diff
@@ -121,14 +126,15 @@ testing = do
...
-                 ) b' (map ((accA, aiA),) txsA ++ map ((accB, aiB),) txsB)
-  checkEqualBlockReleaseSchedule bs''
+                 ) (bs1, bs2) (map ((accA, aiA),) txsA ++ map ((accB, aiB),) txsB)
+  let expectedRS = OrdMap.unionWith (<>)
+        (OrdMap.singleton (head timestampsA) (Set.singleton aiA))
+        (OrdMap.singleton (head timestampsB) (Set.singleton aiB))
+  checkEqualBlockReleaseSchedule expectedRS bs''
   let taggedAmounts = OrdMap.toAscList $ OrdMap.unionWith (++) (OrdMap.map ((:[]) . (accA,)) amountsA) (OrdMap.map ((:[]) . (accB,)) amountsB)
   _ <- foldlM (\b (ts, accs) -> do
             newB <- bsoProcessReleaseSchedule b ts
```
abizjak, line 140: Should we not also test this for P5?

abizjak (commented):
In general I don't see any major issues and this can be merged.

But I would strongly prefer that we don't use maxBound::Timestamp as a sentinel value.
I think it does work fine, but it is just a bit messy and I don't think it would complicate matters to just use a `Maybe` or `Some` other than having another helper function to compare the first release.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
td202: That's possible, and it's not a problem. Using a Maybe would just add complexity everywhere this is used. I've add documentation that you shouldn't infer that the map is empty because the timestamp is maxBound.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
td202: It does not. We do require the instance `OrdFixedTrieKey Timestamp`, which is derived via `Word64`, and does rely on big-endian serialization for `Word64`.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
td202: I don't like it, but the formatter does it.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
td202: I don't think it's warranted in this case. "For testing purposes" is mainly meant as an explanation of why this code exists in the first place.

concordium-consensus/tests/globalstate/GlobalStateTests/AccountReleaseScheduleTest.hs
td202: This does test for P5.

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
abizjak: I'm not sure why you say it does not, but then in the followup sentence you say that it does. In particular the findMin operation requires this. I understand that there is a OrdFixedTrieKey abstraction, but it would help so much to understand this if we just explicitly stated that using the Trie here makes sense because of big endian serialziation.

eb-concordium (approved):

abizjak (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/ReleaseSchedule.hs
td202: It does not rely on the `Serialize Timestamp` instance. Documenting this here would be like documenting why the `Ord` instance is correct for the key type in any structure that uses `Map`, which seems out of place to me.

Issue #566, opened 2022-10-26
Change hashing scheme for transaction outcomes


EmilLa1:
A full feature description is available here: https://concordium.atlassian.net/wiki/spaces/EN/pages/1040547877/Change+hashing+of+transaction+outcomes

A short summary 
The goal of this epic is to no longer include the exact `RejectReason`s in the hash of the transaction outcomes within a block as this makes it impossible to change the logic around verifying transactions e.g. change the order of the checks etc. 
Note. The exact `RejectReason` is not important as it will have no effects on the chain besides the cost of executing the transaction. 

The new hashing scheme should moreover be using a Merkle tree instead of a hashing list in order to make it possible for new types of queries (this could later be used for tuning the catchup mechanism) by allowing queries such as "the outcome of a transaction X in block Y is Z", but this task is not in the scope for this feature.

PR #572, opened 2022-10-26
New hashing scheme for TransactionOutcomes. 

EmilLa1:
## Purpose

The purpose of this PR is outlined in the associated epic which can be found here: https://github.com/Concordium/concordium-node/issues/566

This PR builds upon the PR in `concordium-base` https://github.com/Concordium/concordium-base/pull/247 which introduces the types required for versioning the `TransactionOutcomes`. 

The transaction outcomes are now versioned and implies the following. 

Transaction outcomes are now part of one of two gadts `BasicTransactionOutcomes` for the `Basic` block state implementation or `PersistentTransactionOutcomes` for the `Persistent` block state implementation. 

The `TransactionOutcomesVersion` determines whether they are kept as is (currently) based on the protocol version. Semantically nothing is changed for P1 to P4. 

For protocol version 5 the `TransactionOutcomes` are being stored in a merkle tree, hence we use the `LFMB` tree(s) we already have. (One for the `Basic` block state implementation and one for the `Persistent` block state implementation. 

Note. The new  version of the `TranasctionOutcomes` are not using the `RejectReason`s when computing the hash for each individual `TransactionSummary`. (See `base` changes`). 

## Changes

Most notably types that have been added. 

- `MerkleTransactionOutcomes` for each of the block state implementations. This holds a merkle tree for normal outcomes and for special outcomes. 
- `BasicTransactionOutcomes` and `PersistentTransactionOutcomes` gadts which ensures that the correct transaction outcomes are being used based on the `ProtocolVersion`.
- And a few supporting types/constraints which supports the above changes. 

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Block.hs
```diff
@@ -313,7 +312,7 @@ instance BlockData (Block pv) where
     blockTransactions (NormalBlock bb) = blockTransactions bb
 
     -- move into gendata?
-    blockTransactionOutcomesHash GenesisBlock{} = getHash emptyTransactionOutcomes
+    blockTransactionOutcomesHash GenesisBlock{} = getHash emptyTransactionOutcomesV0
```
abizjak, line 315: This needs to depend on the protocol version.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -377,6 +386,38 @@ class (Monad m, BlockStateTypes m) => ModuleQuery m where
...
+  getHash (TransactionSummaryV1 summary) = H.hash $! S.runPut $!
+      putMaybe S.put (tsSender summary) <>
+      S.put (tsHash summary) <>
+      S.put (tsCost summary) <>
+      S.put (tsEnergyCost summary) <>
+      S.put (tsType summary) <>
+      -- TODO: This is wrong, we only do this for failure, not for all summaries.
+      -- |We simply put a '1' indicating a failure.
+      S.putWord8 1 <>
+      S.put (tsIndex summary)
```
abizjak, line 409: Fix.

concordium-consensus/src/Concordium/GlobalState/Block.hs
```diff
@@ -313,7 +314,18 @@ instance BlockData (Block pv) where
...
+        case transactionOutcomesVersion @(TransactionOutcomesVersionFor pv) of
+            -- |There are no transactions in the genesis block.
+            -- Hence the 'blockTransactionOutcomesHash's are simply constants.
+            -- For 'STOV0' we can easily use 'emptyTransactionOutcomesV0' for generating the constant.
+            -- In the case of 'STOV1' we simply hash the utf8 encoded string "GENESIS_TRANSACTION_OUTCOMES_HASH"
+            -- since the 'emptyTransactionOutcomes' is determined by the concrete block state implementation in use
+            -- See 'MerkleTransactionOutcomes' in the 'Basic.BlockState' and 'Persistent.BlockState' modules and we do
+            -- not know that context here. So we simply compute the hash of the above mentioned string.
+            STOV0 -> getHash emptyTransactionOutcomesV0
+            STOV1 -> TransactionOutcomesHash $! Hash.hash $! LByteString.toStrict $! ByteString.toLazyByteString $! ByteString.stringUtf8 "GENESIS_TRANSACTION_OUTCOMES_HASH"
```
abizjak, line 327: TODO.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -379,6 +388,46 @@ class (Monad m, BlockStateTypes m) => ModuleQuery m where
...
+    deriving(Eq, Show)
+
+-- |A 'HashableTo' instance for a 'TransactionSummary'' which
+-- omits the exact reject reason.
+-- Failures are simply tagged with a '0x1' byte.
+-- Note. The hash is computed lazily on purpose as the summary can be large
+-- i.e., the resulting events.
+instance HashableTo H.Hash TransactionSummaryV1 where
+  getHash (TransactionSummaryV1 summary) = H.hashLazy $! S.runPutLazy $!
+      putMaybe S.put (tsSender summary) <>
```
abizjak, line 404: Make this fixed size to prevent any issues with collisions.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -379,6 +388,46 @@ class (Monad m, BlockStateTypes m) => ModuleQuery m where
...
+      S.put (tsHash summary) <>
+      S.put (tsCost summary) <>
+      S.put (tsEnergyCost summary) <>
+      S.put (tsType summary) <>
+      encodeValidResult (tsResult summary) <>
+      S.put (tsIndex summary)
```
abizjak, line 410: Move this up, since it is fixed size.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -379,6 +388,49 @@ class (Monad m, BlockStateTypes m) => ModuleQuery m where
...
+      -- |Encode the 'ValidResult' omitting the exact 'RejectReason' if the
+      -- transaction failed. Otherwise we encode the resulting events in the
+      -- resulting outcome hash.
+      encodeValidResult :: S.Putter ValidResult
+      encodeValidResult (TxSuccess events) = S.putWord8 0 <> putListOf putEvent events
+      -- We omit the exact 'RejectReason'.
+      encodeValidResult (TxReject _) = S.putWord8 1
+      encodeSender :: S.Putter (Maybe AccountAddress)
+      encodeSender Nothing = S.putByteString $! BS.replicate 32 0
+      encodeSender (Just sender) = S.put sender
```
abizjak, line 421: While this is almost certainly fine, even better is to also add a tag 0 for nothing, 1 for sender.

EmilLa1: Good catch! Fixed

abizjak (approved):
The same comment as in the corresponding base applies.

It would fit more with the other parts of the codebase if the summary and the hash were parametrized by a either `pv` or `tov`, depending on what makes more sense.

But that would break quite a few things (as in, they would need to be refactored) so we cannot do it now.

PR #581, opened 2022-10-31
Fix migration of V0 account release schedule to V1.

### LINKED ISSUE #580: Scheduled releases are migrated incorrectly on P4->P5 protocol update
**Bug Description**
The scheduled releases are migrated incorrectly in the protocol update from P4 to P5. Specifically, if there are multiple pending releases from a single scheduled transfer, these can be duplicated on protocol update.

The root cause of this issue is that the migration creates a copy of the entire entry for a scheduled transfer for each of the releases in the transfer (that are still pending at the time of the protocol update).

**Steps to Reproduce**

1. At protocol P4, send a scheduled transfer with multiple releases sufficiently far in the future.
2. Perform a protocol update to P5.
3. Query the account to observe the pending releases.

**Expected Result**

The pending releases should be exactly those scheduled.

**Actual Result**

The pending releases appear to be duplicated.

**Versions**
- Software Version: 5.0.0

td202:
## Purpose

Fixes #580.

This corrects a bug in the migration of account release schedules on protocol update. Previously, the migration lead to scheduled transfers appearing to be duplicated on protocol update.

## Changes

- Reworked the function `fromAccountReleaseScheduleV0` to fix the error.
- Add a test case for the bug.
- Improve documentation.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/AccountReleaseScheduleV1.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/AccountReleaseScheduleV1.hs
@@ -156,7 +156,7 @@ insertEntry entry = ins
 -- | Insert a new schedule in the structure.
 --
 -- Precondition: The given list of timestamps and amounts MUST NOT be empty. Moreover, the
--- timestamps MUST be increasing.
+-- timestamps MUST be in ascending order.
 addReleases :: ([(Timestamp, Amount)], TransactionHash) -> AccountReleaseSchedule -> AccountReleaseSchedule
 addReleases (l@(h : t), rseTransactionHash) AccountReleaseSchedule{..} =
     AccountReleaseSchedule (insertEntry entry arsReleases) (arsTotalLockedAmount + sum (snd <$> l))
--- a/concordium-consensus/src/Concordium/Scheduler/Environment.hs
+++ b/concordium-consensus/src/Concordium/Scheduler/Environment.hs
@@ -377,9 +377,21 @@ class (StaticInformation m, ContractStateOperations m, MonadProtocolVersion m) =
   -- |Transfer a scheduled amount from the first address to the second and run
   -- the computation in the modified environment.
   --
-  -- Precondition: The list of releases MUST be non-empty and the timestamps MUST be in increasing
-  -- order.
-  withScheduledAmount :: IndexedAccount m -> IndexedAccount m -> Amount -> [(Timestamp, Amount)] -> TransactionHash -> m a -> m a
+  -- Precondition: The list of releases MUST be in ascending order of timestamps, and the total
+  -- of the scheduled releases MUST be equal to the amount being sent.
+  withScheduledAmount :: 
+    -- |Sender address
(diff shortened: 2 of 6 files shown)


abizjak (approved):

concordium-consensus/src/Concordium/Scheduler/Environment.hs
```diff
@@ -376,6 +376,9 @@ class (StaticInformation m, ContractStateOperations m, MonadProtocolVersion m) =
 
   -- |Transfer a scheduled amount from the first address to the second and run
   -- the computation in the modified environment.
+  --
+  -- Precondition: The list of releases MUST be non-empty and the timestamps MUST be in increasing
```
EmilLa1, line 380: Regarding this and `addScheduledAmountToCS`. 
I do not have a complete overview, but have you considered enforcing the non empty list  on the type level?

In that way we wouldn't be relying on the documentation (even it's fine to do so, that's why they are there in the first place :) ) 

But we could fail earlier than in `fromAccountReleaseScheduleV0#mkEntry` then.

EmilLa1, line 380:
```suggestion
  -- Precondition: The list of releases MUST be non-empty and the timestamps MUST be in ascending
```

concordium-consensus/src/Concordium/Scheduler/Environment.hs
```diff
@@ -584,6 +587,9 @@ addAmountToCS' ai !amnt !cs =
 
 
 -- |Record a list of scheduled releases that has to be pushed into the global map and into the map of the account.
+--
+-- Precondition: The list of releases MUST be non-empty and the timestamps MUST be in increasing
```
EmilLa1, line 591:
```suggestion
-- Precondition: The list of releases MUST be non-empty and the timestamps MUST be in ascending
```

concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/AccountReleaseScheduleV1.hs
```diff
@@ -154,7 +155,8 @@ insertEntry entry = ins
 
 -- | Insert a new schedule in the structure.
 --
--- Precondition: The given list of timestamps and amounts MUST NOT be empty.
+-- Precondition: The given list of timestamps and amounts MUST NOT be empty. Moreover, the
+-- timestamps MUST be increasing.
```
EmilLa1, line 159:
```suggestion
-- timestamps MUST be ascending.
```

EmilLa1 (approved):
Great catch!

concordium-consensus/src/Concordium/Scheduler/Environment.hs
td202: I think that's a fairly good idea, but I won't do it in this PR. In fact, for these functions it's OK for the list to be empty (it will just do nothing), and I have updated the documentation accordingly.

EmilLa1: Great!

PR #583, opened 2022-11-01
Additional testing and documentation for account release schedules.

td202:
## Purpose

This adds some additional test cases and documentation for account release schedules.

## Changes

- Add an in-memory version of the blob store for use in testing.
- Tests for migration from persistent V0 account release schedule to persistent V1 account release schedule after performing arbitrary operations.
- Documentation.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -41,7 +41,7 @@ on:
   workflow_dispatch: # allow manual trigger
 
 env:
-  dummy: 7                 # change to force cache invalidation
+  dummy: 8                 # change to force cache invalidation
   CARGO_TERM_COLOR: always # implicitly adds '--color=always' to all cargo commands
   TEST_LEVEL: 1            # for stack tests
 


concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/AccountReleaseScheduleV1.hs
```diff
@@ -2,6 +2,23 @@
...
+-- a hash of these releases.
+--
+-- On disk, the vector is represented as a flat array. This has a disadvantage that it needs to be
+-- rewritten in its entirety for each change. Each entry, however, avoids rewriting the list of
+-- releases by storing an offset to the next entry in the list. This way, only one copy of the
+-- releases for each transaction is stored in the blob store, but the offset is updated to account
+-- for the lapsed releases. The list of releases for each entry is also stored as a flattened
+-- array of timestamps and amounts, in ascending order of timestamp. The hash of the releases for
+-- an entry is computed incrementally, but only the final hash is retained. The transaction hash
+-- responsible for an entry is recorded, but does not contribute to the hash.
```
EmilLa1, line 21: It's a bit unclear to me what exactly you mean here. 
Can it be rephrased? 
In particular what do you mean by responsibility? The tx hash is an outcome of the said release schedule (and other things of course) i.e. the hashed list of releases. So they do contribute to the final transaction hash?

EmilLa1 (approved):
LGTM.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -590,6 +596,57 @@ deriving via (LiftMonadBlobStore (ExceptT e) m)
...
+    loadRaw (BlobRef offset) = do
+        mv <- asks theMemBlobStore
+        bs <- liftIO (readMVar mv)
+        let bs' = LBS.drop (fromIntegral offset) bs
+        case runGetLazy getWord64be (LBS.take 8 bs') of
+            Left e -> error e
+            Right len -> return $! LBS.toStrict (LBS.take (fromIntegral len) (LBS.drop 8 bs'))
+    flushStore = return ()
+
+    getCallbacks = error "getCallbacks not supported for MemBlobStore"
```
abizjak, line 641: I don't see that the new test really crucially needs this memblobstore.

So while I think it would be nice to have, and  other tests should over time be migrated to it, we need to do this properly and implement the callbacks as well.

We also should add a test to make sure that the memblobstore and the actual blob store behave the same, so that the testing with mem blob store is trustworthy.

abizjak (changes requested):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -590,6 +596,57 @@ deriving via (LiftMonadBlobStore (ExceptT e) m)
 deriving via (LiftMonadBlobStore (ReaderT r) m)
     instance MonadBlobStore m => MonadBlobStore (ReaderT r m)
 
+-- * In-memory blob store
+
+-- |An in-memory blob store. This is intended for testing purposes, as it can be more convenient
+-- than working with files. Internally, the blob store is handled as a lazy 'LBS.ByteString' guarded
+-- by an 'MVar'.
+newtype MemBlobStore = MemBlobStore {theMemBlobStore :: MVar LBS.ByteString}
```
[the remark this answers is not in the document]
td202: I don't think that's necessary. I've also now expanded this to be a complete implementation of 'BlobStore'.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/AccountReleaseScheduleV1.hs
td202: I have tried to make this clearer. While some of the same data is used to derive the transaction hash in the first instance, the transaction hash itself makes no contribution to the account release schedule hash.

concordium-consensus/tests/globalstate/GlobalStateTests/BlobStore.hs
```diff
@@ -0,0 +1,258 @@
...
+        0 <= i && i < n && tst (bs Seq.<| s) n cont
+    tst s n (LoadViaCallbacks i j cont) =
+        0 <= i && i < n && 0 <= j && j < Seq.length s && tst s n cont
+
+foreign import ccall "dynamic" callLoadCallback :: LoadCallback -> LoadCallbackType
+foreign import ccall "dynamic" callStoreCallback :: StoreCallback -> StoreCallbackType
+
+foreign import ccall "return_value_to_byte_array" vectorToByteArray :: Ptr Vec -> Ptr CSize -> IO (Ptr Word8)
+foreign import ccall unsafe "free_array_len" freeByteArray :: Ptr Word8 -> Word64 -> IO ()
+foreign import ccall unsafe "box_vec_u8_free" freeVector :: Ptr Vec -> IO ()
```
abizjak, line 190: Why do we need to re-import these here. I am not a fan of duplicating this.

abizjak (approved):

concordium-consensus/tests/globalstate/GlobalStateTests/BlobStore.hs
td202: "free_array_len" we could import from `Concordium.Crypto.FFIHelpers`. The other two are not re-exported from `Concordium.Scheduler.WasmIntegration.V1` and are imported at different types. I think the choice here is reasonable.

PR #585, opened 2022-11-02
Fix grpc V2 API encoding of a baker transaction with no effects.

abizjak:
- Ensure parameter size limit on InvokeContract.
- Bump version to 5.0.2

## Purpose

Two small bugfixes.

## Changes

- Enforce parameter size limit in InvokeContract
- Fix GRPC V2 API so that it correctly encodes a transction with no effects.

Depends on
- [ ] https://github.com/Concordium/concordium-base/pull/255

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/Cargo.lock
+++ b/concordium-node/Cargo.lock
@@ -3150,7 +3150,7 @@ dependencies = [
 
 [[package]]
 name = "rpassword"
-version = "5.0.2"
+version = "5.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


concordium-node/Cargo.lock
```diff
@@ -3150,7 +3150,7 @@ dependencies = [
 
 [[package]]
 name = "rpassword"
-version = "5.0.1"
+version = "5.0.2"
```
abizjak, line 3153:
```suggestion
version = "5.0.1"
```

PR #603, opened 2022-11-10
Make CI fail on warnings

### LINKED ISSUE #606: Fail CI on build warnings
To ensure we don't build up build warnings, the CI should fail on warning.
This is already true for Rust builds, but not Haskell.

limemloh:
## Purpose

Make CI fail when encountering Haskell warnings.
Should be merged after https://github.com/Concordium/concordium-base/pull/266.
Closes #606.

## Changes

- Add ghc-option to fail on warnings.
- Fix the warnings in the haskell code.
- Make the build run after checking formatting.


limemloh:
This PR is not yet ready for review, but I have marked it so, in order to run and test the CI.

.github/workflows/build-test.yaml
```diff
@@ -46,7 +54,67 @@ env:
   TEST_LEVEL: 1            # for stack tests
 
 jobs:
+
+  fourmolu:
```
abizjak, line 58: In the README we have
```
  see the file [.github/workflows/rustfmt.yaml](.github/workflows/rustfmt.yaml),
  look for `nightly-...`).
```

please fix that now that the file is gone and the job is moved.

abizjak (approved):
In the README we have

```

PR #617, opened 2022-11-22
Node should terminate gracefully if it can not start the configured services.

vikt0r0:
## Purpose

If services that were configured to start fail, the node currently just resumes operation. In case this happens, it makes more sense to shut down the node gracefully.

## Changes

- [X] Add channels allowing services to message the main thread and other running threads (e.g. out-of-band catchup).
- [X] Have services send messages to the main thread on errors.
- [X] Let the main thread terminate gracefully in the event of a failure.
- [x] Print causes of errors in error messages from services.

## Tests
Test that the node terminates in case the
- [X] P2P service fails.
- [X] GRPC service fails.
- [X] GRPC2 service fails.
- [x] Instrumentation fails.

## Checklist

      hard-to-understand areas.


vikt0r0:
I'm not sure how to test what happens when the instrumentation service fails. Here are the outputs when the [GRPC|GRPC2|P2P] service can not be started:

[rpc_server_failed_to_start.txt](https://github.com/Concordium/concordium-node/files/10071734/rpc_server_failed_to_start.txt)
[grpc2_server_failed_to_start.txt](https://github.com/Concordium/concordium-node/files/10071735/grpc2_server_failed_to_start.txt)
[p2p_server_failed_to_start.txt](https://github.com/Concordium/concordium-node/files/10071736/p2p_server_failed_to_start.txt)

They should also print the relevant error messages.

abizjak:
> 

You have to build the node with the instrumentation feature `--feature=instrumentation`. Then you can start it by setting the following options

```
    #[structopt(
        long = "prometheus-listen-addr",
        help = "IP to listen for prometheus requests on",
        default_value = "127.0.0.1",
        env = "CONCORDIUM_NODE_PROMETHEUS_LISTEN_ADDRESS"
    )]
    pub prometheus_listen_addr:   String,
    #[structopt(
        long = "prometheus-listen-port",
        help = "Port for prometheus to listen on",
        default_value = "9090",
        env = "CONCORDIUM_NODE_PROMETHEUS_LISTEN_PORT"
    )]
    pub prometheus_listen_port:   u16,
    #[structopt(
        long = "prometheus-server",
        help = "Enable prometheus server for metrics",
        env = "CONCORDIUM_NODE_PROMETHEUS_SERVER"
    )]
    pub prometheus_server:        bool,
```

concordium-node/src/rpc.rs
```diff
@@ -61,10 +61,24 @@ impl RpcServerImpl {
...
+            .serve_with_shutdown(self.listen_addr, shutdown_signal)
+            .await
+            .map_err(|e| e.into());
+        // If we are here, either an error occured, or the server was shut down.
+        if let Err(ref err) = result {
+            // Log an error and notify main thread that an error occured.
+            error!("A runtime error occurred in the GRPC2 server: {}", err);
+            error_sender.send(()).expect(
+                "A runtime error occurred in the RPC server, but it was unable to notify the \
+                 owning thread.",
```
abizjak, line 78: Same as above.

concordium-node/src/grpc2.rs
```diff
@@ -983,12 +984,21 @@ pub mod server {
...
+                        error!("A runtime error occurred in the GRPC2 server: {}", err);
+                        error_sender.send(()).expect(
+                            "A runtime error occurred in the GRPC2 server, but it was unable to \
+                             notify the owning thread.",
+                        );
+                    }
+                    result
                 });
                 Ok(Some(Self {
                     task,
```
abizjak, line 1004: I cannot comment on the actual line below in the "shutdown" function, but the output is not the best now.

In particular if the node fails to start because a port is already occupied the output is now

```
2022-11-24T09:18:06.957760033Z: ERROR: A runtime error occurred in the GRPC2 server: transport error
2022-11-24T09:18:06.957812280Z: INFO: Starting the P2P layer
2022-11-24T09:18:06.958250249Z: INFO: Commencing baking
2022-11-24T09:18:06.958310964Z: INFO: Runner: Starting baker thread
2022-11-24T09:18:06.958352063Z: ERROR: Could not stop the GRPC2 server correctly. Forcing shutdown.
```

In particular the last line is misleading.

We should distinguish two kinds of errors here.

Right now we have this channel that is used to signal the main thread to stop. There are two kinds of writers to this channel.

- the signal handler, which indicates a user decided to kill the node.
- A service has failed to start or has crashed. In that case the `shutdown` function should not even be called since the  server is already dead.

One option for handling this is to not have a channel that just sends a (), but instead send either a boolean or more variants.

One variant means the user has requested a shutdown.

Another means that the grpc2 server has been terminated.

abizjak (commented):
Overall it is looking OK, but there are some cases where the handling is not yet where we want it to be.

concordium-node/src/bin/cli.rs
```diff
@@ -129,51 +142,52 @@ async fn main() -> anyhow::Result<()> {
...
+        let consensus = consensus.clone();
+        let import_stopped = Arc::clone(&import_stopped);
+        tokio::spawn(async move {
+            if shutdown_receiver_stop_import.recv().await.is_err() {
+                error!("Shutdown signal channel dropped. This should not happen.");
+            }
+            import_stopped.store(true, atomic::Ordering::Release);
+            consensus.stop_importing_blocks();
+        });
+    }
```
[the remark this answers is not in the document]
vikt0r0: It was moved because it needed the channels to be in scope before. I mean to move it back.

concordium-node/src/bin/cli.rs
```diff
@@ -62,30 +62,9 @@ async fn main() -> anyhow::Result<()> {
         instantiate_node(&conf, &mut app_prefs, stats_export_service, regenesis_arc.clone())
             .context("Failed to create the node.")?;
 
-    #[cfg(feature = "instrumentation")]
```
vikt0r0, line 65: I moved this until after starting the thread in line 129 in the PR. We should perhaps discuss whether this is problematic.

concordium-node/src/bin/cli.rs
```diff
@@ -129,30 +108,99 @@ async fn main() -> anyhow::Result<()> {
...
+                        );
+                        break;
+                    }
+                }
+                // If we are here, either an error occurred or a signal to shut down the node
+                // was made by the user. In any case, we stop the out-of-band-catchup.
+                match import_stopped.compare_exchange(
+                    false,
+                    true,
+                    atomic::Ordering::Acquire,
```
vikt0r0, line 160: I will probably have to look into memory ordering options to see if these options are sensible.

concordium-node/src/common/mod.rs
```diff
@@ -15,3 +15,13 @@ pub use self::{
     p2p_node_id::P2PNodeId,
     p2p_peer::{P2PPeer, PeerStats, PeerType, RemotePeer},
 };
+
```
vikt0r0, line 18: Perhaps this should go somewhere else?

concordium-node/src/grpc2.rs
vikt0r0: Hope it looks better now.

concordium-node/src/bin/cli.rs
```diff
@@ -198,38 +248,42 @@ async fn main() -> anyhow::Result<()> {
...
-        let timeout_duration = std::time::Duration::from_secs(10);
-        match tokio::time::timeout(timeout_duration, task).await {
-            Ok(res) => {
-                if let Err(err) = res {
-                    if err.is_cancelled() {
-                        info!("RPC server was successfully shutdown by force.");
-                    } else if err.is_panic() {
-                        error!("RPC server panicked: {}", err);
+    // Message rpc to shutdown first, if it did not crash.
+    if !rpc_error.load(atomic::Ordering::Acquire) {
```
vikt0r0, line 258: Since the runtime will kill this task when shutting down, should we wait for thread even if it crashed, or do we not care about it once this happens?

concordium-node/src/bin/cli.rs
```diff
@@ -196,38 +200,45 @@ async fn main() -> anyhow::Result<()> {
...
-        // Force the rpc server to shut down in at most 10 seconds.
-        let timeout_duration = std::time::Duration::from_secs(10);
-        match tokio::time::timeout(timeout_duration, task).await {
-            Ok(res) => {
-                if let Err(err) = res {
-                    if err.is_cancelled() {
-                        info!("RPC server was successfully shutdown by force.");
-                    } else if err.is_panic() {
-                        error!("RPC server panicked: {}", err);
+        if !task.is_finished() {
```
abizjak, line 213: It would be useful to add a short comment about why we do this. That is, in what scenarios the task will already be finished.

concordium-node/src/bin/cli.rs
```diff
@@ -196,38 +200,45 @@ async fn main() -> anyhow::Result<()> {
...
+                    warn!("RPC server was forcefully shut down due to: {}", timed_out);
+                }
             }
         }
     }
 
     // Message grpc2 to shutdown if it exists.
     if let Some(rpc2) = rpc2 {
-        rpc2.shutdown().await
+        if !rpc2.is_finished() {
```
abizjak, line 239: Also here.

abizjak (approved):

PR #627, opened 2022-11-29
Make distributions builds support prometheus. Bump version to 5.1.1

abizjak:
## Purpose

- Prepare for stagenet release
- In order to support using the released distributions internally they have to support the prometheus server. Build the node with support for it.

## Changes

- Bump version
- Enable prometheus configuration in all node distributions

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## Unreleased changes
+
+## 5.1.1
+
 - Relay blocks earlier. In particular this means that blocks are now processed in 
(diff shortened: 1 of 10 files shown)


EmilLa1:
The proposed version is not _really_ following semver as we have a breaking change in the node API because of a removed feature.

EmilLa1 (approved):

abizjak:
This is not a library. Our versioning policy is to bump major version on protocol versions. And intermediate releases are minor version changes with stagenet versions using patch versions.

Thahara (approved):

PR #629, opened 2022-12-01
Add timeout to catchup chunk file downloader.

vikt0r0:
## Purpose

Currently the downloading a catch up chunk file specified by an URL has no timeout, so it may hang indefinitely. The purpose of this PR is to add a timeout to the download of the file, s.t. program execution can continue in case the download fails.

## Changes

- [X] Add connection and request timeouts to the HTTP client.
- [X] Add `--download-blocks-timeout` option to specify the the time limit for downloads.

## Checklist

      hard-to-understand areas.


concordium-node/src/bin/cli.rs
```diff
@@ -660,22 +660,38 @@ async fn download_chunk(
...
-        while let Some(Ok(bytes)) = stream.next().await {
-            buffer.write_all(&bytes)?;
-        }
-        buffer.flush()?;
+        // Download the block chunk file.
+        match tokio::time::timeout(timeout_duration, http_client.get(download_url.clone()).send())
+            .await
+        {
+            Err(timed_out) => {
+                error!(
```
vikt0r0, line 671: In some places, we use the `error!`, and in others we use `warn!` macro. I am not quite sure about the semantics, however.

concordium-node/src/bin/cli.rs
```diff
@@ -660,22 +660,38 @@ async fn download_chunk(
...
-        let chunk_response = http_client.get(download_url.clone()).send().await?;
-        anyhow::ensure!(
-            chunk_response.status().is_success(),
-            "Unable to download the block chunk file from {}: {} {}",
-            download_url,
-            chunk_response.status().as_str(),
-            chunk_response.status().canonical_reason().unwrap()
-        );
+        // Use a 5 minute timeout for downloading the block chunk file.
+        let timeout_duration = std::time::Duration::from_secs(300);
```
vikt0r0, line 664: How long should the timeout be?

vikt0r0:
Should I update the changelog?

concordium-node/src/bin/cli.rs
```diff
@@ -660,22 +660,38 @@ async fn download_chunk(
...
 
-        let file = temp_file.as_file_mut();
-        let mut buffer = std::io::BufWriter::new(file);
-        let mut stream = chunk_response.bytes_stream();
-        while let Some(Ok(bytes)) = stream.next().await {
-            buffer.write_all(&bytes)?;
-        }
-        buffer.flush()?;
+        // Download the block chunk file.
+        match tokio::time::timeout(timeout_duration, http_client.get(download_url.clone()).send())
```
abizjak, line 667: I don't think we should be using this. Reqwest supports timeouts directly

https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout for the connection one (which we should set to a low number of seconds)

And this https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.timeout for the request timeout. This needs to be a reasonably large, like the current timeout you set.

concordium-node/src/bin/cli.rs
abizjak: I think in this case both could apply. Error should usually be where the required action could not be completed and may require user intervention.

Warn is similar, but the service has recovered from the issue automatically.

In this case you could argue that both apply.

concordium-node/src/bin/cli.rs
abizjak: I think for the second option it would be ideal to add a configuration option with a default of 5min. THat'll also make it simple to test.

abizjak:
> Should I update the changelog?

Yes please.

abizjak (approved):
Looks good and works in my tests.

I have two minor comments. After resolving those please merge.

PR #631, opened 2022-12-02
Remove the custom metrics implementation. Always use prometheus.

abizjak:
This removes the instrumentation feature.

## Purpose

Enable a single build of the node that supports a prometheus server if so configured.

## Changes

Remove the instrumentation feature, and the corresponding conditional compilation and custom metrics implementation.

The prometheus server is only started if the user requests it. Otherwise the metrics are only used through the GRPC interface.

## Checklist

      hard-to-understand areas.


EmilLa1:
Why is `base` changed as part of this PR?

concordium-node/src/bin/cli.rs
```diff
@@ -62,25 +60,25 @@ async fn main() -> anyhow::Result<()> {
...
-        tokio::spawn(async move { stats.start_server(SocketAddr::new(pla, plp)).await });
+        if let Some(plp) = conf.prometheus.prometheus_listen_port {
+            let stats = node.stats.clone();
+            let pla = conf
+                .prometheus
+                .prometheus_listen_addr
+                .parse::<IpAddr>()
+                .context("Invalid Prometheus address")?;
+            tokio::spawn(async move { stats.start_server(SocketAddr::new(pla, plp)).await });
+        }
```
EmilLa1, line 76: Should we write a log stmt. that the prometheus server started?

concordium-node/src/dumper.rs
```diff
@@ -1,6 +1,6 @@
 //! Handles the `network_dump` feature.
 
-cfg_if! {
+cfg_if::cfg_if! {
```
EmilLa1, line 3: Is this intentional?

EmilLa1 (commented):
Looks good! 

A CHANGELOG entry is currently missing. 

I'm not sure why the ptrs to `concordium-base´ and `concordium-grpc-api´ are changed either? It seems unrelated to this PR? Is it at least on purpose? :)

concordium-node/src/dumper.rs
abizjak: Yes.

abizjak: > Looks good!
> 
> A CHANGELOG entry is currently missing.
> 
> I'm not sure why the ptrs to `concordium-base´ and `concordium-grpc-api´ are changed either? It seems unrelated to this PR? Is it at least on purpose? :)

Yes. Because they were not on main. And nothing that the node uses there has changed.

concordium-node/src/bin/cli.rs
abizjak: We can't really print it started. We can print it is starting.

abizjak: I added a log statement.

concordium-node/src/dumper.rs
abizjak: > Is this intentional?

And the reason for the change is that this is the only remaining place where `cfg_if` is used. And as part of the cleanup related to the instrumentation flag I also removed this

```

#[macro_use]
extern crate cfg_if;

```

Which used to be needed (the macro_use) in older versions of Rust, but is no longer the way to do it for some years.

EmilLa1 (approved):

PR #634, opened 2022-12-07
Use `select` instead of condition variables for message queue handling.

td202:
## Purpose

This is an alternative solution to #632. That addressed the issue of missed wake-ups in the message queue handlers by setting a timeout (100ms) on waiting on the condition variable. While this handles the issue, it does not solve the underlying problem: missed wake-ups can still happen, but this will just lead to (small) delays. Moreover, it causes many spurious wake-ups, which is generally undesirable.

This change removes the condition variable altogether and instead waits on the message queues using [`select`](https://docs.rs/crossbeam/latest/crossbeam/macro.select.html), which allows waiting for a message on either queue.

## Changes

- Use `select` in the message handler loops to wait for new messages.
- Remove the condition variables previously used.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/Cargo.lock
+++ b/concordium-node/Cargo.lock
@@ -601,7 +601,6 @@ dependencies = [
  "mio 0.7.14",
  "nohash-hasher",
  "noiseexplorer_xx",
- "parking_lot",
  "preferences",
  "prometheus",
  "prost",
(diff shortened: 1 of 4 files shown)


abizjak (commented):
I think parking_lot is no longer needed after these changes.

abizjak (approved):

EmilLa1 (approved):
Good job!
It looks like that the two threads have code duplication, whether it's worth refactoring now, I'll let you be the one to decide.

EmilLa1 (commented):
Also shouldn't the CHANGELOG be updated?
I.e. "_missed wake-ups can still happen, but this will just lead to (small) delays_"

PR #635, opened 2022-12-07
Database exporter issues

vikt0r0:
## Purpose

The database exporter currently has a few issues, primarily regarding usability. It could benefit from improving on a few things, notably:

- When the user provides an invalid database directory, changes are made to the filesystem anyways. User-specified parameters should be checked and invoking the exporter should be idempotent in case wrong parameters were provided.
- Parsing relies on `!!` and `read` in which case the exporter crashes with error messages that do not convey any useful information to the user.
- The exporter relies on genesis indices and block heights to decide which blocks to export, however it does not use the genesis hash. Specifically this means that it is possibly to export a database corresponding to one chain "on top of" an exported database from another chain.
- The block index file becomes messy over time, with duplicate entries. This works, but it looks ugly and is perhaps error-prone when compared to using a parser and accompanied with some intermediate representation of the file, as the following snippet shows:
```
...
blocks-2-1300001.dat,2,1300001,1383656
# genesis hash ea4a52a04ba905c2692b4598aa344707327781d588573d374125b449a1ce0bcc
blocks-3-1.dat,3,1,100000
blocks-3-100001.dat,3,100001,200000
blocks-3-200001.dat,3,200001,300000
blocks-3-300001.dat,3,300001,400000
blocks-3-400001.dat,3,400001,500000
blocks-3-500001.dat,3,500001,600000
blocks-3-600001.dat,3,600001,700000
blocks-3-700001.dat,3,700001,800000
# genesis hash ea4a52a04ba905c2692b4598aa344707327781d588573d374125b449a1ce0bcc
# genesis hash ea4a52a04ba905c2692b4598aa344707327781d588573d374125b449a1ce0bcc
# genesis hash ea4a52a04ba905c2692b4598aa344707327781d588573d374125b449a1ce0bcc
...
```
It would be nicer to have it look like:
```
...
blocks-2-1300001.dat,2,1300001,1383656
# genesis hash ea4a52a04ba905c2692b4598aa344707327781d588573d374125b449a1ce0bcc
blocks-3-1.dat,3,1,100000
blocks-3-100001.dat,3,100001,200000
blocks-3-200001.dat,3,200001,300000
blocks-3-300001.dat,3,300001,400000
blocks-3-400001.dat,3,400001,500000
blocks-3-500001.dat,3,500001,600000
blocks-3-600001.dat,3,600001,700000
blocks-3-700001.dat,3,700001,800000
...
```
- When a chunk is replaced by a new chunk in the block index, the new chunk is versioned and its filename is updated in the block index file. This leaves the old chunk file in the file system. It would be useful to include a flag indicating whether to use versioned chunk files or to simply overwrite existing chunks when applicable.
- If an error occurs, a line is removed from the `blocks.idx` file, but it is not reinserted. This block contains the finalization records of the database. When this happens, in some cases it seems that the chunk files listed in the block index file no longer corresponds to the format of the exported database. This can be currently fixed by performing a subsequent (successful) run of the exporter, but it would be better to have the exporter recover the database to an intact state when this happens.

## Changes

- [x] Add datatype for representing the block index file, `blocks.idx`.
- [x] Add `attoparsec` parser for parsing the block index file. The choice of library was made because it is a transitive dependency also used by `aeson`.
- [x] Add functionality for writing the block index representation to a string representation corresponding to the contents of `blocks.idx`
- [x] Add functionality for cleaning up existing, messy block index files output from previous versions of the database exporter.
- [x] Ensure that block index output files do not contain duplicate or dangling block hash section "headers". 
- [x] Add flag for indicating whether chunks should be overwritten or versioned copies should be made.
- [x] Add functionality for restoring the database to an intact state in case of an error.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -49,7 +49,7 @@ on:
   workflow_dispatch: # allow manual trigger
 
 env:
-  dummy: 8                 # change to force cache invalidation
+  dummy: 9                 # change to force cache invalidation
   CARGO_TERM_COLOR: always # implicitly adds '--color=always' to all cargo commands
   TEST_LEVEL: 1            # for stack tests
 
--- a/concordium-consensus/src/Concordium/ImportExport.hs
+++ b/concordium-consensus/src/Concordium/ImportExport.hs
@@ -39,10 +39,22 @@
 module Concordium.ImportExport where
 
 import Control.Monad
+import Control.Monad.Catch
+import Control.Monad.IO.Class
+import Control.Monad.State (MonadState, evalStateT)
 import Control.Monad.Trans.Except
+import qualified Data.Attoparsec.Text as AP
+import Data.Bits
 import qualified Data.ByteString as BS
-import Data.List.Split
+import Data.Char (isHexDigit)
+import Data.Sequence (
+    Seq (..),
+    fromList,
+    singleton,
+    (><),
+ )
 import Data.Serialize
+import qualified Data.Text as T
 import Data.Word
 import System.Directory
 import System.FilePath
@@ -56,10 +68,6 @@ import Concordium.GlobalState.Persistent.LMDB
 import Concordium.Logger
 import Concordium.Types
 import Concordium.Utils.Serialization.Put
-import Control.Monad.Catch
-import Control.Monad.IO.Class
-import Control.Monad.State (MonadState, evalStateT)
-import Data.Bits
 import Lens.Micro.Platform
 
 -- |State used for exporting the database.
@@ -123,23 +131,150 @@ initialHandle p = do
     createDirectoryIfMissing True (takeDirectory p)
     chunkExists <- doesFileExist p
     if chunkExists
-        then initialHandle (replaceExtensions p newExt)
+        then do
+            newExt <- getNewExt
+            initialHandle (replaceExtensions p newExt)
         else do
             hdl <- openBinaryFile p WriteMode
             return (p, hdl)
   where
+    -- The old, potentially unversioned, file extension.
+    -- For instance
+    --   `oldExt blocks-2-131073.6.dat == ".6.dat"` and
+    --   `oldExt blocks-2-131073.dat == ".dat"`.
+    oldExt :: String
     oldExt = takeExtensions p
-    newExt =
+    -- The old file extension with an added or incremented version
+    -- number. For instance
+    --   if `p == "file.6.dat"` then `newExt == return ".7.dat"` and
+    --   if `p == "file.dat"`, then `newExt == return ".2.dat"`.
+    getNewExt :: IO String
+    getNewExt =
+        -- If the old extension is a single extension, i.e. with no
+        -- version number, we add one.
         if oldExt == takeExtension p
-            then ".2" ++ oldExt
-            else
-                (\x -> "." ++ x ++ ".dat")
-                    . show
-                    . (+ (1 :: Int))
-                    . read
-                    . (!! 1)
-                    . splitOn "."
-                    $ oldExt
+            then return $ ".2" ++ oldExt
+            else case getVersionNumber (T.pack oldExt) of
+                Left err ->
+                    fail $
+                        "Unable to parse a version number from extension '"
+                            <> oldExt
+                            <> "': "
+                            <> err
+                Right v -> return $ "." <> show (v + 1) <> ".dat"
+    -- Given a file extension, attempt to get its version number.
+    -- Returns @Left@ if a version number is not present and @Right@
+    -- otherwise.
+    getVersionNumber :: T.Text -> Either String Integer
+    getVersionNumber = AP.parseOnly $ do
+        _ <- AP.char '.'
+        AP.decimal
+
+-- |Data type used to represent a line with chunk information in the block index file.
+-- A chunk contains exported data for all blocks of height in the range `blockHeightFirst`
+-- to `blockHeightLast` and of genesis index `genesisIndex`. When a chunk is exported, a
+-- line with the above information and the filename of the chunk is added to the block
+-- index file.
+data BlockIndexChunkInfo = BlockIndexChunkInfo
+    { filename :: T.Text, -- Name of the chunk file.
+      genesisIndex :: GenesisIndex, -- Genesis index of the blocks contained in the chunk.
+      blockHeightFirst :: BlockHeight, -- Height of the first block contained in the chunk.
+      blockHeightLast :: BlockHeight -- Height of the last block contained in the chunk.
+    }
+    deriving (Show)
+
+-- |Data type used to represent the contents of a block index file.
+-- The block index file contains an index of the exported chunks containing the block and
+-- and finalization record data of the exported database. The index consists of a number of
+-- sections each comprising a section header line that specifies a genesis blockhash followed
+-- by a sequence of lines that each represent a chunk containing blocks with that genesis
+-- blockhash. A section with genesis blockhash corresponding to a genesis index `i` follows
+-- another section with genesis blockhash corresponding to genesis index `j` iff. `i == j`
+-- or `i == j+1`.
+type BlockIndex = Seq (BlockHash, Seq BlockIndexChunkInfo)
+
+-- |Parse a blockhash.
+parseBlockHash :: AP.Parser BlockHash
+parseBlockHash = do
+    hash <- AP.count 64 $ AP.satisfy isHexDigit
+    return $! BlockHash (read hash)
+
+-- |Parse a line of the block index file containing block genesis hash and indicating the start
+-- of a new section.
+parseGenesisDataLine :: AP.Parser BlockHash
+parseGenesisDataLine = do
+    _ <- AP.string (T.pack "# genesis hash ")
+    hash <- parseBlockHash
+    AP.skip AP.isEndOfLine
+    return hash
+
+-- |Parse a line of the block index file containing chunk information of a section.
+parseChunkLine :: AP.Parser BlockIndexChunkInfo
+parseChunkLine = do
+    filename <- AP.takeWhile1 (AP.inClass ".a-zA-Z0-9-")
+    _ <- AP.char ','
+    genesisIndex <- AP.decimal
+    _ <- AP.char ','
+    blockHeightStart <- AP.decimal
+    _ <- AP.char ','
+    blockHeightEnd <- AP.decimal
+    AP.skip AP.isEndOfLine
+    return $ BlockIndexChunkInfo filename genesisIndex blockHeightStart blockHeightEnd
+
+-- | Parse all sections of a block index file.
+parseBlockIndexFile :: AP.Parser BlockIndex
+parseBlockIndexFile = do
+    list <-
+        AP.manyTill
+            ( do
+                genesisData <- parseGenesisDataLine
+                chunks <- AP.many' parseChunkLine
+                return (genesisData, fromList chunks)
+            )
+            AP.endOfInput
+    return $ fromList list
+
+-- |Show the contents of a block index file. The resulting string can be
+-- parsed back into its corresponding @BlockIndex@ using `parseBlockIndexFile`.
+showBlockIndexFile :: BlockIndex -> String
+showBlockIndexFile = concatMap showSection
+  where
+    showSection :: (BlockHash, Seq BlockIndexChunkInfo) -> String
+    showSection (gd, chunks) =
+        showGenesisDataLine gd
+            <> concatMap showChunkLine chunks
+    showGenesisDataLine :: BlockHash -> String
+    showGenesisDataLine gd =
+        "# genesis hash "
+            <> show gd
+            <> "\n"
+    showChunkLine :: BlockIndexChunkInfo -> String
+    showChunkLine (BlockIndexChunkInfo{..}) =
+        T.unpack filename
+            <> ","
+            <> show genesisIndex
+            <> ","
+            <> show blockHeightFirst
+            <> ","
+            <> show blockHeightLast
+            <> "\n"
+
+-- |Normalize the block index.
+-- Eliminates redundant sections of a `BlockIndex` by 1) merging consecutive
+-- sections that have the same genesis block hash and by 2) removing sections
+-- with no chunks. This is useful since older versions of the database exporter
+-- may insert empty sections or consecutive sections with the same genesis hash.
+-- This is due to `exportSections` returning a "tail" of a block index
+-- corresponding to the sections it exported. The last section of the previously
+-- exported block index will therefore match that of the first section of the tail,
+-- so appending the tail to produce the updated block index will introduce some
+-- undesired clutter and redundancy.
+normalizeBlockIndex :: BlockIndex -> BlockIndex
+normalizeBlockIndex ((_, Empty) :<| l) = normalizeBlockIndex l
+normalizeBlockIndex ((gh1, chunks1) :<| (gh2, chunks2) :<| r)
+    | gh1 == gh2 = normalizeBlockIndex ((gh1, chunks1 >< chunks2) :<| r)
+    | otherwise = (gh1, chunks1) :<| normalizeBlockIndex ((gh2, chunks2) :<| r)
+normalizeBlockIndex l = l
 
 -- |Export a database in V3 format, as a collection of block file chunks, given the data directory
 -- root and the export directory root.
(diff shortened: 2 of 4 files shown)


td202 (commented):
Looks good. Just a few comments.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -39,10 +39,18 @@
...
+import Control.Monad.IO.Class
+import Control.Monad.State (MonadState, evalStateT)
 import Control.Monad.Trans.Except
+import qualified Data.Attoparsec.Text as AP
+import Data.Bits
 import qualified Data.ByteString as BS
-import Data.List.Split
+import Data.Char (isHexDigit)
+import Data.Maybe (fromMaybe)
+import Data.Sequence
```
td202, line 51: I recommend importing qualified, or giving an import list of `Sequence`, since there are many names that collide with `Prelude`.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +254,185 @@ exportDatabaseV3 ::
...
+                                                            -- nothing was previously exported for the
+                                                            -- section corresponding to the current
+                                                            -- genesis index.
+                                                            Empty -> genHash
+                                                            -- blocks were previously exported for the
+                                                            -- current genesis index, so use the genesis
+                                                            -- block hash that was previously exported
+                                                            -- for that section and compare it with that
+                                                            -- of the database.
+                                                            bi -> fst . sLast $ bi
```
td202, line 388:
```suggestion
                                                            _ :|> (gh, _) -> gh
```

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +254,185 @@ exportDatabaseV3 ::
...
+                            when overwriteExistingChunks $ renameFile lastChunkFile $ lastChunkFile <> "~"
+                            -- evict the last chunk from the block index since we want to
+                            -- export blocks starting from the genesis index and the height
+                            -- of the first block contained within it
+                            return
+                                ( genesisIndex,
+                                  blockHeightFirst,
+                                  sInit blockIndex :|> (gd, sInit chunks),
+                                  Just lastChunkFile
+                                )
```
td202, line 305:
```suggestion
                    _ :|> (gd, chunks) -> case chunks of
                        initChunks :|> BlockIndexChunkInfo{..} -> do
                            -- the filename of the most recently exported chunk
                            let lastChunkFile = outDir </> T.unpack filename
                            -- if we overwrite files, back up the most recently exported
                            -- chunk file so we can restore it if something goes wrong.
                            when overwriteExistingChunks $ renameFile lastChunkFile $ lastChunkFile <> "~"
                            -- evict the last chunk from the block index since we want to
                            -- export blocks starting from the genesis index and the height
                            -- of the first block contained within it
                            return
                                ( genesisIndex,
                                  blockHeightFirst,
                                  sInit blockIndex :|> (gd, initChunks),
                                  Just lastChunkFile
                                )
```
We should also have an `Empty` case. If normalization guarantees no empty chunks, then this can just be an error. However, I'm not sure that's currently the case.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +120,130 @@ sectionHeaderLength :: Word64
...
+normalizeBlockIndex l = l
+
+-- |Get the last element of a list.
+-- Fails with an error if the list is empty.
+sLast :: Seq a -> a
+sLast s = index s (Data.Sequence.length s - 1)
+
+-- |`init` for sequences.
+sInit :: Seq a -> Seq a
+sInit s = deleteAt (Data.Sequence.length s - 1) s
```
td202, line 246: We can always use `:|>` instead of these.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
-    bracket
-        (openFile indexFile AppendMode)
-        hClose
-        (\indexHdl -> exportSections dbDir outDir indexHdl chunkSize genIndex startHeight)
-
--- |Export a database section for each genesis index.
+                -- we export blocks starting from height 1 because genesis blocks need not be exported.
+                return (0, 1, Empty, Nothing)
+
+    let lastChunkFileName = fromMaybe "" lastChunkFilenameM
```
abizjak, line 309: Why is this sensible? The empty string is generally not a valid file name. At the very least mention why this is sensible.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
+    let lastChunkBackupName = lastChunkFileName <> "~"
+    lastChunkBackupExists <- doesFileExist lastChunkBackupName
+
+    blockIndexTail <- exportSections dbDir outDir chunkSize overwriteExistingChunks genIndex startHeight blockIndex
+    case blockIndexTail of
+        -- in case nothing was was exported, restore the most recently
```
abizjak, line 315:
```suggestion
        -- in case nothing was exported, restore the most recently
```

concordium-consensus/tools/database-exporter/DatabaseExporter/CommandLineParser.hs
```diff
@@ -38,6 +39,18 @@ config =
...
+                                        ( long "overwrite-chunks"
+                                            <> value True
+                                            <> showDefault
+                                            <> metavar "BOOL"
+                                            <> help
+                                                ( "Overwrite chunk files if they already exist; "
+                                                    <> "otherwise choose an available filename by adding "
+                                                    <> "a version number to the extension of the chunk file."
+                                                )
+                                        )
```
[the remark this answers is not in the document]
abizjak: This option is also not documented in the README.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
-        then initialHandle (replaceExtensions p newExt)
+    if chunkExists && not overwriteExistingChunk
+        then do
+            newExt <- getNewExt
+            initialHandle overwriteExistingChunk (replaceExtensions p newExt)
         else do
             hdl <- openBinaryFile p WriteMode
             return (p, hdl)
   where
+    -- The old, potentially unversionsioned, file extension
```
abizjak, line 143:
```suggestion
    -- The old, potentially unversioned, file extension
```

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+    oldExt :: String
     oldExt = takeExtensions p
-    newExt =
+    -- The next versioned file extension
+    getNewExt :: IO String
+    getNewExt =
         if oldExt == takeExtension p
```
abizjak, line 149: Since takeExtension and takeExtension*s* are so similar, and the behaviour is very hacky, this warrants documentation.

abizjak: Is the real check here that there is only a single extension? Then that should probably be how it is done, or at least stated.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+            then return $ ".2" ++ oldExt
+            else case tryGetVersionNumber (T.pack oldExt) of
+                Left err -> fail err
+                Right v -> return $ "." <> show (v + 1) <> ".dat"
+    -- Given a file name attempt to get its version number
+    tryGetVersionNumber :: T.Text -> Either String Integer
+    tryGetVersionNumber = AP.parseOnly $ do
+        AP.skipMany $ AP.skip (AP.inClass "a-zA-Z0-9-")
+        _ <- AP.char '.'
+        AP.decimal
```
abizjak, line 159: Please document the assumptions of this function. What does "attempt" mean.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+-- |Data type used to represent a line with genesis block hash in the block index file.
+type BlockIndexGenesisHash = BlockHash
+
+-- |Data type used to represent a line with chunk information in the block index file.
+data BlockIndexChunkInfo = BlockIndexChunkInfo
+    { filename :: T.Text,
+      genesisIndex :: GenesisIndex,
+      blockHeightFirst :: BlockHeight,
+      blockHeightLast :: BlockHeight
+    }
```
abizjak, line 170: Document, are the min/max inclusive or not?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+                Right v -> return $ "." <> show (v + 1) <> ".dat"
+    -- Given a file name attempt to get its version number
+    tryGetVersionNumber :: T.Text -> Either String Integer
+    tryGetVersionNumber = AP.parseOnly $ do
+        AP.skipMany $ AP.skip (AP.inClass "a-zA-Z0-9-")
+        _ <- AP.char '.'
+        AP.decimal
+
+-- |Data type used to represent a line with genesis block hash in the block index file.
+type BlockIndexGenesisHash = BlockHash
```
abizjak, line 162: I don't really see the purpose of this type alias. It confused me more than it provided value.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+-- |Data type used to represent a line with chunk information in the block index file.
+data BlockIndexChunkInfo = BlockIndexChunkInfo
+    { filename :: T.Text,
+      genesisIndex :: GenesisIndex,
+      blockHeightFirst :: BlockHeight,
+      blockHeightLast :: BlockHeight
+    }
+    deriving (Show)
+
+-- |Data type used to represent the block index file.
```
abizjak, line 173: This documenation is not very useful. It would be better to state the intent. 

"An index file is a list of lists of chunks. Each list of chunks is associated with a hash of the genesis to which all the blocks belong to."

Or some such.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+    if not atEnd
+        then do
+            list <-
+                AP.manyTill
+                    ( do
+                        genesisData <- parseGenesisDataLine
+                        chunks <- AP.many' parseChunkLine
+                        return (genesisData, fromList chunks)
+                    )
+                    AP.endOfInput
```
abizjak, line 210: Which case does just
```haskell

                AP.manyTill
                    ( do
                        genesisData <- parseGenesisDataLine
                        chunks <- AP.many' parseChunkLine
                        return (genesisData, fromList chunks)
                    )
                    AP.endOfInput
```
not handle? Why do you have to check whether we are at the end of the stream manually ?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+            <> ","
+            <> show blockHeightFirst
+            <> ","
+            <> show blockHeightLast
+            <> "\n"
+
+-- |Normalize the block index by eliminating redundant genesis hash lines.
+-- This is necessary since older versions of the database exporter may
+-- insert superfluous (i.e. repeated) lines containing the same genesis
+-- hashes.
```
abizjak, line 237: I think the comment here does not really reflect how this function is used. It is used upon output.

If only the input is the problem then the parser could already handle this to always return a normalized block index. Why is that not enough?

abizjak: Specifically this is used in output here https://github.com/Concordium/concordium-node/pull/635/files#diff-64912044dc5a6fbbce28889ab7e8ab320e77a1dabaa2246ca15cc18b853abf81R322

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
-                return (lastChunkGenesisIndex, lastChunkFirstBlock)
+                blockIndex <- case AP.parseOnly parseBlockIndexFile (T.pack indexContents) of
+                    Left err -> fail $ "An error occurred while parsing '" <> indexFile <> "': " <> err
+                    -- normalize the block index, s.t. superfluous genesis hash lines are removed.
+                    Right bi -> do
+                        return $ normalizeBlockIndex bi
+                case blockIndex of
+                    -- we export blocks starting from height 1 because genesis blocks need not be exported.
+                    Empty -> return (0, 1, Empty, Nothing)
+                    -- the following is safe due to normalization of the blockindex.
```
abizjak, line 286: I don't see what this is referring to. I don't see anything unsafe here at all.

If you are referring to the "fail" below, then the comment should be placed there, and be more to the point "normalized block index does not contain an empty list of chunks for any genesis index" or some such.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
     -- |Database directory
     FilePath ->
     -- |Export directory
     FilePath ->
-    -- |Index file handle
-    Handle ->
     -- |Chunk size in blocks
     Word64 ->
+    -- |Overwrite existing chunks
+    Bool ->
```
abizjak, line 337: Why was this added?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
+                    Empty -> return (0, 1, Empty, Nothing)
+                    -- the following is safe due to normalization of the blockindex.
+                    blockIndexInit :|> (gd, chunks) -> case chunks of
+                        Empty -> fail $ "A section for genesis hash '" <> show gd <> "' in block index file has no entries."
+                        initChunks :|> BlockIndexChunkInfo{..} -> do
+                            -- the filename of the most recently exported chunk
+                            let lastChunkFile = outDir </> T.unpack filename
+                            -- if we overwrite files, back up the most recently exported
+                            -- chunk file so we can restore it if something goes wrong.
+                            when overwriteExistingChunks $ renameFile lastChunkFile $ lastChunkFile <> "~"
```
abizjak, line 294:
```suggestion
                            when overwriteExistingChunks $ renameFile lastChunkFile $ lastChunkFileBackupName
```

At least don't duplicate the "backup file name convention".

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
+    case blockIndexTail of
+        -- in case nothing was was exported, restore the most recently
+        -- exported chunk s.t. the files and block index remain unchanged,
+        -- since this chunk was evicted from the index in the above.
+        Empty -> when lastChunkBackupExists $ renameFile lastChunkBackupName lastChunkFileName
+        -- otherwise write the updated block index file to the filesystem
+        -- and delete the backup file.
+        _ -> do
+            writeFile indexFile $ showBlockIndexFile $ normalizeBlockIndex $ blockIndex <> blockIndexTail
+            when lastChunkBackupExists $ removeFile lastChunkBackupName
```
abizjak, line 323: I guess technically these backup files could exist without this program creating them. So these checks have some caveats about them. Maybe mention that we assume that the tool is the only one writing to the supplied directory.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -116,30 +125,122 @@ sectionHeaderLength :: Word64
...
+-- boolean flag indicates whether the file should be overwritten or whether a new name should be
+-- chosen by adding to or incrementing a version number in its extension. It is expected that an
+-- unversioned filename only has a single extension.
+initialHandle :: Bool -> FilePath -> IO (FilePath, Handle)
+initialHandle overwriteExistingChunk p = do
     createDirectoryIfMissing True (takeDirectory p)
     chunkExists <- doesFileExist p
-    if chunkExists
-        then initialHandle (replaceExtensions p newExt)
+    if chunkExists && not overwriteExistingChunk
```
abizjak, line 135: Why make this more complicated than it needs to be with an additional option?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
+                    -- we export blocks starting from height 1 because genesis blocks need not be exported.
+                    Empty -> return (0, 1, Empty, Nothing)
+                    -- the following is safe due to normalization of the blockindex.
+                    blockIndexInit :|> (gd, chunks) -> case chunks of
+                        Empty -> fail $ "A section for genesis hash '" <> show gd <> "' in block index file has no entries."
+                        initChunks :|> BlockIndexChunkInfo{..} -> do
+                            -- the filename of the most recently exported chunk
+                            let lastChunkFile = outDir </> T.unpack filename
+                            -- if we overwrite files, back up the most recently exported
+                            -- chunk file so we can restore it if something goes wrong.
```
abizjak, line 293: I am not sure that this comment is particularly useful. If something goes wrong this program will likely crash, in which case the file will not be restored.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -150,90 +251,186 @@ exportDatabaseV3 ::
...
+                                                                    chunkSize
+                                                                    overwriteExistingChunks
+                                                            return (False, singleton (genHash, chunks))
+                                        )
+                                        (DBState dbh 0)
                     closeDatabase dbh
-                    exportSections dbDir outDir indexHdl chunkSize (genIndex + 1) 1
-        else putStrLn ("The tree state database does not exist at " ++ treeStateDir)
+                    if exportError
+                        then -- if an error occurred, return
```
abizjak, line 415: "return only what we managed to export until now" or some such. It is clear that there is a return.

abizjak (changes requested):

concordium-consensus/src/Concordium/ImportExport.hs
abizjak: I do not see that in the issue description. What I do see is that there is a **suggestion** that the old chunk is deleted in the last item. That suggestion is a good one, and should be implemented. But for simplicity it should just be done, we should not have options unless we need them.

This is also very different than overwriting an existing file. It is essential that we do not use the same names because these files are cached.

concordium-consensus/src/Concordium/ImportExport.hs
abizjak: See my comment above.

concordium-consensus/tools/database-exporter/DatabaseExporter/CommandLineParser.hs
```diff
@@ -38,6 +39,18 @@ config =
...
+                                        ( long "overwrite-chunks"
+                                            <> value False
+                                            <> showDefault
+                                            <> metavar "BOOL"
+                                            <> help
+                                                ( "Overwrite chunk files if they already exist; "
+                                                    <> "otherwise choose an available filename by adding "
+                                                    <> "a version number to the extension of the chunk file."
+                                                )
+                                        )
```
abizjak, line 53: I think this option should be removed in-line with the comment I have above. That would also make it in-line with the issue description.

vikt0r0:
Not sure why the CI build step suddenly fails in the linking phase? There should not be any breaking changes introduced since the last successful check?

abizjak:
> Not sure why the CI build step suddenly fails in the linking phase? There should not be any breaking changes introduced since the last successful check?

That is commonly due to caching issues. Try to invalidate the cache in the .github/workflows/build-test.yaml file (by bumping dummy)

.github/workflows/build-test.yaml
```diff
@@ -49,7 +49,7 @@ on:
   workflow_dispatch: # allow manual trigger
 
 env:
-  dummy: 8                 # change to force cache invalidation
+  dummy: 6                 # change to force cache invalidation
```
abizjak, line 52: Can we try to just increase this number all the time please.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -153,87 +288,202 @@ exportDatabaseV3 ::
...
                     closeDatabase dbh
-                    exportSections dbDir outDir indexHdl chunkSize (genIndex + 1) 1
-        else putStrLn ("The tree state database does not exist at " ++ treeStateDir)
+                    -- if an error occurred, return sections
+                    -- that were succesfully written to the
+                    -- file system; otherwise export the section
+                    -- corresponding to the incremented genesis
+                    -- index
+                    if exportError
+                        then return sectionData
```
abizjak, line 464: It would be good that the program terminates with non-zero status code if an error occurs.

That way they can more easily monitor it. So this error would ideally be propagated outwards to `main`.

abizjak (approved):

.github/workflows/build-test.yaml
vikt0r0: Is it OK if I change it to `# increment to force cache invalidation`?

PR #649, opened 2022-12-14
Upgrade compiler tooling and libraries.

vikt0r0:
## Purpose

Upgrade stackage snapshot to `lts-20.4` and ghc to 9.2.5. Upgrade `flatbuffers` to `v22.10.16` as the currently used version has a security advisory.

## Changes

- [X] Bump `resolver`in `stack.yaml` and fix build errors and warnings.
- [x] Bump `flatbuffers` to `v22.10.16` and fix build errors and warnings.

## Depends on

https://github.com/Concordium/concordium-base/pull/283

## Checklist

      hard-to-understand areas.


vikt0r0:
Seems like the CI build step fails since the lastest flatbuffers on snap is. 2.0.0. I'll have a look at updating the script tomorrow.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/EncryptedAmount.hs
```diff
@@ -187,32 +187,35 @@ addIncomingEncryptedAmount !newAmount old = do
...
                         old
                             { _incomingEncryptedAmounts = rest Seq.|> newAmountRef,
-                              _aggregatedAmount = Just (xPlusY, 2),
+                              _aggregatedAmount = Just $!! (xPlusY, n + 1),
                               _startIndex = _startIndex old + 1
                             }
-                else return $ old{_incomingEncryptedAmounts = _incomingEncryptedAmounts old Seq.|> newAmountRef}
-        Just (e, n) -> do
-            -- we have to aggregate always
-            -- irrefutable because of check above
```
vikt0r0, line 205: Is this really irrefutable? Looking at the commit history, it seems that some of it was moved outside the check. I blindly added the `error` to the case which preserves the semantics and removes the warning, but maybe it is worth investigating.

concordium-consensus/src/Concordium/GlobalState/Account.hs
```diff
@@ -131,24 +131,28 @@ addIncomingEncryptedAmount newAmount old =
...
                             { _incomingEncryptedAmounts = rest Seq.|> newAmount,
                               _startIndex = _startIndex old + 1,
                               _aggregatedAmount = Just (x <> y, 2)
                             }
+                    -- this does not happen due to the check above
+                    _ -> error "_incomingEncryptedAmounts should consist of two or more elements"
                 else old & incomingEncryptedAmounts %~ (Seq.|> newAmount)
         Just (e, n) ->
             -- we have to aggregate always
-            -- irrefutable because of check above
```
vikt0r0, line 145: Same as my other comment above:

Is this really irrefutable? Looking at the commit history, it seems that some of it was moved outside the check. I blindly added the `error` to the case which preserves the semantics and removes the warning, but maybe it is worth investigating.

concordium-node/Cargo.toml
```diff
@@ -51,7 +51,7 @@ noiseexplorer_xx = { path = "../deps/external/noiseexplorer_xx" }
 bytesize = "1.0"
 crossbeam-channel = "0.5"
 semver = { version = "1", features = ["serde"] }
-flatbuffers = { version = "2" }
+flatbuffers = { version = "2.1.2" }
```
vikt0r0, line 54: Settled on flatbuffers crate version 2.1.2 since any newer version causes a lot of build errors. Should I look into upgrading it to the latest version or is 2.1.2 fine for now?

Also, I could only find the following security advisories, they both seem to be fixed in >= 2.0.0, so am I overlooking the security advisory we are upgrading to resolve?

https://rustsec.org/advisories/RUSTSEC-2020-0009.htm
https://rustsec.org/advisories/RUSTSEC-2019-0028.html

concordium-consensus/tests/consensus/ConcordiumTests/Update.hs
```diff
@@ -164,7 +168,11 @@ bake bid n = do
     mb <- bakeForSlot bid n
     maybe
         (fail $ "Could not bake for slot " ++ show n)
-        (\BS.BlockPointer{_bpBlock = NormalBlock block} -> return block)
+        ( \bs -> case bs of
+            BS.BlockPointer{_bpBlock = NormalBlock block} -> return block
+            -- This does not happen
```
abizjak, line 173: why.

concordium-consensus/tests/globalstate/GlobalStateTests/Trie.hs
```diff
@@ -37,13 +37,15 @@ testBranchAtFromList = forAll genBranchList $ \v ->
 testUpdateBranchSome :: Property
 testUpdateBranchSome = forAll genBranchList $ \v i x ->
     let v' = Trie.branchesToList $ Trie.updateBranch i (Some x) $ Trie.branchesFromList v
-        (h, _ : r) = splitAt (fromIntegral i) v
+        res = splitAt (fromIntegral i) v
+        (h, r) = (fst res, tail $ snd res)
```
abizjak, line 41:
```suggestion
        (h, rr) = splitAt (fromIntegral i) v
        r = tail rr
```
is more readable.

Also below.

concordium-consensus/tests/globalstate/GlobalStateTests/Updates.hs
```diff
@@ -272,7 +275,10 @@ testing2'1 = do
         getBlockStates
             >>= createAccountWith limitDelta
             >>= addBakerWith limit
-            >>= \(BCSuccess _ _, a) -> modifyStakeTo (limit - 1) a
+            >>= \bs -> case bs of
+                (BCSuccess _ _, a) -> modifyStakeTo (limit - 1) a
+                -- this does not happen
```
abizjak, line 280: Please explain why.

concordium-consensus/tests/scheduler/SchedulerTests/SmartContracts/V0/SmartContractTests.hs
```diff
@@ -317,7 +317,10 @@ simpleTransferTestCases =
     contractAddr = Types.AddressContract $ ContractAddress 0 0
 
     -- When using pointing to 32 0s in memory, this is the address that is used.
-    Right accRef = addressFromText "2wkBET2rRgE8pahuaczxKbmv7ciehqsne57F9gtzf1PVdr2VP3"
+    accRef = case addressFromText "2wkBET2rRgE8pahuaczxKbmv7ciehqsne57F9gtzf1PVdr2VP3" of
+        Right ref -> ref
+        -- This should not happen, since addressFromText was passed a valid address string.
+        Left _ -> error "addressFromText did not return an account reference"
```
abizjak, line 323:
```suggestion
    accRef = fromRight (addressFromText "2wkBET2rRgE8pahuaczxKbmv7ciehqsne57F9gtzf1PVdr2VP3")
```

concordium-node/Cargo.toml
abizjak: Yes.

And that is why we **also** need to update the flatbuffer compiler. The actual compiler used to generate code from the flatbuffers description.

This is installed on the system, and will be the thing that needs to be updated in the build jobs as well.

concordium-node/README.md
```diff
@@ -7,7 +7,7 @@
   - For macOS one should use the binutils provided by Xcode.
 - cmake >= 3.8.0
 - [flatc](http://google.github.io/flatbuffers/flatbuffers_guide_building.html)
-  v2.0.0 is what we currently use. Either build from the v2.0.0 tag of the repository using CMake and copy the `flatc` binary somewhere in your PATH, or download a released binary from <https://github.com/google/flatbuffers/releases/tag/v2.0.0> and place it somewhere in your PATH.
+  v2.1.2 is what we currently use. Either build from the v2.1.2 tag of the repository using CMake and copy the `flatc` binary somewhere in your PATH, or download a released binary from <https://github.com/google/flatbuffers/releases/tag/v2.1.2> and place it somewhere in your PATH.
```
abizjak, line 10: These versions are incorrect.

scripts/static-binaries/build-on-ubuntu.sh
```diff
@@ -14,7 +14,7 @@ set -euxo pipefail
 extra_features=${EXTRA_FEATURES:-""}
 
 protoc_version=3.15.3
-flatbuffers_version=v2.0.0
+flatbuffers_version=v2.1.2
```
abizjak, line 17: This is wrong.

service/windows/installer/resources/licenses.html
```diff
@@ -25784,7 +25784,7 @@ <h3>Apache-2.0</h3>
 See the License for the specific language governing permissions and
 limitations under the License.
 </pre>
-<h2><a href="https://hackage.haskell.org/package/ghc-9.0.2">ghc</a>
+<h2><a href="https://hackage.haskell.org/package/ghc-9.2.5">ghc</a>
```
abizjak, line 25787: Do not touch the licenses please. They are generated and the text of the license likely changed as well.

service/windows/installer/resources/licenses.html
```diff
@@ -25854,7 +25854,7 @@ <h3>BSD-3</h3>
 OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 DAMAGE.
 </pre>
-<h2><a href="https://github.com/ghc/ghc/tree/ghc-9.0.2-release/libraries/ghc-boot">ghc-boot</a>
+<h2><a href="https://github.com/ghc/ghc/tree/ghc-9.2.5-release/libraries/ghc-boot">ghc-boot</a>
```
abizjak, line 25857: Revert

service/windows/installer/resources/licenses.html
```diff
@@ -26089,7 +26089,7 @@ <h3>BSD3</h3>
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 </pre>
-<h2><a href="https://github.com/ghc/ghc/tree/ghc-9.0.2-release/libraries/ghci">ghci</a>
+<h2><a href="https://github.com/ghc/ghc/tree/ghc-9.2.5-release/libraries/ghci">ghci</a>
```
abizjak, line 26092: Revert this, and the previous license file changes.

abizjak (changes requested):

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -1340,72 +1343,75 @@ migratePersistentAccount StateMigrationParametersTrivial acc = do
...
+            return $!
+                PersistentAccount
+                    { accountNonce = _accountNonce,
+                      accountAmount = _accountAmount,
+                      ..
+                    }
+        _ ->
+            error $
+                "migratePersistentAccountFromV0 invariant violation: "
+                    <> "migrationParams should be StateMigrationParametersP4ToP5"
```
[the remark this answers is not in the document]
vikt0r0: It was changed because the compiler complained about redundant constraints and I was told to avoid pragmas. Is there a way to do it for a single declaration, rather than the whole file?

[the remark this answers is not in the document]
abizjak: The pragmas should be avoided, and some should be avoided at all cost, such as disabling completeness checks for pattern matching.

Others are sometimes necessary due to compiler bugs, but they are also not dangerous, such as redundant typeclass constraints. They should still be avoided, but if the choice is between disabling a critical feature or just a good one then that choice should be clear.

I think you have to disable the specific warning per file. I don't know of a way to do it per declaration.

concordium-consensus/tests/consensus/ConcordiumTests/Konsensus.hs
```diff
@@ -294,7 +294,11 @@ invariantSkovFinalization :: SkovState (Config t) -> FullBakerInfo -> Finalizati
 invariantSkovFinalization (SkovState sd@TS.SkovData{..} FinalizationState{..} _) baker maxFinComSize = do
     let finParams = finalizationParameters maxFinComSize
     invariantSkovData sd
-    let (flHead Seq.:|> (lfr, lfb)) = _finalizationList
+    let (flHead, lfr, lfb) =
+            case _finalizationList of
+                (flHead' Seq.:|> (lfr', lfb')) -> (flHead', lfr', lfb')
+                -- This does not happen
```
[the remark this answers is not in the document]
vikt0r0: It seems that this should be stated as an invariant of SkovState when invoking the function. Is this true?

concordium-consensus/tests/scheduler/SchedulerTests/SmartContracts/V0/SmartContractTests.hs
abizjak: No, my memory failed me is what that is. :P 

You could still do `fromRight (error "This should not happen because I say so.")) w`

EmilLa1 (commented):
Generally looks OK. 

But please whenever you write "this does not happen" then also explain _why_ it does not happen. 

Comments should strive to be as informative as possible, otherwise there is no reason they should be put there in the first place.

concordium-consensus/tests/consensus/ConcordiumTests/PassiveFinalization.hs
```diff
@@ -231,7 +231,11 @@ bake bid n = do
     mb <- bakeForSlot bid n
     maybe
         (fail $ "Could not bake for slot " ++ show n)
-        (\BS.BlockPointer{_bpBlock = NormalBlock block} -> return block)
+        ( \bs -> case bs of
+            BS.BlockPointer{_bpBlock = NormalBlock block} -> return block
+            -- This should not happen
```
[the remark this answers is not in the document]
vikt0r0: I think I need some help with this. It is not clear to me even from recursively looking up the relevant invocations.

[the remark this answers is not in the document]
abizjak:
```
            _ → fail "baking always produces a non-genesis block"
```
BakeForSlot produces a normal block always. Genesis blocks are never constructed by the node.

concordium-consensus/tests/consensus/ConcordiumTests/Update.hs
```diff
@@ -164,7 +168,11 @@ bake bid n = do
     mb <- bakeForSlot bid n
     maybe
         (fail $ "Could not bake for slot " ++ show n)
-        (\BS.BlockPointer{_bpBlock = NormalBlock block} -> return block)
+        ( \case
+            BS.BlockPointer{_bpBlock = NormalBlock block} -> return block
+            -- This does not happen
+            _ -> error "bs should be BlockPointer"
```
abizjak, line 174:
```suggestion
            _ -> error "baking always produces a normal (non-genesis) block."
```

concordium-consensus/tests/consensus/ConcordiumTests/Update.hs
abizjak: See comment/suggestion below

concordium-node/src/network/serialization/fbs.rs
```diff
@@ -105,7 +133,7 @@ fn _deserialize(buffer: &[u8]) -> anyhow::Result<NetworkMessage> {
 
 fn deserialize_packet(root: &network::NetworkMessage) -> anyhow::Result<NetworkPayload> {
     let packet = if let Some(payload) = root.payload() {
-        network::NetworkPacket::init_from_table(payload)
+        network_packet_init_from_table(payload)
     } else {
         bail!("missing network message payload (expected a packet)")
     };
```
abizjak, line 139:
```
let packet = root.payload_as_network_packet().context("missing network message payload (expected a packet)")?
```

concordium-node/src/network/serialization/fbs.rs
```diff
@@ -139,28 +167,25 @@ fn deserialize_packet(root: &network::NetworkMessage) -> anyhow::Result<NetworkP
 
 fn deserialize_request(root: &network::NetworkMessage) -> anyhow::Result<NetworkPayload> {
     let request = if let Some(payload) = root.payload() {
```
abizjak, line 169: Similarly use `root.payload_as_network_request` here.

concordium-node/src/network/serialization/fbs.rs
```diff
@@ -139,28 +167,25 @@ fn deserialize_packet(root: &network::NetworkMessage) -> anyhow::Result<NetworkP
...
-                    network_ids.safe_slice().iter().copied().map(NetworkId::from).collect();
+                let network_ids = network_ids.iter().map(NetworkId::from).collect();
                 Ok(NetworkPayload::NetworkRequest(NetworkRequest::GetPeers(network_ids)))
             } else {
                 bail!("missing network ids in a GetPeers request")
             }
         }
         network::RequestVariant::Handshake => {
-            if let Some(handshake) = request.payload().map(network::Handshake::init_from_table) {
+            if let Some(handshake) = request.payload().map(handshake_init_from_table) {
```
abizjak, line 188: Use `payload_as_handshake()` instead.

concordium-node/src/network/serialization/fbs.rs
```diff
@@ -139,28 +167,25 @@ fn deserialize_packet(root: &network::NetworkMessage) -> anyhow::Result<NetworkP
...
 
     match request.variant() {
         network::RequestVariant::Ping => Ok(NetworkPayload::NetworkRequest(NetworkRequest::Ping)),
         network::RequestVariant::GetPeers => {
-            if let Some(network_ids) = request
-                .payload()
-                .map(network::NetworkIds::init_from_table)
-                .and_then(|payload| payload.ids())
+            if let Some(network_ids) =
+                request.payload().map(network_ids_init_from_table).and_then(|payload| payload.ids())
```
abizjak, line 179: Use `payload_as_network_ids` instead.

concordium-node/src/network/serialization/fbs.rs
```diff
@@ -240,7 +263,7 @@ fn deserialize_request(root: &network::NetworkMessage) -> anyhow::Result<Network
 
 fn deserialize_response(root: &network::NetworkMessage) -> anyhow::Result<NetworkPayload> {
     let response = if let Some(payload) = root.payload() {
-        network::NetworkResponse::init_from_table(payload)
+        network_response_init_from_table(payload)
     } else {
```
abizjak, line 267: Use `payload_as_network_response` instead.

scripts/static-binaries/build-on-ubuntu.sh
```diff
@@ -14,7 +14,7 @@ set -euxo pipefail
 extra_features=${EXTRA_FEATURES:-""}
 
 protoc_version=3.15.3
-flatbuffers_version=v2.0.0
+flatbuffers_version=v22.10.26
```
abizjak, line 17: This looks like a different version than the one you mention in the README. Which one is the intended one?

abizjak (commented):
The Haskell parts look good now modulo minor issues.

Flatbuffers should be changed to remove the "safe" wrappers, and instead use the generated `payload_as_*` functions.

concordium-consensus/tests/globalstate/GlobalStateTests/Updates.hs
```diff
@@ -296,7 +313,15 @@ testing2'3 = do
...
             >>= addBakerWith limit
-            >>= \(BCSuccess _ _, a) -> modifyStakeTo (limit + 100) a
+            >>= \case
+                (BCSuccess _ _, a) -> modifyStakeTo (limit + 100) a
+                -- this does not happen, since
+                -- \* the account is valid;
+                -- \* the account is not a baker;
+                -- \* the account is not a delegator;
+                -- \* the account has sufficient balance to cover the stake,
+                -- @(BCSuccess [], _)@ is returned, see `bsoConfigureBaker`.
```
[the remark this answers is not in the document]
vikt0r0: I know that's how the test work, but it seems that due to the way the account is constructed, we know that this case really does not happen? Hence the comment.

concordium-consensus/tests/scheduler/SchedulerTests/Payday.hs
```diff
@@ -125,12 +125,21 @@ instance Arbitrary (UpdateValue 'ChainParametersV1) where
...
             snd
                 . maximumBy (\(a, _) (b, _) -> compare a b)
                 . takeWhile ((<= ps) . fst)
-                . filter (\(_, UVMintDistribution _) -> True)
+                . filter
+                    ( \pair ->
+                        case pair of
+                            (_, UVMintDistribution _) -> True
+                            _ -> error "pair should match (a, UVMintDistribution)"
+                    )
```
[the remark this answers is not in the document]
vikt0r0: You're right.

concordium-consensus/tests/globalstate/GlobalStateTests/Updates.hs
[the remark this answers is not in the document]
vikt0r0: While the change I did preserves semantics, I believe for this test to behave as intended, one should probably just return whatever was produced when it is not `BCSuccess`, and then evaluate the function to `return ()` to have the test fail when this is the case?

abizjak (approved):

PR #678, opened 2023-01-10
Fix storing node settings.

### LINKED ISSUE #677: The main.config.json file is not properly stored by the node
If the file `main.config.json` exists then it is parsed during startup (unless the node id is specified). It is then written to with any new settings. However the writing does not properly overwrite the file (it is not truncated) which means it can end up being corrupt.

An example of a corruption would be if a node version was `4.0.11` in the file and was changed to `5.0.0` then the file would end up with an extra `}` at the end.

abizjak:
## Purpose

Fixes #677 

## Changes

Overwrite the file when storing the new configuration.

## Commentary

The way the node id is stored is too complex and this part of the codebase could use some cleanup. However it is not worth it now. When we do node ID revision we should substantially revise this so that we don't have so much complexity for loading this trivial configuration.

In that epic we will also have to provide some way of securely loading the secret keys when the node starts.

## Checklist

      hard-to-understand areas.


concordium-consensus/package.yaml
```diff
@@ -267,6 +267,7 @@ tests:
     - hspec-expectations >= 0.8
     - vector >= 0.12
     - math-functions >= 0.2
+    - temporary >= 1.3
```
EmilLa1, line 270: Where is this used?

EmilLa1 (approved):
LGTM

Bargsteen (approved):
LGTM too ;)

concordium-consensus/package.yaml
abizjak: In a test. I forgot to commit this previously (when migrating a consensus test) so the CI is failing now on main. So I just fixed it in this PR.

Issue #667, opened 2023-01-05
Remove basic block state usage in Consensus tests


limemloh:
A few of the unit tests related to consensus (the current consensus) uses basic block state and should be rewritten to use the persistent block state implementation.

The relevant unit test files in `concordium-consensus/tests/consensus/` are: 
- [x] `ReceiveTransactionsTest.hs`
- [ ] `Konsensus.hs`
- [x] `PassiveFinalization.hs`
- [x] `FinalizationRecover.hs`
- [ ] `CatchUp.hs`

limemloh:
The remaining work on removing basic block state from `Konsensus.hs` more time than expected and is therefore put on hold for now, in order to prioritize the actual simplification.
We should get back to this afterwards.

limemloh:
The remaining work on removing basic block state from `Konsensus.hs` more time than expected and is therefore put on hold for now, in order to prioritize the actual simplification.
We should get back to this afterwards.

limemloh:
Since we are working on a new consensus and not including any new features to the current consensus at the moment, this task will be postponed indefinitely. It should be implemented if a bug is discovered in the current consensus.

abizjak:
I believe we agreed to remove the tests that are currently commented out. Can we do that instead of leaving them commented out?

abizjak:
Closing this since the remaining tests have been removed.

PR #680, opened 2023-01-12
Migrate 'ReceiveTransactionsTest' to use persistent state.

abizjak:
## Purpose

Part of https://github.com/Concordium/concordium-node/issues/667

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
@@ -72,6 +72,7 @@ import qualified Concordium.ID.Parameters as ID
 import qualified Concordium.ID.Types as ID
 import Concordium.Kontrol.Bakers
 import Concordium.Logger (MonadLogger)
+import Concordium.TimeMonad (TimeMonad)
 import Concordium.Types
 import Concordium.Types.Accounts (AccountBaker (..))
 import qualified Concordium.Types.Accounts as BaseAccounts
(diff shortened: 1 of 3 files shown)


limemloh (approved):

EmilLa1 (approved):
Looks good!

A few comments.. Which is probably something I've written at some point... 

1)The test only runs on PV1, maybe they should run for all PVs considering the tests are not protocol dependent. (I think such mechanism has been introduced ?) If not,  I think at least the tests should be run on P5. (line 206, couldn't comment on the exact line via github.)

2) I'm not really keen on the naming of the `MyState`, `MyBlockState` etc. (I'm aware that I am probably the culprit for this) But I don't have a better suggestion either - but feel free to change the names to better reflect what they do if you have any suggestions. 

Maybe `MyBlockState` => `PVHashedBlockState`,  `MyState` => `PVSkovState`,  `MyBlockStateMonad` => `PVBlockStateMonad`

abizjak:
> Looks good!
> 
> A few comments.. Which is probably something I've written at some point...
> 
> 1)The test only runs on PV1, maybe they should run for all PVs considering the tests are not protocol dependent. (I think such mechanism has been introduced ?) If not, I think at least the tests should be run on P5. (line 206, couldn't comment on the exact line via github.)
> 
>     2. I'm not really keen on the naming of the `MyState`, `MyBlockState` etc. (I'm aware that I am probably the culprit for this) But I don't have a better suggestion either - but feel free to change the names to better reflect what they do if you have any suggestions.
> 
> 
> Maybe `MyBlockState` => `PVHashedBlockState`, `MyState` => `PVSkovState`, `MyBlockStateMonad` => `PVBlockStateMonad`

I have renamed.

PR #682, opened 2023-01-13
Move grpc2 types to base

abizjak:
## Purpose

Move GRPC2 type definitions to concordium-base so that we don't have to duplicate generation and conversion both in the node and in the client.

There are only necessary changes apart from moving the module. Any further desired changes should be done 
afterwards.

## Depends on

- [ ] https://github.com/Concordium/concordium-base/pull/296

## Checklist

      hard-to-understand areas.


concordium-consensus/Setup.hs
```diff
@@ -41,8 +39,6 @@ makeRust args flags = do
 
 main =
     defaultMainWithHooks $
-        generatingProtos
-            "../concordium-grpc-api/v2"
-            simpleUserHooks
-                { preConf = makeRust
-                }
```
vikt0r0, line 48: How come we do not need `generatingProtos`from base here, but we do in the client?

vikt0r0 (approved):

limemloh (approved):

Bargsteen (approved):

concordium-consensus/Setup.hs
abizjak: It's because in the client we are **also** generating V1 api until you've made the changes there.

After the client is moved to V2 API completely we will not need it anymore.

PR #692, opened 2023-01-24
Remove unused modules in GlobalState/Basic

limemloh:
## Purpose

Now that Basic BlockState is removed, even more can be removed from `GlobalState.Basic`.

## Changes

- Remove some unused code and modules from `GlobalState.Basic`.
- Move some modules only used by tests to a module in tests.


concordium-consensus/tests/globalstate/Basic/AccountTable.hs
```diff
@@ -5,7 +5,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE ViewPatterns #-}
 
-module Concordium.GlobalState.Basic.BlockState.AccountTable where
+module Basic.AccountTable where
```
EmilLa1, line 8: Why are these modules renamed?

EmilLa1 (approved):
Great job!

concordium-consensus/tests/globalstate/Basic/AccountTable.hs
limemloh: Since they are only used by some tests now, I moved them out of `src` and into `tests`, later we should remove them entirely, but that will be another PR.

PR #694, opened 2023-01-25
Fix test documentation to not mislead future readers.

abizjak:
## Purpose

Fix explanation of why the test should fail.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
GJ

Bargsteen:
Nice.

We should also delete or add an assertion to the line in `a_modify`:
```
    ;; delete the iter
    (call $state_iterator_delete (i64.const 0)) ;; the iter created in the first receive function.`
```
It tries to delete an iterator created in `a_modify_proxy`, which, as I understand it, isn't allowed.

PR #713, opened 2023-02-03
Fix logging when Prometheus push gateway is enabled

limemloh:
## Purpose

If enabling both prometheus push metrics and exporter in the node, only the enabling of the exporter is logged.
This is a fix.

### DIFF
--- a/concordium-node/src/stats_export_service.rs
+++ b/concordium-node/src/stats_export_service.rs
@@ -324,11 +324,11 @@ impl StatsExportService {
         listen_addr: SocketAddr,
         error_sender: tokio::sync::broadcast::Sender<()>,
     ) -> Result<(), ()> {
-        log::info!("Starting Prometheus server listening on {}", listen_addr);
+        log::info!("Starting Prometheus exporter listening on {}", listen_addr);
         let result = gotham::plain::init_server(listen_addr, self.router()).await;
         if let Err(()) = result {
             // Log an error and notify main thread that an error occured.
-            error!("A runtime error occurred in the Prometheus server.");
+            error!("A runtime error occurred in the Prometheus exporter.");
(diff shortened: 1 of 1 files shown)


concordium-node/src/stats_export_service.rs
```diff
@@ -391,6 +389,7 @@ pub fn start_push_gateway(
     id: P2PNodeId,
 ) {
     if let Some(prom_push_addy) = conf.prometheus_push_gateway.as_ref() {
+        info!("Enabling prometheus push gateway at {}", prom_push_addy);
```
abizjak, line 392: This would be more appropriately "starting".

concordium-node/src/stats_export_service.rs
```diff
@@ -376,9 +376,7 @@ pub fn instantiate_stats_export_engine(
 ) -> anyhow::Result<Arc<StatsExportService>> {
     if conf.prometheus.prometheus_listen_port.is_some() {
         info!("Enabling prometheus server");
-    } else if let Some(ref push_gateway) = conf.prometheus.prometheus_push_gateway {
-        info!("Enabling prometheus push gateway at {}", push_gateway);
-    };
```
abizjak, line 381: I don't think we should split it like this. Both should be logged here I think. It's inconsistent to only do it one way.

We have "enabling" here for both, and "starting" when they are actually starting. Or we have nothing here and just rely on "starting" messages, which is also  fine.

abizjak (approved):

concordium-node/src/stats_export_service.rs
limemloh: I'll go with the latter. 
Also I changed some of the logs from "Prometheus Server" to "Prometheus exporter"

Issue #710, opened 2023-02-02
Extend the metrics of the Prometheus exporter


limemloh:
This issue is tracking potential metrics to add to the built-in Prometheus exporter. The metrics should only be relevant for someone maintaining a node.

- [x] Block height of last finalized block.
- [x] Timestamp of last arrived block.
- [x] Timestamp of last finalized block. 
  #720
- [x] Received transactions with labels for
  - Total valid transactions received. Consensus accepts the message.
  - Total invalid consensus messages received. Consensus rejects the message.
  - Total duplicate messages received by consensus. (To monitor the effectiveness of the deduplication of the network layer)
  - Total dropped
- [x] Received blocks with labels for
  - Total valid
  - Total invalid
  - Total duplicate
  - Total dropped
- [x] Received finalization messages (labeled).
  - Total valid
  - Total invalid
  - Total duplicate
  - Total dropped
  #724
- [x] Sent finalization messages.
- [x] Current soft banned peers.
- [x] Total packets received, (ensure the one we have is counting before deserializing network messages).
- [x] Total peers during runtime.
- [x] The current start time.
- [x] The current node version.
  #725

The above metrics are being released, the unimplemented metrics are tracked here #755

PR #724, opened 2023-02-10
Add metric for total received messages

limemloh:
## Purpose

Related to https://github.com/Concordium/concordium-node/issues/710
Extend Prometheus metrics with metric for total received messages.

## Changes

- Introduce metric `consensus_received_messages_total` with a label for message type and a label for result.
- Removed `network_inbound_high_priority_message_drops_total`, `network_inbound_low_priority_message_drops_total`, `network_inbound_high_priority_messages_total` and `network_inbound_high_priority_messages_total` as they can now be derived from `consensus_received_messages_total`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/consensus_ffi/helpers.rs
+++ b/concordium-node/src/consensus_ffi/helpers.rs
@@ -107,6 +107,18 @@ impl PacketType {
             PacketType::CatchUpStatus => false,
         }
     }
+
+    /// Get the label. This is used when updating metrics of the prometheus
+    /// exporter.
+    pub fn as_label(&self) -> &str {
+        match self {
+            PacketType::Block => "block",
+            PacketType::Transaction => "transaction",
+            PacketType::FinalizationRecord => "finalization record",
+            PacketType::FinalizationMessage => "finalization message",
+            PacketType::CatchUpStatus => "catch-up status message",
+        }
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Error)]
@@ -247,6 +259,16 @@ impl ConsensusFfiResponse {
             | ContinueCatchUp => false,
             PendingBlock => packet_type != PacketType::Block,
             Success | PendingFinalization | Asynchronous => true,
+        }
+    }
+
+    /// Get the label. This is used when updating metrics of the prometheus
+    /// exporter.
+    pub fn as_label(&self) -> &str {
+        match self {
+            ConsensusFfiResponse::Success => "valid",
+            ConsensusFfiResponse::DuplicateEntry => "duplicate",
+            _ => "invalid",
         }
     }
 }
--- a/concordium-node/src/consensus_ffi/helpers.rs
+++ b/concordium-node/src/consensus_ffi/helpers.rs
@@ -110,7 +110,7 @@ impl PacketType {
 
     /// Get the label. This is used when updating metrics of the prometheus
     /// exporter.
-    pub fn as_label(&self) -> &str {
+    pub fn label(&self) -> &str {
         match self {
             PacketType::Block => "block",
             PacketType::Transaction => "transaction",
@@ -264,11 +264,13 @@ impl ConsensusFfiResponse {
 
     /// Get the label. This is used when updating metrics of the prometheus
     /// exporter.
-    pub fn as_label(&self) -> &str {
-        match self {
-            ConsensusFfiResponse::Success => "valid",
-            ConsensusFfiResponse::DuplicateEntry => "duplicate",
-            _ => "invalid",
+    pub fn label(&self) -> &str {
+        if self.is_successful() {
+            "valid"
+        } else if let ConsensusFfiResponse::DuplicateEntry = self {
+            "duplicate"
+        } else {
+            "invalid"
         }
     }
 }
--- a/concordium-node/src/stats_export_service.rs
+++ b/concordium-node/src/stats_export_service.rs
@@ -82,7 +82,6 @@ pub struct StatsExportService {
     /// Possible values of `message` are:
     /// - `"block"`
     /// - `"transaction"`
-    /// - `"finalization record"`
     /// - `"finalization message"`
     /// - `"catch-up status message"`
     ///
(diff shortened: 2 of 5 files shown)


concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -107,6 +107,18 @@ impl PacketType {
             PacketType::CatchUpStatus => false,
         }
     }
+
+    /// Get the label. This is used when updating metrics of the prometheus
+    /// exporter.
+    pub fn as_label(&self) -> &str {
```
abizjak, line 113: What does `as` add to the name of this function?

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -249,6 +261,16 @@ impl ConsensusFfiResponse {
             Success | PendingFinalization | Asynchronous => true,
         }
     }
+
+    /// Get the label. This is used when updating metrics of the prometheus
+    /// exporter.
+    pub fn as_label(&self) -> &str {
```
abizjak, line 267: The `as` is more misleading here, since the operation is not reversible.

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -249,6 +261,16 @@ impl ConsensusFfiResponse {
...
+        match self {
+            ConsensusFfiResponse::Success => "valid",
+            ConsensusFfiResponse::DuplicateEntry => "duplicate",
```
abizjak, line 270: We should use the `is_success`  for `valid`. Not just the `Success` variant.

Duplicate is what we want here I think though since it measures purely duplicate messages.

concordium-node/src/stats_export_service.rs
```diff
@@ -86,6 +76,23 @@ pub struct StatsExportService {
     pub last_arrived_block_timestamp: IntGauge,
     /// The block height of the last finalized block.
     pub last_arrived_block_height: GenericGauge<AtomicU64>,
+    /// Total number of consensus messages received. Labelled with message type
+    /// (`message=<type>`) and the outcome (`result=<outcome>`).
+    ///
+    /// Possible values of `message` are:
+    /// - `"block"`
+    /// - `"transaction"`
+    /// - `"finalization record"`
```
abizjak, line 85: These are never sent on the network.

concordium-node/src/stats_export_service.rs
```diff
@@ -86,6 +76,23 @@ pub struct StatsExportService {
...
+    /// - `"finalization message"`
+    /// - `"catch-up status message"`
+    ///
+    /// Possible values of `result` are:
+    /// - `"valid"` Successful outcome.
+    /// - `"invalid"` Messages being rejected as invalid.
+    /// - `"dropped"` Messages being dropped due to a full queue.
+    /// - `"duplicate"` duplicate consensus messages (this is not accounting
+    ///   deduplication of the network layer).
```
abizjak, line 94:
```suggestion
    /// - `"duplicate"` duplicate consensus messages (these are duplicate messages determined so by consensus, **after** the message has already been deduplicated at the network layer).
```

docs/prometheus-exporter.md
```diff
@@ -116,3 +106,20 @@ The block height of the last arrived block.
...
+- `"transaction"`
+- `"finalization record"`
+- `"finalization message"`
+- `"catch-up status message"`
+
+Possible values of `result` are:
+- `"valid"` Successful outcome.
+- `"invalid"` Messages being rejected as invalid.
+- `"dropped"` Messages being dropped due to a full queue. See either `network_inbound_low_priority_message_queue_size` or `network_inbound_high_priority_message_queue_size` for more on this.
+- `"duplicate"` Duplicate consensus messages (this is not including deduplication of the network layer).
```
abizjak, line 125: Please clarify what is meant by "this is not including ..." part. It is not going to be clear to readers. Explain what the value is measuring instead.

limemloh: I have used something similar to what you suggested in the previous comments

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -249,6 +261,18 @@ impl ConsensusFfiResponse {
             Success | PendingFinalization | Asynchronous => true,
         }
     }
+
+    /// Get the label. This is used when updating metrics of the prometheus
+    /// exporter.
+    pub fn label(&self) -> &str {
```
EmilLa1, line 267: Do we want more fine grained labels? I.e. 
`is_acceptable`, `is_pending`, `is_rebroadcastable` .

EmilLa1 (approved):

docs/prometheus-exporter.md
```diff
@@ -116,3 +106,19 @@ The block height of the last arrived block.
 Timestamp for the processed last arrived block (as Unix time in milliseconds).
 
 Note that this is the time when the node has last processed a new block. It is **not** the objective timestamp of the block (i.e., slot time).
+
+### `consensus_received_messages_total`
```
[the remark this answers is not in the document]
limemloh: I don't think adding total is providing more information, and is just here because the Prometheus documentation recommends having it in this case.

Also changing this now, would make it inconsistent with the other metrics like this, so I'll leave it as is

concordium-node/src/consensus_ffi/helpers.rs
limemloh: If you think it would be useful information when maintaining a node, we should add it, but if not, then it would just add more for the maintainer to understand.

abizjak: Let's go with this for now. If we find the need we can revise in the future.

docs/prometheus-exporter.md
```diff
@@ -55,46 +55,36 @@ Current number of connected peers. This is incremented when a peer completes a h
...
-
-### `network_inbound_low_priority_messages_total`
-
-Total inbound low priority messages received. This is incremented when a consensus message is enqueue in the low priority queue. Note: Does not include the dropped messages, see `network_inbound_low_priority_message_drops_total`.
-
 ### `network_inbound_high_priority_message_queue_size`
 
 Current number of consensus messages in the inbound high priority queue. Start dropping messages when larger than 16 * 1024.
 
+High priority messages are blocks, finalizations records, finalization messages and catch-up status messages.
```
abizjak, line 62: Records are still mentioned here.

abizjak (approved):

PR #725, opened 2023-02-10
Extend with even more metrics

limemloh:
## Purpose

Related to #710.

## Checklist

      hard-to-understand areas.


concordium-node/src/stats_export_service.rs
```diff
@@ -93,6 +93,25 @@ pub struct StatsExportService {
...
+    /// - `"finalization record"`
+    /// - `"finalization message"`
+    /// - `"catch-up status message"`
+    pub sent_messages: IntCounterVec,
+    /// Current number of soft banned peers.
+    pub soft_banned_peers: IntGauge,
+    /// Total number of peers connected since startup.
+    pub peers: IntCounter,
+    /// General node information, so far only contains a label `version` with
+    /// the version of the node.
```
abizjak, line 111: I'm not very keen on this "general" description. If we don't have any specifics that would fit this, then please document what it is, and not "it could be more" with no specifics.

abizjak: This is also missing the description of how this is used.

There is only one value for this, 1, and the label is the actual data.

I'm not really a fan of this use of Prometheus. It is clearly hacky to have a boolean flag for each possible version of the node.

docs/prometheus-exporter.md
```diff
@@ -123,3 +123,22 @@ Possible values of `result` are:
 - `"invalid"` Messages being rejected as invalid.
 - `"dropped"` Messages being dropped due to a full queue. See either `network_inbound_low_priority_message_queue_size` or `network_inbound_high_priority_message_queue_size` for more on this.
 - `"duplicate"` Duplicate consensus messages (this is not including deduplication of the network layer).
+
+### `network_soft_banned_peers`
+
+Current number of soft banned peers. The node temporarily ban peers if they fail to follow the protocol.
```
abizjak, line 129:
```suggestion
Current number of soft banned peers. The node temporarily bans peers if they fail to follow the protocol.
```

docs/prometheus-exporter.md
```diff
@@ -123,3 +123,22 @@ Possible values of `result` are:
...
+
+### `network_peers_total`
+
+Total number of peers connected since startup. This is incremented when a peer completes a handshake.
```
abizjak, line 133: "completes a handshake" is the definition of becoming a peer as we use it in the network layer.

Before that it is just a "connection". So the second part of the sentence does not add value. 

We could additionally have a metric (maybe we do already) on "connections received or established". That would be useful information.

limemloh: So I have removed the last sentence.

We have the metric `network_connections_received_total` which is counting every "attempt" at connecting, this counts even if the connection is banned, duplicate or the node is at its limit.
If you mean something different from this, then we should add it to the metric wishlist.

concordium-node/src/bin/cli.rs
```diff
@@ -58,6 +58,8 @@ async fn main() -> anyhow::Result<()> {
     let (node, server, poll) =
         instantiate_node(&conf, &mut app_prefs, stats_export_service, regenesis_arc.clone())
             .context("Failed to create the node.")?;
+    // Set the startime in the stats.
+    node.stats.node_startup_timestamp.set(node.start_time.timestamp_millis());
```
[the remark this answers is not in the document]
limemloh: I have added it to the bootstrapper

concordium-node/src/p2p/connectivity.rs
```diff
@@ -608,6 +608,7 @@ pub fn connection_housekeeping(node: &Arc<P2PNode>) -> bool {
         if !soft_bans.is_empty() {
             let now = Instant::now();
             soft_bans.retain(|_, expiry| *expiry > now);
+            node.stats.soft_banned_peers.set(soft_bans.len() as i64);
```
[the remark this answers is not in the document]
limemloh: Good catch

concordium-node/src/stats_export_service.rs
limemloh: I'll make it more specific.

I understand why you dislike this approach, but it is from the Prometheus documentation https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels.
We could choose to not have this information, I guess.

concordium-node/src/connection/mod.rs
```diff
@@ -555,6 +555,7 @@ impl Connection {
         self.remote_peer.self_id = Some(id);
         self.remote_peer.external_port = peer_port;
         self.handler.stats.connected_peers.inc();
+        self.handler.stats.peers.inc();
```
EmilLa1, line 558:
```suggestion
        self.handler.stats.total_peers.inc();
```
?

concordium-node/src/stats_export_service.rs
```diff
@@ -223,6 +241,47 @@ impl StatsExportService {
...
+        )?;
+        registry.register(Box::new(sent_consensus_messages.clone()))?;
+
+        let soft_banned_peers = IntGauge::with_opts(Opts::new(
+            "network_soft_banned_peers",
+            "Current number of soft banned peers",
+        ))?;
+        registry.register(Box::new(soft_banned_peers.clone()))?;
+
+        let peers = IntCounter::with_opts(Opts::new(
```
EmilLa1, line 260: total peers?
Just so we don't accidentally confuse it with the current peers.

EmilLa1 (approved):

docs/prometheus-exporter.md
```diff
@@ -122,3 +122,31 @@ Possible values of `result` are:
...
+Current number of soft banned peers. The node temporarily bans peers if they fail to follow the protocol.
+
+### `network_peers_total`
+
+Total number of peers connected since startup.
+
+### `node_info`
+
+Information of the node software. Provides the node version using a label (`version=<version>`).
+Always have the value 1.
```
abizjak, line 147:
```suggestion
Always has the value 1.
```

concordium-node/src/stats_export_service.rs
```diff
@@ -223,6 +241,47 @@ impl StatsExportService {
...
+            "network_peers_total",
+            "Total number of peers since startup",
+        ))?;
+        registry.register(Box::new(total_peers.clone()))?;
+
+        let node_info = IntGauge::with_opts(
+            Opts::new(
+                "node_info",
+                "Node software information. Provides the node version using a label \
+                 (`version=<version>`). Always have the value 1",
```
abizjak, line 270:
```suggestion
                 (`version=<version>`). Always has the value 1",
```

abizjak (approved):

PR #727, opened 2023-02-13
Clean up a few structures related to the new consensus.

EmilLa1:
## Purpose

Tidy up a few types for the new consensus protocol based on reviews.

## Changes

- [x] Remove the `BakerSignVerifyKey` from the `BakedBlock` structure.
- [x] the block signature signs H(genesis)
- [x] the timeout signature signs H(genesis)
- [x] the aggregate signature in a timeout signs qc.epoch as well as qc.round, currentround and H(genesis)
- [x] the data structure within the TC that keeps track of the signed rounds now includes the epoch as well

## Checklist

      hard-to-understand areas.


concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Types.hs
```diff
@@ -34,28 +34,38 @@ genBlsSignature = flip Bls.sign someBlsSecretKey . BS.pack <$> vector 10
...
 genQuorumSignature = QuorumSignature <$> genBlsSignature
 
+-- |Generate a block hash for a genesis
+genGenesisBlockHash :: Gen BlockHash
+genGenesisBlockHash = BlockHash . Hash.Hash . FBS.pack <$> vector 32
+
 -- |Generate a quorum certificate in a way that is suitable for testing serialization.
-genQuorumCertificate :: Gen QuorumCertificate
-genQuorumCertificate = do
+-- The genesis block hash must be provided.
```
abizjak, line 42: Can you elaborate on the Maybe then? It is unclear what "must" means when you can omit it.

abizjak (approved):

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -414,12 +431,12 @@ checkTimeoutCertificateSignature tsmGenesis toKeys TimeoutCertificate{..} =
     Bls.verifyAggregateHybrid msgsKeys (theTimeoutSignature tcAggregateSignature)
   where
     msgsKeys =
-        [ (timeoutSignatureMessageBytes TimeoutSignatureMessage{tsmRound = tcRound, ..}, toKeys fs)
-          | (tsmQCRound, fs) <- finalizerRoundsList tcFinalizerQCRounds
+        [ (timeoutSignatureMessageBytes TimeoutSignatureMessage{tsmRound = tcRound, tsmEpoch = tcEpoch, ..}, toKeys fs)
+          | ((tsmQCRound, tsmQCEpoch), fs) <- finalizerRoundsList tcFinalizerQCRounds
```
chportma, line 435: The `toKeys` function needs to change. (By the way, is it implemented somewhere ? I couldn't find the source by searching through the files for that word. But maybe it is in submodule that I haven't loaded.) It should take the epoch as input as well, so that it knows which set of keys to retrieve. That was the reason for including the epoch of the best QC in the `TimeoutSignatureMessage`. I believe it shouldn't be possible for there to be more than two different epochs in the `tcFinalizerQCRounds`, and under certain conditions there should only be one epoch and we can forget the old keys. But we still need to work those details out.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -431,6 +448,8 @@ data TimeoutMessageBody = TimeoutMessageBody
       tmFinalizerIndex :: !FinalizerIndex,
       -- |Round number of the round being timed-out.
       tmRound :: !Round,
+      -- |Epoch of which the round being timed-out.
+      tmEpoch :: !Epoch,
```
chportma, line 452: Same here. Remove `tmEpoch`.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -590,12 +615,12 @@ data BakedBlock = BakedBlock
       bbRound :: !Round,
       -- |Block epoch number.
       bbEpoch :: !Epoch,
+      -- |The genesis 'BlockHash'.
+      bbGenesis :: !BlockHash,
```
chportma, line 619: If I am understanding the Haskell code correctly, now you even have the genesis hash twice as part of a block. Once as a field on its own, once as a field in the QC. `blockGenesis` will grab the one in the QC. When building a `BlockHeader`, `bbBlockHeader` will grab the one from `bbGenesis`. I would remove them both.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -287,12 +291,14 @@ instance Serialize FinalizationEntry where
         put feFinalizedQuorumCertificate
         let QuorumCertificate{..} = feSuccessorQuorumCertificate
         put qcEpoch
```
chportma, line 293: As pointed out in a comment in slack, the line `put qcEpoch` is now redundant, since both QCs should have the same epoch for it to be a valid `FinalizationEntry` (if that is the final decision on finalization entries).

chportma (changes requested):
If some discussion is needed about what changes should be made we could discuss it in the meeting today. I won't have much time for a conversation on slack before the meeting anyway.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: It's just a function being passed into the `checkTimeoutCertificateSignature`. With this signature `(FinalizerSet  -> [BakerAggregationVerifyKey])`. I have spiced it up a bit so now it is declared as 
`(FinalizerSet -> Epoch -> [BakerAggregationVerifyKey]).

td202: I think exactly how this function retrieves the keys may be handled differently in future when we have a better understanding of how we retrieve the finalization committee(s) involved.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -741,28 +737,47 @@ getSignedBlock spv tt = do
...
+-- The same genesis hash must be provided when signing the block in order to verify the block signature.
+-- This function outputs the bytestring consisting of H(genesisBlockHash || blockHash).
+blockSignatureMessageBytes ::
+    -- |Hash of the genesis block
+    BlockHash ->
+    -- |Hash of the block
+    BlockHash ->
+    -- |The H(genesisBlockHash || blockHash) bytestring.
+    BS.ByteString
+blockSignatureMessageBytes genesisHash bHash = Hash.hashToByteString $! Hash.hashOfHashes (blockHash genesisHash) (blockHash bHash)
```
td202, line 750: Instead of `H(genesisBlockHash || blockHash)`, we could just sign `genesisBlockHash || blockHash`. We should revisit when we're reviewing our signature message discipline.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Types.hs
```diff
@@ -34,6 +34,10 @@ genBlsSignature = flip Bls.sign someBlsSecretKey . BS.pack <$> vector 10
 genQuorumSignature :: Gen QuorumSignature
 genQuorumSignature = QuorumSignature <$> genBlsSignature
 
+-- |Generate a block hash for a genesis
+genGenesisBlockHash :: Gen BlockHash
+genGenesisBlockHash = BlockHash . Hash.Hash . FBS.pack <$> vector 32
```
td202, line 39: Could rename to `genBlockHash`. Documentation should note this is an arbitrary hash value from the range of hashes.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Types.hs
```diff
@@ -69,13 +74,21 @@ genFinalizerRounds =
...
         fs <- genFinalizerSet
-        return (r, fs)
+        return ((r, e), fs)
+
+-- |Generate round and epoch
+genRoundAndEpoch :: Gen (Round, Epoch)
+genRoundAndEpoch = do
+    r <- Round <$> arbitrary
+    e <- arbitrary
+    return (r, e)
```
td202, line 86: If there's no correlation between the generated round and epoch, then I think there's no real point generating them as a pair.

td202 (approved):
In the list of changes, it says "the timeout signature signs H(genesis)". However, this doesn't seem to be a change: the genesis hash was already part of `TimeoutSignatureMessage`.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Haven't done this now. But I'll create an issue for "reviewing signature message discipline" .

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -208,6 +208,8 @@ instance Show FinalizerSet where
 data QuorumCertificate = QuorumCertificate
     { -- |Hash of the block this certificate refers to.
       qcBlock :: !BlockHash,
+      -- |Hash of the genesis block this certificate refers to.
+      qcGenesis :: !BlockHash,
       -- |Round of the block this certificate refers to.
```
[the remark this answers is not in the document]
EmilLa1: Addressed this, please have a look and confirm that things are as expected.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -321,12 +328,16 @@ instance HashableTo Hash.Hash (Option FinalizationEntry) where
...
 data TimeoutSignatureMessage = TimeoutSignatureMessage
-    { -- |Hash of the genesisBlock
+    { -- |Hash of the genesis block.
       tsmGenesis :: !BlockHash,
       -- |Round number of the timed-out round
       tsmRound :: !Round,
-      -- |Round number of the highest known valid quorum certificate
-      tsmQCRound :: !Round
+      -- |The epoch of which the round has timed-out.
+      tsmEpoch :: !Epoch,
```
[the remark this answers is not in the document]
EmilLa1: Addressed this, please have a look and confirm that things are as expected.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -368,16 +381,18 @@ instance Serialize FinalizerRounds where
...
+finalizerRoundsList :: FinalizerRounds -> [((Round, Epoch), FinalizerSet)]
 finalizerRoundsList = Map.toAscList . theFinalizerRounds
 
 -- |A timeout certificate aggregates signatures on timeout messages for the same round.
 -- Finalizers may have different QC rounds.
 data TimeoutCertificate = TimeoutCertificate
     { -- |The round that has timed-out.
       tcRound :: !Round,
+      -- |The epoch of which the round has timed-out.
+      tcEpoch :: !Epoch,
```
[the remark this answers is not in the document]
EmilLa1: Addressed this, please have a look and confirm that things are as expected.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Won't spend more time on this now. I put the epoch in the function there for now.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Addressed this, please have a look and confirm that things are as expected.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Addressed this, please have a look and confirm that things are as expected.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Addressed this, please have a look and confirm that things are as expected.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
chportma: In 3 signatures out of 4 you append the genesisBlockHash to the rest of the message, and in the 4th you hash them together. There is nothing wrong with that from a security point of view, but it certainly messes with my head :-D

chportma (approved):

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Yea, we should probably align this at some point. 
I'll let it be for now, but I have created an issue for it here: https://github.com/Concordium/concordium-node/issues/741

PR #742, opened 2023-02-15
Extend metrics grpc

limemloh:
## Purpose

Related to #755 

## Changes

- Add tower layer to the gRPC api updating metrics.
- Extend prometheus metrics with metric tracking gRPC request duration time.

## Checklist

      hard-to-understand areas.


concordium-node/Cargo.lock
```diff
@@ -566,7 +566,6 @@ version = "0.1.0"
 dependencies = [
  "aggregate_sig",
  "anyhow",
- "bs58",
```
abizjak, line 569: Why this?

concordium-node/Cargo.lock
```diff
@@ -581,7 +580,6 @@ dependencies = [
  "hex",
  "id",
  "itertools",
- "leb128",
```
abizjak, line 584: And this?

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,116 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: hyper::Request<hyper::Body>) -> Self::Future {
+        // This is necessary because tonic internally uses `tower::buffer::Buffer`.
+        // See https://github.com/tower-rs/tower/issues/547#issuecomment-767629149
+        // for details on why this is necessary
+        let clone = self.inner.clone();
```
abizjak, line 2375: Can you please link to the documentation https://docs.rs/tower/latest/tower/trait.Service.html#be-careful-when-cloning-inner-services which explains this.

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,116 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+        let mut inner = std::mem::replace(&mut self.inner, clone);
+
+        let grpc_request_duration = self.stats.grpc_request_duration.clone();
+
+        Box::pin(async move {
+            let endpoint_name = req.uri().path().to_owned();
+            let request_received = tokio::time::Instant::now();
+            let mut response = inner.call(req).await?;
```
abizjak, line 2383: Is there a reason we don't update metrics on failure of the inner call?

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,116 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+            let duration = request_received.elapsed().as_secs_f64();
+
+            let grpc_status_header = response.headers().get("grpc-status");
+
+            if let Some(header_value) = grpc_status_header {
```
abizjak, line 2388:
```suggestion
            if let Some(header_value) = response.headers().get("grpc-status") {
```

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,116 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+                let status_code_label =
+                    get_grpc_code_label(tonic::Code::from_bytes(header_value.as_bytes()));
+
+                grpc_request_duration
+                    .with_label_values(&[endpoint_name.as_str(), status_code_label])
+                    .observe(duration);
+            } else if let Ok(Some(headers)) =
```
abizjak, line 2395:
```suggestion
            } else if let Ok(Some(trailers)) =
```

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,116 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+                hyper::body::HttpBody::trailers(response.body_mut()).await
+            {
+                if let Some(header_value) = headers.get("grpc-status") {
```
abizjak, line 2398:
```suggestion
                if let Some(header_value) = trailers.get("grpc-status") {
```

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,116 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+
+        let grpc_request_duration = self.stats.grpc_request_duration.clone();
+
+        Box::pin(async move {
+            let endpoint_name = req.uri().path().to_owned();
+            let request_received = tokio::time::Instant::now();
+            let mut response = inner.call(req).await?;
+            let duration = request_received.elapsed().as_secs_f64();
+
+            let grpc_status_header = response.headers().get("grpc-status");
```
abizjak, line 2386: Can you explain here how this is meant to work. 

For each chunk we expect a status code? If not then we don't seem to update the metrics. But then we also check the body for any trailers.

Can you explain (in documentation) why we don't do anything, and why this makes sense, if the grpc-status header is not there.

docs/prometheus-exporter.md
```diff
@@ -150,3 +150,33 @@ Always has the value 1.
...
+
+Histogram tracking the total number of gRPC requests received and their duration. Labelled with the gRPC method name (`method=<name>`) and the gRPC response status (`status=<status>`).
+
+The duration is tracked for the underlaying HTTP2 request response time, meaning the streaming gRPC methods which could span over multiple HTTP2 requests are counted for each.
+As a result the duration represents the time spent by the node of each method and are not influence by the processing speed of the client for each chunk.
+
+The size of the buckets be configured using the `prometheus-grpc-duration-buckets` (`CONCORDIUM_NODE_PROMETHEUS_GRPC_DURATION_BUCKETS`) and is provided as a list of decimal numbers separated by ",". Each value represents the upper inclusive bound of a bucket (in seconds) and a bucket with +Infinity is always added. The values must be sorted in strictly increasing order.
+The default value of the configuration is `"0.050,0.100,0.200,0.500,1.000"`.
+
+For a complete list of possible method names refer to the gRPC api documentation.
```
abizjak, line 163:
```suggestion
For a complete list of possible method names refer to the gRPC API documentation.
```

docs/prometheus-exporter.md
```diff
@@ -150,3 +150,33 @@ Always has the value 1.
 
 Timestamp of starting up the node (Unix time in milliseconds).
 
+### `grpc_request_duration_seconds`
+
+Histogram tracking the total number of gRPC requests received and their duration. Labelled with the gRPC method name (`method=<name>`) and the gRPC response status (`status=<status>`).
+
+The duration is tracked for the underlaying HTTP2 request response time, meaning the streaming gRPC methods which could span over multiple HTTP2 requests are counted for each.
```
abizjak, line 157: It's hard to understand what you mean "for each" at the end. Can you elaborate please.

concordium-node/src/grpc2.rs
limemloh: right, so when I tried that the compiler complained about the async block not being Send anymore.
I'll give it another attempt

concordium-node/src/grpc2.rs
limemloh: I have updated the comment

concordium-node/Cargo.lock
limemloh: not sure why, remove this change

docs/prometheus-exporter.md
limemloh: I have updated the documentation since it did not reflect what actually happens

concordium-node/src/grpc2.rs
limemloh: I have added a comment, and it now also handles the case where no header was found at all, just for good measure

concordium-node/src/configuration.rs
```diff
@@ -93,27 +93,27 @@ pub struct PrometheusConfig {
         default_value = "127.0.0.1",
         env = "CONCORDIUM_NODE_PROMETHEUS_LISTEN_ADDRESS"
     )]
-    pub prometheus_listen_addr:   String,
+    pub prometheus_listen_addr: String,
```
abizjak, line 96: I think we should change this to `IpAddr` https://doc.rust-lang.org/std/net/enum.IpAddr.html

docs/prometheus-exporter.md
```diff
@@ -150,3 +150,33 @@ Always has the value 1.
 
 Timestamp of starting up the node (Unix time in milliseconds).
 
+### `grpc_request_response_time_seconds`
+
+Histogram tracking the total number of gRPC requests received and the time it took to provide a response. Labelled with the gRPC method name (`method=<name>`) and the gRPC response status (`status=<status>`).
+
+The duration tracked is time it takes for the underlaying HTTP request to construct a response, which does not include the time it takes to stream the response to the client.
```
abizjak, line 157: "The duration tracked is time it takes for the underlaying HTTP request to construct a response"

Does not really make sense. Do you mean that the time it takes for the handler to construct a response? I don't think we need to be mentioning HTTP here at all.

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,136 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+                    let status_code_label =
+                        get_grpc_code_label(tonic::Code::from_bytes(header_value.as_bytes()));
+
+                    grpc_request_duration
+                        .with_label_values(&[endpoint_name.as_str(), status_code_label])
+                        .observe(duration);
+                }
+            } else {
+                grpc_request_duration
+                    .with_label_values(&[endpoint_name.as_str(), "unknown"])
```
abizjak, line 2424: Maybe it'd be better to do `get_grpc_code_label(Code::Unknown)` instead of the string `unknown` in case we ever change this.

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,136 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+
+            let (mut response, duration) = {
+                // Forward the request.
+                let result = inner.call(req).await;
+                // Time taken for the inner service to send back a response, meaning for
+                // streaming gRPC methods this is the duration for it to first return a stream.
+                let duration = request_received.elapsed().as_secs_f64();
+                if result.is_err() {
+                    grpc_request_duration
+                        .with_label_values(&[endpoint_name.as_str(), "internal"])
```
abizjak, line 2393: Maybe it'd be better to do `get_grpc_code_label(Code::Internal)`

concordium-node/src/grpc2.rs
```diff
@@ -2290,3 +2296,136 @@ impl<A> Require<tonic::Status> for Option<A> {
...
+
+    fn layer(&self, service: S) -> Self::Service {
+        StatsMiddleware {
+            stats: self.stats.clone(),
+            inner: service,
+        }
+    }
+}
+
+/// Tower middleware for updating the node stats related to the gRPC API.
```
abizjak, line 2317:
```suggestion
/// Tower middleware for updating the node statistics related to the gRPC API.
```

abizjak (approved):

concordium-node/src/grpc2.rs
abizjak: I see. The wonders of async program transformation.

docs/prometheus-exporter.md
limemloh: I have rephrased this now

PR #748, opened 2023-02-21
New consensus leader election

td202:
## Purpose

Implement leader election for the new consensus. This implements the functionality for `getLeader` and `getLENonce` from the bluepaper. Note, however, that the leader election nonce is computed incrementally as part of the block state, and so the functionality of `getLENonce` is split over `updateSeedStateForBlock` and `updateSeedStateForEpoch`.

## Changes

- New functions:
  - `getLeader`: computes the leader for a particular round given the bakers and leader election nonce.
  - `updateSeedStateForBlock`: computes the incremental update to the leader election nonce for a single block.
  - `updateSeedStateForEpoch`: computes the leader election nonce for the next epoch, given that the seed state is updated for all blocks in the current epoch.
- Tests for the above functions.
- A benchmark for `getLeader`.
- Given changes to the seed state representation, some functions in Concordium.Kontrol.Bakers, Concordium.Kontrol.UpdateLeaderElectionParameters and Concordium.Scheduler.TreeStateEnvironment have small changes.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+
+import Concordium.Crypto.VRF (proofToHash)
+import Concordium.GlobalState.BakerInfo
+import Concordium.KonsensusV1.Types
+
+-- |Compute the leader for a given round, given the set of bakers and leadership election nonce
+-- for the epoch. The leader is computed as follows:
+--
+--  * Compute the SHA256 hash of the nonce and round number.
+--  * Treating this as an big-endian unsigned integer, compute the remainder when divided by the
```
chrmatt, line 27: Taking the remainder modulo total stake is different from the description in the bluepaper. Why is it done like this?

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+--    total stake among all the bakers, @targetVal@.
+--  * Find the least index @i@ such that the sum of stakes of bakers with indices less than or equal
+--    to @i@ is greater than @targetVal@. The baker with this index is the leader.
+--
+-- This process has a bias towards bakers with smaller indexes. However, the chance of this bias
+-- affecting the result of any given election is less than 1 in 10^57, and so is effectively
```
chrmatt, line 33: Where does the bias come from and why 10^57?

concordium-consensus/tests/scheduler/SchedulerTests/Helpers.hs
```diff
@@ -150,7 +150,7 @@ createTestBlockStateWithAccounts accounts =
     keys = Types.withIsAuthorizationsVersionForPV (Types.protocolVersion @pv) $ DummyData.dummyKeyCollection
     seedState = case Types.consensusVersionFor (Types.protocolVersion @pv) of
         Types.ConsensusV0 -> initialSeedStateV0 (Hash.hash "") 1_000
-        Types.ConsensusV1 -> initialSeedStateV1 (Hash.hash "")
+        Types.ConsensusV1 -> initialSeedStateV1 (Hash.hash "") 3_600_000
```
chrmatt, line 153: Do the 1000 and 3600000 here have any meaning?

concordium-consensus/benchmarks/leader-election/LeaderElectionBench.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | A simple benchmark for the leader election implementation.
```
EmilLa1, line 6: Please elaborate a bit on the finer details of the benchmark, so that one that comes to revisit this in the future understands the background, your thoughts and the result.

concordium-consensus/src/Concordium/GlobalState/BakerInfo.hs
```diff
@@ -50,6 +51,21 @@ data FullBakers = FullBakers
...
+-- The serialization format is:
+--
+-- * Number of bakers (64 bit, big endian)
+-- * For each baker in ascending order of baker ID:
+--     * Baker ID
+--     * Election (VRF) public key
+--     * Signature public key
+--     * Aggregate signature public key
+--     * Effective stake amount of the baker
+putFullBakers :: Putter FullBakers
```
EmilLa1, line 64: For the sake of it, please add a simple ser/deserialise test.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+-- impossible.
+--
+-- Note: this implementation is linear in the number of bakers. By pre-processing the baker set,
+-- we could have a logarithmic algorithm. However, as the runtime is on the order of 5 us for 1000
+-- bakers and 45 us for 10000, this is not a worthwhile optimisation at the current time.
```
EmilLa1, line 38: Maybe just add a link here to the checked in benchmark.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+    hsh = getHash $ runPutLazy $ put nonce >> put rnd
+    totalStake = toInteger . sum $ snd <$> bakers
+    targetVal = fromIntegral $ Hash.hashToInteger hsh `mod` fromIntegral totalStake
+    grabBaker runningTotal ((bi, amt) : bkrs)
+        | targetVal < runningTotal + amt = bi
+        | otherwise = grabBaker (runningTotal + amt) bkrs
+    grabBaker _ [] = error "getLeader: Empty bakers"
+
+-- |Compute the update for the leadership election nonce due to a particular block nonce.
+updateWithBlockNonce :: BlockNonce -> Hash.Hash -> Hash.Hash
```
EmilLa1, line 58: please document arguments.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+updateWithBlockNonce bn un = Hash.hash $ runPut $ put un <> put (proofToHash bn)
+
+-- |Update the running 'updatedNonce' in seed state.  Blocks after the trigger block do not
+-- contribute.  If the timestamp is at least the epoch transition time, then the
+-- 'epochTransitionTriggered' flag is set in the seed state, indicating that a new epoch should
+-- begin once the block is finalized.
+updateSeedStateForBlock :: Timestamp -> BlockNonce -> SeedState 'SeedStateVersion1 -> SeedState 'SeedStateVersion1
```
EmilLa1, line 65: please document arguments.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+    totalStake = toInteger . sum $ snd <$> bakers
+    targetVal = fromIntegral $ Hash.hashToInteger hsh `mod` fromIntegral totalStake
+    grabBaker runningTotal ((bi, amt) : bkrs)
+        | targetVal < runningTotal + amt = bi
+        | otherwise = grabBaker (runningTotal + amt) bkrs
+    grabBaker _ [] = error "getLeader: Empty bakers"
+
+-- |Compute the update for the leadership election nonce due to a particular block nonce.
+updateWithBlockNonce :: BlockNonce -> Hash.Hash -> Hash.Hash
+updateWithBlockNonce bn un = Hash.hash $ runPut $ put un <> put (proofToHash bn)
```
EmilLa1, line 59: what does `un` stand for?

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -319,68 +319,69 @@ getSlotBakersP4 ::
...
+                                        ePoolParams pp' updates
+                                ePoolParams pp _ = pp
+                                effectivePoolParameters = ePoolParams (chainParams ^. cpPoolParameters) pendingPoolParams
+                                bsc = computeBakerStakesAndCapital @m effectivePoolParameters activeBakers passiveDelegators
+                            let mkFullBaker (biRef, _bakerStake) = do
+                                    _theBakerInfo <- derefBakerInfo biRef
+                                    return FullBakerInfo{..}
+                            fullBakerInfos <- mapM mkFullBaker (Vec.fromList $ bakerStakes bsc)
+                            let bakerTotalStake = sum $ _bakerStake <$> fullBakerInfos
+                            return FullBakers{..}
```
EmilLa1, line 384: why is this changed? Is the formatter being weird?

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -456,33 +458,34 @@ getDefiniteSlotBakersP4 ::
...
+                                nextPaydaySlot
+                                (unOParam blockTimeParameters)
+                                pendingTimeParameters
+                    let nextPaydayLength =
+                            paydayTimeParameters
+                                ^. tpRewardPeriodLength
+                                    . to (epochToSlot . rewardPeriodEpochs)
+                    if blockEpoch + 1 == nextPayday && slot < nextPaydaySlot + nextPaydayLength
+                        then Just <$> getNextEpochBakers bs
+                        else return Nothing
```
EmilLa1, line 488: Same comment as above here.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+--
+-- This process has a bias towards bakers with smaller indexes. However, the chance of this bias
+-- affecting the result of any given election is less than 1 in 10^57, and so is effectively
+-- impossible.
+--
+-- Note: this implementation is linear in the number of bakers. By pre-processing the baker set,
+-- we could have a logarithmic algorithm. However, as the runtime is on the order of 5 us for 1000
+-- bakers and 45 us for 10000, this is not a worthwhile optimisation at the current time.
+getLeader ::
+    -- |(Non-empty) list bakers and their effective stakes.
```
EmilLa1, line 40:
```suggestion
    -- |(Non-empty) list of bakers and their effective stakes.
```

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
```diff
@@ -0,0 +1,95 @@
...
+    [(bakerInfo, Amount)] ->
```
EmilLa1, line 41: why is this polymorphic?

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/LeaderElectionTest.hs
```diff
@@ -0,0 +1,103 @@
...
+dummyVRFKeys :: VRF.KeyPair
+dummyVRFKeys = fst $ VRF.randomKeyPair (mkStdGen 0)
+
+dummyBlockNonce :: BlockNonce
+dummyBlockNonce = VRF.prove dummyVRFKeys ""
+
+-- |Tests for 'getLeader' on specific values with specific outcomes.
+-- These are intended as a regression test, as a change in the behaviour of 'getLeader' would likely
+-- cause them to fail.
+testGetLeader :: Spec
```
EmilLa1, line 31: Please elaborate a bit on the chosen values for the test.

EmilLa1 (approved):
Looks good, few minor comments.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
td202: The hash function generates values in the range `[0 .. 2^256 - 1]`. We want to sample a value in the range `[0 .. totalStake - 1]`, where `totalStake` is going to be at most `2^64`. If we sampled from a range `[0 .. k * totalStake - 1]` (for integer `k`), then taking the remainder modulo `totalStake` gives a uniformly distributed output (assuming the input is uniform). If we select `k` to be maximal such that `k * totalStake - 1 <= 2^256 -1` then we can see that there are at most 2^64 values in the range `[0 .. 2^256 - 1]` that are not in the range `[0 .. k * totalStake - 1]`. If we want to eliminate bias completely, then we can say, if we get such a value then we resample. But the chance of us getting such a value is (2^64 / 2^256) = 1.593... * 10^-58. Hence less than 1 in 10^57.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
td202: The bluepaper deals with real numbers in the interval `[0,1)`. For the implementation we instead use integers in `[0 .. totalStake)`, where `totalStake` is the sum of the effective stakes of the bakers (accounting for (non-passive) delegation and stake limiting). We then use the remainder to map the hash into this interval. This is easier to work with, but should achieve the same effect.

chrmatt: The real numbers would correspond to dividing the hash by 2^256. If you don't want to divide by a large number, the stakes could by multiplied by that. This would not have this bias discussed below. Would that be less efficient?

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -319,68 +319,69 @@ getSlotBakersP4 ::
...
     m FullBakers
-getSlotBakersP4 genData bs slot = do
-    SeedState{epochLength, epoch = blockEpoch} <- getSeedState bs
-    let epochToSlot :: Epoch -> Slot
-        epochToSlot e = fromIntegral e * (epochLength ^. unconditionally)
-    nextPayday <- getPaydayEpoch bs
-    let nextPaydaySlot = epochToSlot nextPayday
+getSlotBakersP4 genData bs slot =
+    getSeedState bs >>= \case
+        SeedStateV0{epochLength, epoch = blockEpoch} -> do
```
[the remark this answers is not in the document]
td202: I have added some documentation that this is not used in V1. The type system checks that `SeedStateV1` is actually impossible because of the constraint `SeedStateVersionFor (MPV m) ~ 'SeedStateVersion0`.

concordium-consensus/src/Concordium/KonsensusV1/LeaderElection.hs
td202: If we divide by 2^256 and work with real arithmetic, it doesn't change the bias. To conceptualise it, let's shrink the numbers and suppose that the hash generates a value in the range `[0 .. 7]` and there are three bakers, each with stake 1.

If we consider as implemented, the bakers win on the inputs as follows:
1. `{x ∈ [0 .. 7] | x mod 3 = 0} = {0,3,6}`
2. `{x ∈ [0 .. 7] | x mod 3 = 1} = {1,4,7}`
3. `{x ∈ [0 .. 7] | x mod 3 = 2} = {2,5}`

If we consider dividing by 8 and then splitting in regions we get:
1. `{x ∈[0 .. 7] | 0 ≤ x < 1/3} = {0,1,2}`
2. `{x ∈[0 .. 7] | 1/3 ≤ x < 2/3} = {3,4,5}`
3. `{x ∈[0 .. 7] | 1/3 ≤ x < 1} = {6,7}`

Even though the mapping is different, the bias is the same.

chrmatt: Thanks, makes sense. And the modulo thing is more efficient I assume?

We should probably change the bluepaper then and add some of your reasoning about the bias.

concordium-consensus/tests/scheduler/SchedulerTests/Helpers.hs
td202: Not really. They are the epoch length in slots and timestamp (in ms since UNIX epoch) for triggering the first epoch transition respectively. The actual values are not relevant to the tests.

chportma (approved):

eb-concordium (approved):

Issue #764, opened 2023-03-01
Some messages are labeled as invalid catch up messages


mh-concordium:
**Bug Description**
Prometheus exporter output contains 
`consensus_received_messages_total{message="catch-up status message",result="invalid"} 7
`
These messages are not actually invalid catch-up messages but some other messages.

**Steps to Reproduce**
Set up Prometheus and start the node. Observe the output in the browser.

**Expected Result**
Only the messages that node would not be able to understand would be marked as invalid. 

**Actual Result**
Number of these messages is different every time we start the node.

**Versions**
- Software Version: 5.2.3

PR #765, opened 2023-03-01
Fix catchup message labelling

limemloh:
## Purpose

Fixes https://github.com/Concordium/concordium-node/issues/764

## Changes

Label consensus ffi responses ContinueCatchUp as a valid message in prometheus metrics.

### DIFF
--- a/concordium-node/src/consensus_ffi/helpers.rs
+++ b/concordium-node/src/consensus_ffi/helpers.rs
@@ -265,11 +265,9 @@ impl ConsensusFfiResponse {
     /// Get the label. This is used when updating metrics of the prometheus
     /// exporter.
     pub fn label(&self) -> &str {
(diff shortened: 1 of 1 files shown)


concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -267,6 +267,8 @@ impl ConsensusFfiResponse {
     pub fn label(&self) -> &str {
         if self.is_successful() {
             "valid"
+        } else if let ConsensusFfiResponse::ContinueCatchUp = self {
+            "valid"
```
abizjak, line 271:
```suggestion
        if self.is_successful() || matches!(self, ConsensusFfIResponse::ContinueCatchUp) {
            "valid"
```

Issue #755, opened 2023-02-27
Extend the metrics of the Prometheus exporter part 2


limemloh:
This issue is tracking potential metrics to add to the built-in Prometheus exporter. The metrics should only be relevant for someone maintaining a node.

Continuation of #710

- [x] Total gRPC requests received. With label for gRPC response code and name of the endpoint called.
- [x] Histogram with gRPC response time.
  #742
- [x] Count of current gRPC requests in-flight.
  #769
- [x] Is baker. Is in baker committee and have baker keys setup.
- [x] Is finalizer. Is in finalization committee.
- [x] Total soft banned peers.
  #772
- [x] Current active lottery power. (Non-zero indicates the node is part of the baking committee).
  #779
- [x] Size of the in-memory transaction table, i.e. number of non-finalized transactions maintained by the node. (This is useful during TPS tests).
  #782
- [x] Total baked blocks (since process startup).
- [x] Total baked blocks becoming finalized.
  #784
- [x] Unrecognized protocol update. (Useful for setting up alerts indicating a node is outdated)
  #786

The above metrics have been implemented. A list of unimplemented metrics can be found here #790.

PR #769, opened 2023-03-03
Extend prometheus exporter with grpc_in_flight_requests metric

limemloh:
## Purpose

Related to #755.

## Changes

- Extend Prometheus exporter with `grpc_in_flight_requests` metric.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/stats_export_service.rs
+++ b/concordium-node/src/stats_export_service.rs
@@ -19,6 +19,7 @@ use prometheus::{
     TextEncoder,
 };
 use std::{net::SocketAddr, sync::RwLock, thread, time};
+use tower_http::metrics::in_flight_requests::InFlightRequestsCounter;
 
 use crate::configuration;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


concordium-node/src/stats_export_service.rs
```diff
@@ -299,6 +326,17 @@ impl StatsExportService {
         )?;
         registry.register(Box::new(grpc_request_response_time.clone()))?;
 
+        let grpc_pending_requests_gauge = IntGauge::with_opts(Opts::new(
```
abizjak, line 329: I don't think "pending" is an accurate description. It is requests in flight as the tower-http metric is named.

Pending to me implies that it is waiting to be  replied, but that is not what we are tracking here.

limemloh: I have renamed it now

concordium-node/src/stats_export_service.rs
```diff
@@ -19,6 +19,7 @@ use prometheus::{
     TextEncoder,
 };
 use std::{net::SocketAddr, sync::RwLock, thread, time};
+use tower_http::metrics::in_flight_requests::InFlightRequestsCounter;
```
abizjak, line 23:
```suggestion
```

abizjak: And maybe reformat.

abizjak (approved):

concordium-node/src/stats_export_service.rs
```diff
@@ -44,6 +46,27 @@ impl PrometheusStateData {
...
+/// Wrapper for implementing the prometheus::Collector trait for
+/// InFlightRequestsCounter, which syncs the prometheus gauge with the counter
+/// on scrape.
+struct GrpcInFlightRequestsCollector {
+    /// The prometheus gauge exposed.
+    pub gauge:   IntGauge,
+    /// Counter used to track in flight requests by
+    /// `tower_http::metrics::InFlightRequestsLayer`.
+    pub counter: InFlightRequestsCounter,
+}
```
Bargsteen, line 58: Why is the struct private but the fields public?
Also, what are the semantics of that? 🤔

Bargsteen (approved):
LGTM

concordium-node/src/stats_export_service.rs
abizjak: The semantics for anything like this is that all components of the path need to be visible for you to access the item.

But for this case there is no difference between pub and non/pub fields I think. Since the type cannot be exported we cannot access its fields outside of the module anyhow.

limemloh: Right, although it makes no difference, ill remove these to make the definition less noisy

PR #772, opened 2023-03-06
Extend metrics with committee status

limemloh:
## Purpose

Related to https://github.com/Concordium/concordium-node/issues/755.

## Changes

- Extend Prometheus exporter with metrics exposing the current baking, finalization committee status and the total number of soft-banned peers. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/bin/cli.rs
+++ b/concordium-node/src/bin/cli.rs
@@ -31,7 +31,9 @@ use concordium_node::{
     read_or_die,
     rpc::RpcServerImpl,
     spawn_or_die,
-    stats_export_service::{instantiate_stats_export_engine, StatsExportService},
+    stats_export_service::{
+        instantiate_stats_export_engine, StatsConsensusCollector, StatsExportService,
+    },
     utils::get_config_and_logging_setup,
 };
 use mio::{net::TcpListener, Poll};
@@ -139,6 +141,10 @@ async fn main() -> anyhow::Result<()> {
         notification_context,
     )?;
     info!("Consensus layer started");
+
+    // Start stats collecting which depend on querying consensus.
+    let consensus_collector = StatsConsensusCollector::new(consensus.clone())?;
+    node.stats.registry.register(Box::new(consensus_collector))?;
 
     // A flag to record that the import was stopped by a signal handler.
     let import_stopped = Arc::new(atomic::AtomicBool::new(false));
--- a/concordium-node/src/consensus_ffi/helpers.rs
+++ b/concordium-node/src/consensus_ffi/helpers.rs
@@ -336,36 +336,10 @@ impl TryFrom<i64> for ConsensusFfiResponse {
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum ConsensusIsInBakingCommitteeResponse {
-    ActiveInCommittee,
+    ActiveInCommittee = 0,
     NotInCommittee,
     AddedButNotActiveInCommittee,
     AddedButWrongKeys,
-}
-
-impl ConsensusIsInBakingCommitteeResponse {
-    /// Get the label. This is used when updating metrics of the prometheus
-    /// exporter.
-    pub fn label(&self) -> &'static str {
-        use ConsensusIsInBakingCommitteeResponse::*;
-        match self {
-            ActiveInCommittee => "active_in_committee",
-            NotInCommittee => "not_in_committee",
-            AddedButNotActiveInCommittee => "added_but_not_active_in_committee",
-            AddedButWrongKeys => "added_but_wrong_keys",
-        }
-    }
-
-    /// Get all possible labels. This is used to reset when updating metrics of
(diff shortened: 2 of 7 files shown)


concordium-node/src/bin/cli.rs
```diff
@@ -140,6 +142,10 @@ async fn main() -> anyhow::Result<()> {
     )?;
     info!("Consensus layer started");
 
+    // Start stats collecting which depend on querying consensus.
+    let consensus_collector = StatsConsensusCollector::new(consensus.clone())?;
+    node.stats.registry.register(Box::new(consensus_collector))?;
```
abizjak, line 147: Is there a real reason for starting the prometheus server before, and then breaking abstraction here to register an additional metric?

Could we not move the prometheus server start here?

concordium-node/src/consensus_ffi/helpers.rs
```diff
@@ -342,6 +342,31 @@ pub enum ConsensusIsInBakingCommitteeResponse {
...
+    }
+
+    /// Get all possible labels. This is used to reset when updating metrics of
+    /// the prometheus exporter.
+    pub const fn labels() -> &'static [&'static str] {
+        &[
+            "active_in_committee",
+            "not_in_committee",
+            "added_but_not_active_in_committee",
+            "added_but_wrong_keys",
```
abizjak, line 365: Perhaps use `ActiveInCommittee.label(), ...`, and so on. To avoid typos.

concordium-node/src/stats_export_service.rs
```diff
@@ -67,11 +72,84 @@ impl prometheus::core::Collector for GrpcInFlightRequestsCollector {
...
 
+/// Collector tracking stats which needs to be synced with consensus on scrape
+/// through the FFI.
+pub struct StatsConsensusCollector {
+    /// Reference to consensus to support consensus queries.
+    consensus:              ConsensusContainer,
+    /// The baking committee status of the node for the current best block.
+    baking_committee:       IntGaugeVec,
+    /// Whether the node is a member of the finalization committee for the
+    /// current best block.
```
abizjak, line 83: This is not really for best block. It is for the current finalization round.

concordium-node/src/stats_export_service.rs
```diff
@@ -67,11 +72,84 @@ impl prometheus::core::Collector for GrpcInFlightRequestsCollector {
...
+                 status",
+            )
+            .variable_label("status"),
+            &["status"],
+        )?;
+
+        let finalization_committee = IntGauge::with_opts(Opts::new(
+            "consensus_finalization_committee",
+            "The finalization commmittee status of the node for the current best block. The \
+             metric will have a value of 1 when member of the finalization committee",
```
abizjak, line 103: As above, this is not for best block. It is for the current state of finalization.

concordium-node/src/stats_export_service.rs
```diff
@@ -67,11 +72,84 @@ impl prometheus::core::Collector for GrpcInFlightRequestsCollector {
...
+        })
+    }
+
+    /// Update labelled metric to reflect the provided baking committee status.
+    fn set_baking_committe(&self, status: ConsensusIsInBakingCommitteeResponse) {
+        for status_label in ConsensusIsInBakingCommitteeResponse::labels() {
+            self.baking_committee.with_label_values(&[status_label]).set(0);
+        }
+        self.baking_committee.with_label_values(&[status.label()]).set(1);
+    }
```
abizjak, line 119: This seems rather awkward. Why did you not choose to instead have a gauge and just set it to the current value?

That seems lighter.

Bargsteen (approved):
LGTM if Ales' suggestions are addressed.

concordium-node/src/stats_export_service.rs
limemloh: Do you mean map the enum to a number and set the gauge to this number?

concordium-node/src/bin/cli.rs
limemloh: We are not really breaking the abstraction here, these metrics will just not be enabled until this point.
We could delay setting up the endpoint for scrapping until this point, but I don't really see a point.

Anyway, the challenge is that we need to pass stats to P2PNode to get access in the code related to networking.

concordium-node/src/stats_export_service.rs
abizjak: Yes.

concordium-node/src/bin/cli.rs
abizjak: Ok.

concordium-node/src/stats_export_service.rs
limemloh: For others reading, this was discussed offline and the conclusion was to have one metric, where the value represents the variant as suggested by Ales.
This avoids updating multi metrics and makes the code simpler. We did not find any major downsides to doing this.

PR #774, opened 2023-03-07
Advance round

EmilLa1:
## Purpose

Implementation of `advanceRound` from the blue paper
(Except the `makeBlock` call in the tail as this will be implemented in a future pr.)

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -31,4 +28,87 @@ class MonadMulticast m where
...
+    BakerId ->
+    -- |A collection of bakers and finalizers.
+    BakersAndFinalizers ->
+    -- |'True' if the consensus is part of the finalization committee.
+    -- Otherwise 'False'
+    Maybe FinalizerInfo
+isBakerFinalizer bakerId bakersAndFinalizers = do
+    -- This is O(n) but in principle we could do binary search here as the 'committeeFinalizers' are
+    -- sorted by ascending baker id.
+    Vector.find (\finalizerInfo -> finalizerBakerId finalizerInfo == bakerId) finalizers
```
chrmatt, line 52: It probably doesn't matter much since we don't have huge numbers of finalizers, but wouldn't it be easier if each baker just "remembers" when going to a new payday whether they are a finalizer?

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -31,4 +28,87 @@ class MonadMulticast m where
...
+  where
+    -- Reset the timer if this consensus instance is member of the
+    -- finalization committee for the current 'Epoch'.
+    resetTimerIfFinalizer bakerId currentTimeout = do
+        currentEpoch <- rsCurrentEpoch <$> use roundStatus
+        gets (getBakersForLiveEpoch currentEpoch) >>= \case
+            Nothing -> return () -- No bakers or finalizers could be looked up for the current 'Epoch' so we do nothing.
+            Just bakersAndFinalizers -> do
+                if isJust $! isBakerFinalizer bakerId bakersAndFinalizers
+                    then -- The consensus runer is a finalizer for the current epoch then we reset the timer
```
chrmatt, line 112: typo `runer`

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -203,6 +203,44 @@ data RoundStatus = RoundStatus
...
+-- |Advance the provided 'RoundStatus' to the provided 'Round'.
+-- If @Maybe (TimeoutCertificate, QuorumCertificate)@ is 'Nothing'
+-- then the round advances as it were a decendant of 'Round' where a 'QuorumCertificate' was successfully
+-- produced. Otherwise if @Just ..@ then it means that the old round timed out.
+-- All properties from the old 'RoundStatus' are being carried over to new 'RoundStatus'
+-- except the following.
+-- * 'rsCurrentRound' will become the provided 'Round'.
+-- * 'rsCurrentQuorumSignatureMessages' will be 'emptySignatureMessages'.
+-- * 'rsCurrentTimeoutSignatureMessages' will be 'emptySignatureMessages'.
+-- * 'rsPreviousRoundTC' will be either 'Absent' (if 'Nothing'
```
chrmatt, line 215: Missing paranthesis.

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -31,4 +28,87 @@ class MonadMulticast m where
...
+advanceRound newRound timedOut = do
+    myBakerId <- bakerId <$> view bakerIdentity
+    currentRoundStatus <- use roundStatus
+    -- Reset the timeout timer if the consensus runner is part of the
+    -- finalization committee.
+    resetTimerIfFinalizer myBakerId (rsCurrentTimeout currentRoundStatus)
+    -- Advance the round.
+    roundStatus .=! advanceRoundStatus newRound timedOut currentRoundStatus
+    -- Write the new round status to disk.
+    writeCurrentRoundStatus =<< use roundStatus
```
td202, line 99: We should have a function `setRoundStatus` (in Implementation.hs) that does both things because they should always be kept in sync.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -203,6 +203,44 @@ data RoundStatus = RoundStatus
...
+        { rsCurrentRound = toRound,
+          rsCurrentQuorumSignatureMessages = emptySignatureMessages,
+          rsCurrentTimeoutSignatureMessages = emptySignatureMessages,
+          rsPreviousRoundTC = previousRoundTC,
+          ..
+        }
+  where
+    previousRoundTC = case mTcQc of
+        Nothing -> Absent
+        Just (tc, qc) -> Present (tc, qc)
```
td202, line 242:
```suggestion
    previousRoundTC = maybe Absent Present mTcQc
```

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -203,6 +203,44 @@ data RoundStatus = RoundStatus
...
+    -- |The advanced 'RoundStatus'.
+    RoundStatus
+advanceRoundStatus toRound mTcQc RoundStatus{..} =
+    RoundStatus
+        { rsCurrentRound = toRound,
+          rsCurrentQuorumSignatureMessages = emptySignatureMessages,
+          rsCurrentTimeoutSignatureMessages = emptySignatureMessages,
+          rsPreviousRoundTC = previousRoundTC,
+          ..
+        }
```
td202, line 238: I think record update syntax would be more beneficial here.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -203,6 +203,44 @@ data RoundStatus = RoundStatus
...
+    -- 'TimeoutCertificate'.
+    -- In the latter case, then 'TimeoutCertificate' provided must be
+    -- the 'TimeoutCertificate' for the round that timed out and the 'QuorumCertificate'
+    -- is the one verifing the 'TimeoutCertificate'.
+    Maybe (TimeoutCertificate, QuorumCertificate) ->
+    -- |The 'RoundStatus' we are advancing from.
+    RoundStatus ->
+    -- |The advanced 'RoundStatus'.
+    RoundStatus
+advanceRoundStatus toRound mTcQc RoundStatus{..} =
```
td202, line 231: This function simply updates fields on a record, so I don't think there's any benefit to having it separated from `advanceRound` like this.

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
EmilLa1: No indeed I do not think it matters given the size of the finalization committee. But good idea @chrmatt! And GJ @td202 on that 781'er. Resolving conversation for now.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -203,6 +203,44 @@ data RoundStatus = RoundStatus
...
+-- If @Maybe (TimeoutCertificate, QuorumCertificate)@ is 'Nothing'
+-- then the round advances as it were a decendant of 'Round' where a 'QuorumCertificate' was successfully
+-- produced. Otherwise if @Just ..@ then it means that the old round timed out.
+-- All properties from the old 'RoundStatus' are being carried over to new 'RoundStatus'
+-- except the following.
+-- * 'rsCurrentRound' will become the provided 'Round'.
+-- * 'rsCurrentQuorumSignatureMessages' will be 'emptySignatureMessages'.
+-- * 'rsCurrentTimeoutSignatureMessages' will be 'emptySignatureMessages'.
+-- * 'rsPreviousRoundTC' will be either 'Absent' (if 'Nothing'
+advanceRoundStatus ::
```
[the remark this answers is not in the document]
EmilLa1: Yea. Well this is an implementation of the `advanceRound` of the blue paper. And yes, on its own it only handles the TC case. 

However when progressing via a QC (`processQuorumSignature` in the blue paper`) then we set the highest QC if we can construct it and then advance the round. 

**I am extending this function now so it handles both cases.**

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
EmilLa1: Fixed doc.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
EmilLa1: Changed the function so this is no longer relevant. But thanks for the hint!

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
EmilLa1: I added a setRoundStatus in the Implementation.hs module now.

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -31,4 +26,67 @@ class MonadMulticast m where
...
+    -- |If we are advancing from a round that timed out
+    -- then this will be @Left 'TimeoutCertificate, 'QuorumCertificate')@
+    -- The 'TimeoutCertificate' is from the round we're
+    -- advancing from and the associated 'QuorumCertificate' verifies it.
+    --
+    -- Otherwise if we're progressing via a 'QuorumCertificate' then @Right QuorumCertificate@
+    -- should be the QC we're advancing round via.
+    Either (TimeoutCertificate, QuorumCertificate) QuorumCertificate ->
+    m ()
+advanceRound newRound timedOut = do
```
chportma, line 72: `timedOut` also has a value if we haven't timed out, so maybe this could be better named. Maybe `newCertificates`?

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -31,4 +26,67 @@ class MonadMulticast m where
...
+  where
+    -- Reset the timer if this consensus instance is member of the
+    -- finalization committee for the current 'Epoch'.
+    resetTimerIfFinalizer bakerId currentTimeout currentEpoch = do
+        gets (getBakersForLiveEpoch currentEpoch) >>= \case
+            Nothing -> return () -- No bakers or finalizers could be looked up for the current 'Epoch' so we do nothing.
+            Just bakersAndFinalizers -> do
+                when (isJust $ finalizerByBakerId (bakersAndFinalizers ^. bfFinalizers) bakerId) $
+                    -- The consensus runner is a finalizer for the current epoch then we reset the timer
+                    resetTimer currentTimeout
```
chportma, line 92: A problem here this is that people leaving the finalization committee might not start a timer. But if their highestQC is still in the previous epoch, we would like these people to throw the timeout. So they have to start the timer for that.

The bluepaper is probably wrong in this regard. I did not consider edge cases of people joining and leaving the finalization committee when writing it up. It is complicated enough as it is, and I prefer leave those cases to you.

There is a similar issue with people joining the finalization committee. Being a finalizer is not enough to be able to throw a timeout when the timer expires. We also have to have a highestQC in an epoch where we are a finalizer. One option is to take care of this is when throwing the timeout (`uponTimeoutEvent` in the bluepaper). An alternative is to not even start a timer in that case. The first is probably the better option, because the highestQC can change between the moment were we start the timer and when it expires. But you might also want it to be consistent with the case of people leaving, and start timers based on the epoch of the highestQC and not of the currentepoch. Anyway, since I haven't seen the code for `uponTimeoutEvent` yet, I am just pointing this out here so that you can check what is being done.

Note that I wrote a post in slack about the case of people joining: https://concordium.slack.com/archives/CGFCLTN8Z/p1677852233604619

chportma (changes requested):

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
EmilLa1: Yes. I agree that the current name is..... let's say "not ideal" :) .. 

I'll update it.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
EmilLa1: It's gonna do more when it's merged with: https://github.com/Concordium/concordium-node/pull/787

In particular it's also gonna clear up the signature messages for the current round.

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
EmilLa1: Always resetting the timer now. I've put some documentation that emphasizes the importance of when throwing a timeout it is checked that the consensus runner is either part of the current epoch or was part of the prior epoch. 

This will be handled in a pr based on this branch: https://github.com/Concordium/concordium-node/tree/upon-timeout-event

eb-concordium (approved):

td202 (approved):

chportma (approved):

PR #780, opened 2023-03-08
Advance epoch

EmilLa1:
## Purpose

Implementation of the `advanceEpoch` protocol from the blue paper. 

Note that computing the new leadership election nonce is left as a todo as it is already implemented on this branch: https://github.com/Concordium/concordium-node/tree/block-verification

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
@@ -10,7 +10,6 @@ import qualified Data.Vector as Vector
 
 import Lens.Micro.Platform
 
-import qualified Concordium.KonsensusV1.TreeState.LowLevel as LowLevel
 import Concordium.KonsensusV1.TreeState.Implementation
 import Concordium.KonsensusV1.TreeState.LowLevel (MonadTreeStateStore (writeCurrentRoundStatus))
 import qualified Concordium.KonsensusV1.TreeState.LowLevel as LowLevel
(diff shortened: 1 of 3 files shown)


concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -90,3 +90,37 @@ advanceRound newRound timedOut = do
...
+-- the provided 'Epoch' and 'FinalizationEntry'
+-- TODO: implement.
+computeLeadershipElectionNonce ::
+    -- |The 'Epoch' to compute the 'LeadershipElectionNonce' for.
+    Epoch ->
+    -- |The witness for the new 'Epoch'
+    FinalizationEntry ->
+    -- |The new 'LeadershipElectionNonce'
+    LeadershipElectionNonce
+computeLeadershipElectionNonce epoch finalizationEntry = undefined
```
chportma, line 104: This doesn't look quite right. To compute the new LE nonce you will need the temporary nonce that you update every new block and the baker set. FinalizationEntry is probably not needed. But since you haven't implemented it yet, maybe you don't care if this needs to be changed. So ignore this comment.

chportma (approved):

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -245,6 +245,32 @@ advanceRoundStatus toRound (Right qc) RoundStatus{..} =
           ..
         }
 
+-- |Advance the proived 'RoundStatus' to the provided 'Epoch'.
```
eb-concordium, line 248:
```suggestion
-- |Advance the provided 'RoundStatus' to the provided 'Epoch'.
```

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Types.hs
```diff
@@ -245,6 +245,32 @@ advanceRoundStatus toRound (Right qc) RoundStatus{..} =
...
+-- In particular this does the following to the provided 'RoundStatus'
+--
+-- * Set the 'rsCurrentEpoch' to the provied 'Epoch'
```
eb-concordium, line 251:
```suggestion
-- * Set the 'rsCurrentEpoch' to the provided 'Epoch'
```

PR #783, opened 2023-03-09
Extend metrics baked blocks

limemloh:
## Purpose

Related to #755.

## Changes

- Add metric showing the total baked blocks by the node.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/consensus_ffi/ffi.rs
+++ b/concordium-node/src/consensus_ffi/ffi.rs
@@ -265,7 +265,8 @@ pub struct execute_block {
 pub struct ExecuteBlockCallback(*mut execute_block);
 
 type LogCallback = extern "C" fn(c_char, c_char, *const u8);
-type BroadcastCallback = extern "C" fn(i64, u32, *const u8, i64);
+type BroadcastCallback =
+    unsafe extern "C" fn(*const BroadcastCallbackContext, i64, u32, *const u8, i64);
 type CatchUpStatusCallback = extern "C" fn(u32, *const u8, i64);
 type DirectMessageCallback = extern "C" fn(
     peer_id: PeerIdFFI,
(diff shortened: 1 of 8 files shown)


concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -2990,62 +3002,65 @@ pub extern "C" fn on_finalization_message_catchup_out(
...
+        Err(e) => {
+            error!("{}", e);
+            return;
+        }
+    };
+
+    if let CallbackType::Block = callback_type {
+        let context = unsafe { context_ptr.read() };
+        context.metric_baked_blocks.inc();
+    }
```
abizjak, line 3062: Why not do this inside the sending_callback? The type is deserialized there as well.

abizjak:
After the discussion at the standup yesterday I was thinking about another approach using the existing `notifyBlock` and `notifyFinalizedBlock.`

This would allow us to have two metrics, one for blocks produced, and another for blocks that have ended up on the finalized chain.

The idea is to modify the `notifyCallback` to report also whether the block we are notify is ours or not.

This requires only minor modifications in Haskell, namely modifying

```haskell
    handleBlock bp = liftSkov $ do
        lift (asks (notifyBlockArrived . mvCallbacks)) >>= \case
            Nothing -> return ()
            Just notifyCallback -> do
                versionsRef <- lift (asks mvVersions)
                versions <- liftIO (readIORef versionsRef)
                ownBaker <- lift (asks (fmap (bakerId . bakerIdentity) . mvBaker))
                let latestEraGenesisHeight =
                        case Vec.last versions of
                            EVersionedConfiguration vc -> vcGenesisHeight vc
                let height = localToAbsoluteBlockHeight latestEraGenesisHeight (bpHeight bp)
                let isOurs = case ownBaker of
                      Nothing -> False
                      Just ourId -> Just ourId == (blockBaker <$> blockFields (_bpBlock bp))
                liftIO (notifyCallback isOurs (bpHash bp) height)
````

and similarly for `handleFinalize`.

Then we handle the metric updates in the context we already do when counting received blocks. Does this make sense?

PR #784, opened 2023-03-10
Extend metrics baked blocks

limemloh:
## Purpose

Related to #755.

## Changes

- Add metric showing the total baked blocks by the node.
- Add metric showing the total finalized baked blocks by the node.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/MultiVersion.hs
+++ b/concordium-consensus/src/Concordium/MultiVersion.hs
@@ -47,7 +47,7 @@ import Concordium.Afgjort.Finalize.Types
 import Concordium.Birk.Bake
 import Concordium.GlobalState
 import Concordium.GlobalState.Block
-import Concordium.GlobalState.BlockPointer (BlockPointerData (..))
+import Concordium.GlobalState.BlockPointer (BlockPointer (..), BlockPointerData (..))
 import Concordium.GlobalState.Finalization
 import Concordium.GlobalState.Parameters
 import Concordium.GlobalState.TreeState (PVInit (..), TreeStateMonad (getLastFinalizedHeight))
@@ -102,7 +102,11 @@ instance
                         case Vec.last versions of
                             EVersionedConfiguration vc -> vcGenesisHeight vc
                 let height = localToAbsoluteBlockHeight latestEraGenesisHeight (bpHeight bp)
-                liftIO (notifyCallback (bpHash bp) height)
+                ownBaker <- lift (asks (fmap (bakerId . bakerIdentity) . mvBaker))
+                let isHomeBaked = case ownBaker of
+                        Nothing -> False
+                        Just ourBakerId -> Just ourBakerId == (blockBaker <$> blockFields (_bpBlock bp))
+                liftIO (notifyCallback (bpHash bp) height isHomeBaked)
     handleFinalize _ lfbp bps = liftSkov $ do
         lift (asks (notifyBlockFinalized . mvCallbacks)) >>= \case
             Nothing -> return ()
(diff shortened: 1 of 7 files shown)


Bargsteen (approved):
LGTM

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -102,7 +102,11 @@ instance
                         case Vec.last versions of
                             EVersionedConfiguration vc -> vcGenesisHeight vc
                 let height = localToAbsoluteBlockHeight latestEraGenesisHeight (bpHeight bp)
-                liftIO (notifyCallback (bpHash bp) height)
+                ownBaker <- lift (asks (fmap (bakerId . bakerIdentity) . mvBaker))
+                let isHomeBaked = case ownBaker of
+                        Nothing -> False
+                        Just ourBakerId -> Just ourBakerId == (blockBaker <$> blockFields (_bpBlock bp))
+                liftIO (notifyCallback (bpHash bp) height isHomeBaked)
```
abizjak, line 109: This is not a comment about this code directly, but about an alternative.

It would be good to state here clearly as well (and not just in the prometheus documentation) that home-baked has a specific meaning about baker-ids and the caveat about catchup.

It would also be good to mention an alternative for this, namely that we could change `handleBlock` to take `homeBaked` as an argument which would directly tell us whether this is produced during baking or not. This is easy for handleBlock, however it is not easy for handleFinalize, although it woudl be possible to keep track of this, but that would add complexity.

Can you please add this comment here explaining why the alternative was not chosen as well?

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -173,11 +184,11 @@ data Callbacks = Callbacks
...
       notifyRegenesis :: Maybe BlockHash -> IO (),
       -- |Notify a block was added to the tree. The arguments are
-      -- the hash of the block, and its absolute height.
-      notifyBlockArrived :: Maybe (BlockHash -> AbsoluteBlockHeight -> IO ()),
+      -- the hash of the block, its absolute height and whether this node was the baker.
+      notifyBlockArrived :: Maybe (BlockHash -> AbsoluteBlockHeight -> Bool -> IO ()),
       -- |Notify a block was finalized. The arguments are the hash of the block,
-      -- and its absolute height.
-      notifyBlockFinalized :: Maybe (BlockHash -> AbsoluteBlockHeight -> IO ())
+      -- its absolute height and whether this node was the baker.
```
abizjak, line 190:
```suggestion
      -- its absolute height and whether the block was produced by the baker id configured for this node.
```

There is a difference between this sentence and what you wrote. Namely the same node can switch what baker id it is running with for example.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -173,11 +184,11 @@ data Callbacks = Callbacks
       -- if an unrecognized update took effect.
       notifyRegenesis :: Maybe BlockHash -> IO (),
       -- |Notify a block was added to the tree. The arguments are
-      -- the hash of the block, and its absolute height.
-      notifyBlockArrived :: Maybe (BlockHash -> AbsoluteBlockHeight -> IO ()),
+      -- the hash of the block, its absolute height and whether this node was the baker.
```
abizjak, line 187:
```suggestion
      -- the hash of the block, its absolute height and whether the block was produced by the baker id configured for this node.
```

docs/prometheus-exporter.md
```diff
@@ -213,3 +213,16 @@ Is only non-zero when active member of the baking committee.
 ### `consensus_non_finalized_transactions`
 
 The current number of non-finalized transactions stored by the node.
+
+
+### `consensus_baked_blocks_total`
+
+Total number of blocks baked by the node since startup.
+
+Blocks received as part of catchup are also counted, when the block was baked by the same baker ID as the node.
```
abizjak, line 222:
```suggestion
Blocks received as part of catchup are also counted, when the block was baked by the same baker ID as the node is configured with.
```

docs/prometheus-exporter.md
```diff
@@ -213,3 +213,16 @@ Is only non-zero when active member of the baking committee.
...
+
+### `consensus_finalized_baked_blocks_total`
+
+Total number of finalized blocks baked by the node since startup.
+
+Finalized blocks received as part of catchup are also counted, when the block was baked by the same baker ID as the node.
```
abizjak, line 228:
```suggestion
Finalized blocks received as part of catchup are also counted, when the block was baked by the same baker ID as the node is configured with.
```

abizjak (commented):
I would like to review the phrasing of additional documentation I requested.

The code looks good otherwise.

concordium-consensus/src/Concordium/MultiVersion.hs
limemloh: I have added a comment for `handleBlock` with the details, please review this

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -92,6 +92,14 @@ instance
...
     Skov.HandlerConfigHandlers UpdateHandler (VersionedSkovM fc pv)
     where
+    -- Notice that isHomeBaked (in the code below) represents whether this block is baked by the
+    -- baker ID of this node and it could be the case that the block was not baked by this node,
+    -- if another node using the same baker ID.
+    -- The information is used to count the number of baked blocks exposed by a prometheus metric,
+    -- which alternatively could be implemented by observing the broadcastBlock callback, which
+    -- would make it independent of the baker ID. However, The broadcastBlock callback cannot be
+    -- used for counting finalized baked blocks, so for simplicity and consistency the current
+    -- approach depending on baker ID was chosen.
```
abizjak, line 102:
```suggestion
    -- The information is used to count the number of baked blocks exposed by a prometheus metric.
    -- An alternative implementation would be to extend the @onBlock@ handler (part of @OnSkov@) 
    -- to take an extra argument indicating whether the block was just baked or processed as part of a received
    -- block. This would mean that only blocks baked since start by this node would be counted,
    -- not blocks received as part of catchup. However the same cannot be done for finalized blocks as easily
    -- and so for consistency between these two methods we chose to also count blocks received as part of catchup
    -- in both.
```

abizjak (approved):

PR #786, opened 2023-03-13
Extend metrics with indicator for unsupported pending protocol updates

limemloh:
## Purpose

Related to https://github.com/Concordium/concordium-node/issues/755.

## Changes

- add new callback to consensus to provide information to the metric, which indicates unsupported pending protocol updates.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -405,9 +405,9 @@ startConsensus ::
     Ptr NotifyContext ->
     -- |The callback used to invoke upon new block arrival, and new finalized blocks.
     FunPtr NotifyCallback ->
-    -- |Context for when signalling a unsupported protocol update is pending or effective.
+    -- |Context for when signalling a unsupported protocol update is pending.
     Ptr NotifyUnsupportedUpdatesContext ->
-    -- |The callback used to signal a unsupported protocol update is pending or effective.
+    -- |The callback used to signal a unsupported protocol update is pending.
     FunPtr NotifyUnsupportedUpdatesCallback ->
     -- |Handler for generated messages
     FunPtr BroadcastCallback ->
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -325,10 +325,8 @@ type NotifyCallback =
     Word64 ->
     -- |Absolute block height of either the arrived block or finalized depending on the type of event.
     Word64 ->
-    -- |Byte where a value of 1 indicates the block arrived/finalized was baked by this node.
-    Word8 ->
-    -- |Byte where a value of 1 indicates an unsupported pending protocol update is finalized.
-    -- Only used when notifying a finalized block.
+    -- |Byte where a value of 1 indicates the block arrived/finalized was baked by the baker ID
+    -- setup for this node.
     Word8 ->
     IO ()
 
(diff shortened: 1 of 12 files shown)


concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -124,6 +124,12 @@ instance
     -- The information is used to count the number of finalized baked blocks exposed by a prometheus
     -- metric.
     handleFinalize _ lfbp bps = liftSkov $ do
+        -- Check for protocol update.
+        result <- checkForProtocolUpdate
+        let haveUnsupportedUpdate = case result of
+                CFPURUnsupportedPending -> True
```
abizjak, line 130: Shouldn't we return true for both unsupported variants?

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -729,7 +750,8 @@ startupSkov genesis = do
                         case nextPV of
                             Nothing -> do
                                 mvrLogIO $ activateConfiguration newEConfig
-                                liftSkovUpdate newEConfig' checkForProtocolUpdate
+                                _ <- liftSkovUpdate newEConfig' checkForProtocolUpdate
```
abizjak, line 753: This is not exactly ideal since if the user has shut down the node and then restarted it again the flag will say 0 even though the node is in a state where it has an unrecognized protocol update.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -454,6 +459,19 @@ newGenesis (PVGenesisData (gd :: GenesisData pv)) vcGenesisHeight = case consens
                     notifyRegenesis (Just (genesisBlockHash gd))
     ConsensusV1 -> error "New consensus version not implemented yet." -- TODO: implement
 
+-- |Outcome for checking for protocol update.
+data CheckForProtocolUpdateResult
+    = -- | No protocol update are either pending or effective at this time.
```
Bargsteen, line 464:
```suggestion
    = -- | No protocol updates are either pending or effective at this time.
```

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -454,6 +459,19 @@ newGenesis (PVGenesisData (gd :: GenesisData pv)) vcGenesisHeight = case consens
...
+      CFPURNone
+    | -- | A protocol update is effective at this time, but is unsupported by this node version.
+      CFPURUnsupportedEffective
+    | -- | A protocol update is effective at this time and is unsupported by this node version.
```
Bargsteen, line 468:
```suggestion
    | -- | A protocol update is effective at this time and is supported by this node version.
```

Bargsteen (commented):
Looks fine to me if Ales' comments are addressed 😊

concordium-consensus/src/Concordium/MultiVersion.hs
limemloh: I have taken a different approach now, by introducing a dedicated callback just for signalling a pending unsupported protocol update.

concordium-consensus/src/Concordium/MultiVersion.hs
limemloh: I have changed the design, so now it is calling a callback directly for signalling unsupported pending updates.
But the other case where it is "unsupported effective" will only be visible for a short time, since the node shuts down right after.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -348,6 +348,19 @@ pub struct NotificationHandlers {
...
+/// protocol update. The callback is called with:
+/// - the context
+/// - effective time of unsupported update (milliseconds since unix epoch).
+type NotifyUnsupportedUpdatesCallback =
+    unsafe extern "C" fn(*mut NotifyUnsupportedUpdatesContext, u64);
+
+pub struct NotifyUnsupportedUpdatesContext {
+    /// If non-zero the value represents the effective timestamp of unsupported
+    /// protocol update as milliseconds since unix epoch.
+    pub unsupported_pending_protocol_version: prometheus::IntGauge,
```
abizjak, line 361: Is there a specific reason why we are not using
```
GenericGauge<AtomicU64>
```
instead?

Then we don't have to do `as i64`

limemloh: The only reason is to follow the type of the other metrics with timestamp, but it would not make any difference in the output of a scrape.
I'll change it

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -573,23 +577,29 @@ checkForProtocolUpdate = liftSkov body
...
+                                "An unsupported protocol update ("
+                                    ++ err
+                                    ++ ") will take effect at "
+                                    ++ show (timestampToUTCTime $ transactionTimeToTimestamp ts)
+                                    ++ ": "
+                                    ++ showPU pu
+                        callbacks <- lift $ asks mvCallbacks
+                        case notifyUnsupportedProtocolUpdate callbacks of
+                            Just notifyCallback -> liftIO $ notifyCallback $ transactionTimeToTimestamp ts
+                            Nothing -> return ()
```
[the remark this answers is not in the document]
limemloh: Right, I'll change this

concordium-consensus/src/Concordium/External.hs
```diff
@@ -353,6 +356,27 @@ mkNotifyBlockFinalized f = \bh height isHomeBaked -> do
...
+    -- |Effective timestamp of unsupported protocol update as milliseconds since unix epoch.
+    Word64 ->
+    IO ()
+
+foreign import ccall "dynamic"
+    callNotifyUnsupportedUpdates ::
+        FunPtr NotifyUnsupportedUpdatesCallback ->
+        NotifyUnsupportedUpdatesCallback
+
+-- |Serialize effective timestamp of unsupported protocol update, to milliseconds since unix epoch.
```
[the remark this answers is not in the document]
limemloh: I have rephrased the comment

abizjak (approved):

Bargsteen (approved):
LGTM

PR #788, opened 2023-03-13
Propagate changes from restructuring rust-src.

abizjak:
## Purpose

Propagate changes from restructuring `rust-src` in https://github.com/Concordium/concordium-base/pull/337

The Haskell changes are there due to updates to base that were not propagated before.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -49,7 +49,7 @@ import Concordium.MultiVersion (
     makeMultiVersionRunner,
  )
 import qualified Concordium.MultiVersion as MV
-import Concordium.Queries (BakerStatus (..), BlockHashInput (BHIGiven))
+import Concordium.Queries (BakerStatus (..))
 import qualified Concordium.Queries as Q
 import Concordium.Scheduler.Types
 import Concordium.Skov (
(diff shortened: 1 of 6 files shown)


eb-concordium (approved):

Bargsteen (approved):

concordium-consensus/src/Concordium/External.hs
```diff
@@ -60,6 +60,7 @@ import Concordium.Skov (
  )
 import Concordium.TimerMonad (ThreadTimer)
 import qualified Concordium.Types.InvokeContract as InvokeContract
+import qualified Concordium.Types.Queries as TQ
```
EmilLa1, line 63: I think `Queries` would be more useful at call sites.

concordium-consensus/src/Concordium/External/GRPC2.hs
```diff
@@ -34,6 +34,7 @@ import qualified Concordium.GlobalState.ContractStateV1 as StateV1
 import Concordium.ID.Types
 import qualified Concordium.Queries as Q
 import Concordium.Types
+import qualified Concordium.Types.Queries as TQ
```
EmilLa1, line 37: Same comment.

concordium-consensus/src/Concordium/External.hs
abizjak: I followed the `as Q` for other Queries we import. But since you commented it is now Queries.

concordium-consensus/src/Concordium/External/GRPC2.hs
abizjak: I followed the `as Q` for other Queries we import. But since you commented it is now Queries.

EmilLa1 (approved):
Looks good. Great job!

PR #799, opened 2023-03-20
Add a health check option to check minimum number of peers.

### LINKED ISSUE #798: Revise node health check
**Task description**

In order to simplify node deployments the node health check should be revised to correspond to the checks currently performed by the node-liveness-checker, so that the latter may be deprecated.

abizjak:
## Purpose

Revise the health check.

Closes #798 

## Changes

- Check for minimum number of peers if so configured.
- If configured as a baker, make sure the node is in the committee.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased changes
 
+- Add an option `--grpc2-health-min-peers` (environment variable `CONCORDIUM_NODE_GRPC2_HEALTH_MIN_PEERS`)
+  that causes the grpc V2 health endpoint to check minimum number of peers.
+- Extend the node health check so that if the node is configured with baker
+  credentials then it is required to be in the baking committee for it to be
+  considered healthy.
+
 ## 5.3.0
 
 - Extend Prometheus exporter with metrics: `grpc_request_duration_seconds`, `grpc_in_flight_requests`, `consensus_baking_committee`, `consensus_finalization_committee`, `consensus_baking_lottery_power`, `consensus_baked_blocks_total`, `consensus_finalized_baked_blocks_total`, `network_soft_banned_peers_total`, `consensus_non_finalized_transactions` and `consensus_unsupported_pending_protocol_version` see [docs/prometheus-exporter.md](https://github.com/Concordium/concordium-node/blob/main/docs/prometheus-exporter.md) for more details.
--- a/concordium-node/src/health.rs
+++ b/concordium-node/src/health.rs
@@ -1,7 +1,12 @@
 //! Implementation of the health check service that is part of the GRPC2
 //! interface.
 
-use crate::consensus_ffi::consensus::ConsensusContainer;
+use crate::{
+    common::PeerType,
+    consensus_ffi::{consensus::ConsensusContainer, helpers::ConsensusIsInBakingCommitteeResponse},
+    p2p::P2PNode,
+};
+use std::sync::Arc;
 
 include!(concat!(env!("OUT_DIR"), "/concordium.health.rs"));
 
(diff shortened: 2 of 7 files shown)


concordium-node/src/health.rs
```diff
@@ -42,6 +49,26 @@ impl health_server::Health for HealthServiceImpl {
             return Err(tonic::Status::unavailable("Last finalized block is too far behind."));
         }
 
+        if let Some(min_allowed_peers) = self.health_min_peers {
+            let num_peers = self.node.get_peer_stats(Some(PeerType::Node)).len();
+            if num_peers < min_allowed_peers {
+                return Err(tonic::Status::unavailable(format!(
+                    "The only has {} peers, but is required to have at least {}.",
```
Thahara, line 56:
```suggestion
                    "The node only has {} peers, but is required to have at least {}.",
```

docs/grpc2.md
```diff
@@ -35,6 +35,11 @@ If these are enabled then the following options become available
   (`CONCORDIUM_NODE_GRPC2_HEALTH_MAX_FINALIZED_DELAY`) is a configuration for the
   `GetNodeHealth` endpoint. It specifies (in seconds) the maximum delay in last
   finalized block time before the health check fails.
+- `--grpc2-health-min-peers` (default is 5min)
```
EmilLa1, line 38: 5 minutes? Not sure I follow.

Thahara (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased changes
 
+- Add an option `--grpc2-health-min-peers` (environment variable `CONCORDIUM_NODE_GRPC2_HEALTH_MIN_PEERS`)
```
EmilLa1, line 5: Couldn't it just always be checking whether the `desired_peers` is fulfilled? As I recall, then the node will keep trying get more peers until the number of desired peers is fulfilled.  

What will the concrete value be for such an option as compared to `desired_peers`? Will it be lower? And if so why? If a node is not able to get its `desired` peers should it then be considered healthy with respect to its settings?

docs/grpc2.md
abizjak: Copy paste issue.

CHANGELOG.md
abizjak: I think using desired peers is too limiting. Just because the node does not have the desired number of peers does not make it unhealthy and is not cause for alerts for example.

This bound should be "minimum required" and in practice for startup we sometimes need to set minimum lower than  desired.

docs/grpc2.md
abizjak: FIxed by Stefan.

CHANGELOG.md
abizjak: Having this option does still give the operator an option to set it to the same value as desired peers. But there are use-cases where they should be different.

fdibat: I would personally refrain from using the desired peers since a node can be healthy and operating before having all desired peers. Maybe it could make sense to have the desired peers as default if the environment variable or option is not defined.

EmilLa1: Okay, I was just wondering if this would ever be lower than 'desired'. But if that is the case then lgtm.

EmilLa1 (approved):

CHANGELOG.md
EmilLa1: Well.. I presume the node can be "healthy and operating" also when lower than "minimum peers" unless you always set this to 1.

abizjak: It's as if there is ambiguity and usage dependent scenarios that we have to deal with :) 

Thanks for the feedback.

PR #812, opened 2023-03-29
Extend reporter metrics with recently connected peer counts.

vikt0r0:
## Purpose

Extends the prometheus reporter interface by adding a metric of recently connected peer counts, labelled by the bucket in which the peer is maintained. The following is an example of the relevant snippet of the `/metrics`-endpoint after these changes:

```
...
# HELP peer_bucket_size The number of peers that recently connected to the bootstrapper labelled by the number of the bucket in which they are contained
# TYPE peer_bucket_size gauge
peer_bucket_size{bucket="0"} 2
...
```

## Changes

Extend reporter and update metrics when elements are added to and removed from buckets.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased changes
 - Fix a bug where the prometheus interface was not exposed in the
   bootstrapper when `CONCORDIUM_NODE_PROMETHEUS_LISTEN_PORT` was set.
-- Extend Prometheus exporter with metric `recent_peers`, see
+- Extend Prometheus exporter with metric `peer_bucket_size`, see
   [docs/prometheus-exporter.md](https://github.com/Concordium/concordium-node/blob/main/docs/prometheus-exporter.md) for more details.
 
 ## 5.3.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 # Changelog
 
 ## Unreleased changes
-- Fix a bug where the prometheus interface was not exposed in the
-  bootstrapper when `CONCORDIUM_NODE_PROMETHEUS_LISTEN_PORT` was set.
 - Extend Prometheus exporter with metric `peer_bucket_size`, see
   [docs/prometheus-exporter.md](https://github.com/Concordium/concordium-node/blob/main/docs/prometheus-exporter.md) for more details.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## Unreleased changes
+- Fix a bug where the prometheus interface was not exposed in the
+  bootstrapper when `CONCORDIUM_NODE_PROMETHEUS_LISTEN_PORT` was set.
+- Extend Prometheus exporter with metric `recent_peers`, see
+  [docs/prometheus-exporter.md](https://github.com/Concordium/concordium-node/blob/main/docs/prometheus-exporter.md) for more details.
 
 ## 5.3.1
 
--- a/concordium-node/src/bin/bootstrapper.rs
+++ b/concordium-node/src/bin/bootstrapper.rs
@@ -16,8 +16,10 @@ use concordium_node::{
 };
 
 use concordium_node::stats_export_service::start_push_gateway;
+use std::net::SocketAddr;
 
-fn main() -> anyhow::Result<()> {
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
     let (mut conf, app_prefs) = get_config_and_logging_setup()?;
     conf.connection.max_allowed_nodes = Some(conf.bootstrapper.max_nodes);
     let data_dir_path = app_prefs.get_data_dir();
@@ -44,12 +46,22 @@ fn main() -> anyhow::Result<()> {
         conf.common.id,
         &conf,
         PeerType::Bootstrapper,
-        stats_export_service,
+        stats_export_service.clone(),
         regenesis_arc,
     )
     .context("Failed to create the network node.")?;
 
     start_push_gateway(&conf.prometheus, &node.stats, node.id());
+
+    if let Some(plp) = conf.prometheus.prometheus_listen_port {
+        let (sender, _) = tokio::sync::broadcast::channel(1);
+        tokio::spawn(async move {
+            stats_export_service
+                .start_server(SocketAddr::new(conf.prometheus.prometheus_listen_addr, plp), sender)
+                .await
+        });
+    }
+
     // Set the startime in the stats.
     node.stats.node_startup_timestamp.set(node.start_time.timestamp_millis());
 
--- a/concordium-node/src/bin/bootstrapper.rs
+++ b/concordium-node/src/bin/bootstrapper.rs
@@ -54,6 +54,7 @@ async fn main() -> anyhow::Result<()> {
     start_push_gateway(&conf.prometheus, &node.stats, node.id());
 
     if let Some(plp) = conf.prometheus.prometheus_listen_port {
+        // We ignore the receiver since we do not care about graceful shutdown here.
         let (sender, _) = tokio::sync::broadcast::channel(1);
         tokio::spawn(async move {
             stats_export_service
--- a/concordium-node/src/connection/mod.rs
+++ b/concordium-node/src/connection/mod.rs
@@ -586,8 +586,16 @@ impl Connection {
     pub fn populate_remote_end_networks(&mut self, peer: RemotePeer, networks: &Networks) {
         self.remote_end_networks.extend(networks.iter());
 
+        let increment_bucket = |a: u64| {
+            self.handler.stats.recent_peers.with_label_values(&[a.to_string().as_str()]).inc();
+        };
+
         if self.remote_peer.peer_type != PeerType::Bootstrapper {
-            write_or_die!(self.handler.buckets()).insert_into_bucket(peer, networks.to_owned());
+            write_or_die!(self.handler.buckets()).insert_into_bucket(
+                peer,
+                networks.to_owned(),
+                increment_bucket,
+            );
         }
     }
 
--- a/concordium-node/src/connection/mod.rs
+++ b/concordium-node/src/connection/mod.rs
@@ -586,15 +586,11 @@ impl Connection {
     pub fn populate_remote_end_networks(&mut self, peer: RemotePeer, networks: &Networks) {
         self.remote_end_networks.extend(networks.iter());
 
-        let increment_bucket = |a: u64| {
-            self.handler.stats.recent_peers.with_label_values(&[a.to_string().as_str()]).inc();
-        };
-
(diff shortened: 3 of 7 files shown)


vikt0r0:
I labelled things by buckets, since I saw there was an issue for making use of them down the road. If this is undesired let me know, and I will change it to an atomic integer.

concordium-node/src/stats_export_service.rs
```diff
@@ -467,6 +470,16 @@ impl StatsExportService {
         let last_throughput_measurement_received_bytes = AtomicU64::new(0);
         let avg_bps_in = AtomicU64::new(0);
         let avg_bps_out = AtomicU64::new(0);
+        let recent_peers = IntGaugeVec::new(
+            Opts::new(
+                "recent_peers",
```
abizjak, line 475: I think the name, if it was to follow other metrics we have, would be more like `peer_bucket_size`.

docs/prometheus-exporter.md
```diff
@@ -232,3 +232,7 @@ Finalized blocks received as part of catchup are also counted, when the block wa
 Indicator for unsupported pending protocol updates where a non-zero value indicates the effective time (Unix time in milliseconds) of a pending unsupported protocol update.
 
 This metric is intended for setting up alerts to catch outdated nodes.
+
+### `recent_peers`
```
abizjak, line 236: This should be renamed as well.

concordium-node/src/p2p/maintenance.rs
```diff
@@ -739,8 +739,11 @@ pub fn spawn(
                 && Instant::now().duration_since(last_buckets_cleaned)
                     >= Duration::from_millis(node.config.bucket_cleanup_interval)
             {
+                let decrement_bucket = |a: u64| {
+                    node.stats.recent_peers.with_label_values(&[a.to_string().as_str()]).dec();
+                };
                 write_or_die!(node.buckets())
-                    .clean_buckets(node.config.timeout_bucket_entry_period);
+                    .clean_buckets(node.config.timeout_bucket_entry_period, decrement_bucket);
```
abizjak, line 746: Just pass in the gauge instead of this closure business. I don't see a reason for making things complex.

concordium-node/src/network/buckets.rs
```diff
@@ -142,8 +154,8 @@ mod tests {
         };
 
         // and check that only one is inserted
-        buckets.insert_into_bucket(p2p_peer, Default::default());
-        buckets.insert_into_bucket(p2p_duplicate_peer, Default::default());
+        buckets.insert_into_bucket(p2p_peer, Default::default(), |_| {});
+        buckets.insert_into_bucket(p2p_duplicate_peer, Default::default(), |_| {});
```
abizjak, line 158: You can pass dummy gauges in here.

concordium-node/src/network/buckets.rs
```diff
@@ -105,9 +112,14 @@ impl Buckets {
...
 
     /// Removes the bucket nodes older than then specified amount of time.
-    pub fn clean_buckets(&mut self, timeout_bucket_entry_period: u64) {
+    pub fn clean_buckets(&mut self, timeout_bucket_entry_period: u64, on_removal: impl Fn(u64)) {
         let clean_before = get_current_stamp() - timeout_bucket_entry_period;
-        self.buckets[0].retain(|entry| entry.last_seen >= clean_before);
+        self.buckets[0].retain(|entry| {
+            if entry.last_seen < clean_before {
+                on_removal(0);
+            }
```
abizjak, line 120: This is needlessly complex. You can just

```
bucket.set(self.buckets[0].len())
```
at the end.
It's clearer code.

concordium-node/src/network/buckets.rs
```diff
@@ -48,14 +48,21 @@ impl Default for Buckets {
...
         let bucket = &mut self.buckets[0];
 
-        bucket.insert(Node {
+        if bucket.insert(Node {
             peer,
             networks,
             last_seen: get_current_stamp(),
-        });
+        }) {
+            on_insert_success(0);
```
abizjak, line 64: Just pass in the gauge here and increment it.

concordium-node/src/connection/mod.rs
```diff
@@ -586,8 +586,16 @@ impl Connection {
     pub fn populate_remote_end_networks(&mut self, peer: RemotePeer, networks: &Networks) {
         self.remote_end_networks.extend(networks.iter());
 
+        let increment_bucket = |a: u64| {
+            self.handler.stats.recent_peers.with_label_values(&[a.to_string().as_str()]).inc();
+        };
```
abizjak, line 591: Don't do this complex closure business.

concordium-node/src/bin/bootstrapper.rs
```diff
@@ -44,12 +46,22 @@ fn main() -> anyhow::Result<()> {
...
-        stats_export_service,
+        stats_export_service.clone(),
         regenesis_arc,
     )
     .context("Failed to create the network node.")?;
 
     start_push_gateway(&conf.prometheus, &node.stats, node.id());
+
+    if let Some(plp) = conf.prometheus.prometheus_listen_port {
+        let (sender, _) = tokio::sync::broadcast::channel(1);
```
abizjak, line 57: DOcument here that we don't care about graceful shutdown so we ignore the receiver.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 
 ## Unreleased changes
+- Fix a bug where the prometheus interface was not exposed in the
+  bootstrapper when `CONCORDIUM_NODE_PROMETHEUS_LISTEN_PORT` was set.
```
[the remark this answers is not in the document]
vikt0r0: I see. But just to comment on that, I think either it was a bug or the the following help information is misleading, as it did not listen on a port despite being specified:
```
        --prometheus-listen-port <prometheus-listen-port>
            Port for prometheus to listen on. If set the prometheus server will start listening on the given port [env:
            CONCORDIUM_NODE_PROMETHEUS_LISTEN_PORT=9001]
```

concordium-node/src/p2p/maintenance.rs
vikt0r0: Yes!

concordium-node/src/network/buckets.rs
```diff
@@ -142,8 +158,17 @@ mod tests {
...
+        if let Ok(bucket_size_gauge_dummy) = dummy_gauge {
+            buckets.insert_into_bucket(p2p_peer, Default::default(), &bucket_size_gauge_dummy);
+            buckets.insert_into_bucket(
+                p2p_duplicate_peer,
+                Default::default(),
+                &bucket_size_gauge_dummy,
+            );
+            assert_eq!(buckets.buckets.len(), 1);
+        } else {
+            panic!("Unable to create dummy gauge.")
```
vikt0r0, line 171: Is this appropriate?

limemloh: Yes, it is fine to panic in a test like this.
You could avoid the if statement by using `.expect("...")` when creating the dummy gauge

concordium-node/src/network/buckets.rs
```diff
@@ -105,16 +114,23 @@ impl Buckets {
...
-    pub fn clean_buckets(&mut self, timeout_bucket_entry_period: u64) {
+    pub fn clean_buckets(
+        &mut self,
+        timeout_bucket_entry_period: u64,
+        bucket_size_gauge: &IntGaugeVec,
+    ) {
         let clean_before = get_current_stamp() - timeout_bucket_entry_period;
         self.buckets[0].retain(|entry| entry.last_seen >= clean_before);
+        let new_bucket_size = self.buckets[0].len();
+        bucket_size_gauge.with_label_values(&["0"]).set(new_bucket_size.try_into().unwrap());
```
limemloh, line 125: I think in this case it would be better to cast the size, since we don't want to panic just because of the metrics.
I know it is a bit nitpicking because this can't come into the state where it would panic, but still think it is important to consider and save the reader from having to do the reasoning.
So, something like this:
```suggestion
        bucket_size_gauge.with_label_values(&["0"]).set(new_bucket_size as i64);
```

limemloh (approved):
LGTM well done

concordium-node/src/network/buckets.rs
vikt0r0: Good suggestion, I think it is much more idiomatic (and clearer) to do it like that.

concordium-node/src/network/buckets.rs
```diff
@@ -105,16 +113,23 @@ impl Buckets {
...
     /// Removes the bucket nodes older than then specified amount of time.
-    pub fn clean_buckets(&mut self, timeout_bucket_entry_period: u64) {
+    pub fn clean_buckets(
+        &mut self,
+        timeout_bucket_entry_period: u64,
+        bucket_size_gauge: &IntGaugeVec,
+    ) {
         let clean_before = get_current_stamp() - timeout_bucket_entry_period;
         self.buckets[0].retain(|entry| entry.last_seen >= clean_before);
+        let new_bucket_size = self.buckets[0].len();
```
abizjak, line 123:
```suggestion
        let bucket = &mut self.buckets[0];
        bucket.retain(|entry| entry.last_seen >= clean_before);
        let new_bucket_size = bucket.len();
```

abizjak (approved):

concordium-node/src/stats_export_service.rs
```diff
@@ -467,6 +470,16 @@ impl StatsExportService {
         let last_throughput_measurement_received_bytes = AtomicU64::new(0);
         let avg_bps_in = AtomicU64::new(0);
         let avg_bps_out = AtomicU64::new(0);
+        let peer_bucket_size = IntGaugeVec::new(
+            Opts::new(
+                "peer_bucket_size",
+                "The number of peers that recently connected to the bootstrapper labelled by the \
+                 number of the bucket in which they are contained",
```
abizjak, line 477: I just realized that this help text is not the best. It talks about bootstrappers but actually the buckets are there for normal nodes as well. Maybe just replace bootstrapper with node

docs/prometheus-exporter.md
```diff
@@ -232,3 +232,7 @@ Finalized blocks received as part of catchup are also counted, when the block wa
 Indicator for unsupported pending protocol updates where a non-zero value indicates the effective time (Unix time in milliseconds) of a pending unsupported protocol update.
 
 This metric is intended for setting up alerts to catch outdated nodes.
+
+### `peer_bucket_size`
+
+The number of recently connected peers used to generate the peer list included in handshake responses of the bootstrapper. Labelled by the number of the bucket in which the peer is maintained (`bucket=<number>`).
```
abizjak, line 238: Similarly this text is not the best (similar issue with bootstrapper)

concordium-node/src/stats_export_service.rs
vikt0r0: Good point!

PR #821, opened 2023-04-13
Fix leftover renaming of metric in changelog

limemloh:
## Purpose

Forgot to update the metric name in changelog when renaming `grpc_request_duration_seconds` to `grpc_request_response_time_seconds`


abizjak (approved):
Looks good. @dg-concordium can you please make the change in the release PR in documentation as well. Thanks.

Issue #820, opened 2023-04-13
Node doesn't start on Mac


mh-concordium:
Try to install and start the node for Testnet results with error:
`launchctl error 6
6: Device not configured`

Staring it from the launchpad:
![image.png](https://images.zenhubusercontent.com/6284893f971a3daf4db6e569/855a6f3d-e81d-42c1-af9a-6f88dc575165)

**Version:** 5.3.2

mh-concordium:
Seems to be working fine now with 5.3.2-1. I was able to install it and it is catching up on Testnet.

Bargsteen:
Great 😊

PR #822, opened 2023-04-14
Fix mac build script

Bargsteen:
## Purpose

Fix the mac build script, so that the dylibs are included in the build.
Closes https://github.com/Concordium/concordium-node/issues/820

This issue was caused by a few things:
- We separated out the node collector in a previous PR.
- The macdylibbundler tool v1.0.0 did not quote all paths, which was fine for our previous node setup, but it became an issue with the new node collector (because some of our folder names have spaces in them).
- Version 1.0.5 of macdylibbundler fixes the issue with quoting paths (https://github.com/auriamg/macdylibbundler/commits/master).
- The last issue was that the second invocation of `collectDylibsFor` would overwrite the `libs` folder. I am not sure why this wasn't an issue before. Perhaps the tool doesn't overwrite the dir if the binary doesn't have any dependencies that need to be moved. But this was fixed by changing `--overwrite-dir` to `--create-dir`, which just ensures that the dir is created instead of overwriting it.

## Changes

- Update to macdylibbundler 1.0.5
- Change `--overwrite-dir` to `--create-dir`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/distribution/macOS-package/build.sh
+++ b/scripts/distribution/macOS-package/build.sh
@@ -248,6 +248,10 @@ function collectDylibs() {
         local fileToFix=${1:?"Missing file to fix with dylibbundler"};
         cd "$payloadDir/Library/Concordium Node"
         # Paths to search for dylibs are added with the '-s' flag.
+        # We use `--create-dir` to ensure that the `./libs` folder exists.
+        # But we should not use `--overwrite-dir` even though it implies `--create-dir`
+        # because it will delete the libs folder if it already exists, which would delete 
+        # the results of previous calls to `collectDylibsFor`.
         "$macdylibbundlerDir/dylibbundler" --fix-file "$fileToFix" --bundle-deps --dest-dir "./libs" --install-path "@executable_path/libs/" --create-dir \
             -s "$concordiumDylibDir" \
             -s "$stackSnapshotDir" \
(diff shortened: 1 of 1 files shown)


scripts/distribution/macOS-package/build.sh
```diff
@@ -248,7 +248,7 @@ function collectDylibs() {
         local fileToFix=${1:?"Missing file to fix with dylibbundler"};
         cd "$payloadDir/Library/Concordium Node"
         # Paths to search for dylibs are added with the '-s' flag.
-        "$macdylibbundlerDir/dylibbundler" --fix-file "$fileToFix" --bundle-deps --dest-dir "./libs" --install-path "@executable_path/libs/" --overwrite-dir \
+        "$macdylibbundlerDir/dylibbundler" --fix-file "$fileToFix" --bundle-deps --dest-dir "./libs" --install-path "@executable_path/libs/" --create-dir \
```
abizjak, line 251: It is a bit unclear why this fixes it. It would be good to document why we cannot delete the contents of the directory.

As I read it --overwrite-dir implies --create-dir, so it is the deleting part that was causing problems.

EmilLa1 (approved):

abizjak (approved):

PR #832, opened 2023-05-03
New consensus block execution

td202:
## Purpose

This implements transaction handling in block baking and execution for the new consensus.

## Changes

- Introduce `filterTables` as a (pure) replacement for `maintainTransactions` that can be used across consensus versions. The implementation should achieve the same effect, but does not filter out oversized unprocessed transactions as `maintainTransactions` does. Preliminary testing indicates `filterTables` is more efficient, especially when there are large numbers of failed or unprocessed transactions.

- Replace `maintainTransactions` with `filterTransactionTables`, a new `TreeStateMonad` operation that is implemented by `filterTables`.

- Refactor `TreeStateMonad` operations `getAccountNonFinalized`, `getCredential` and `getNonFinalizedChainUpdates` extracting pure `lookupAccountTransactions`, `lookupCredentail` and `lookupChainUpdates`.

- Remove `checkedAddPendingTransaction`, `checkedAddPendingUpdate` and `purgeTransaction`, which are unused after removing `maintainTransactions`.

- Factor transaction grouping out of `constructBlock` into `groupPendingTransactions` and `getGroupedPendingTransactions`.

- Implement scheduler for new consensus. A number of operations from `TreeStateEnvironment` are reused, in particular `doBlockReward` and `distributeRewards`.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -1,45 +1,200 @@
...
+    }
+
+-- |The result of executing the prologue.
+data PrologueResult m = PrologueResult
+    { -- |The block state after prologue execution.
+      prologueBlockState :: UpdatableBlockState m,
+      -- |If the block should pay out for a payday, these parameters determine the pay out.
+      -- Otherwise, they are 'Nothing'.
+      prologuePaydayParameters :: Maybe PaydayParameters
+    }
```
abizjak, line 102: And in all of these.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -1,45 +1,200 @@
...
--- Note: this does not update the next epoch bakers or capital distribution, which should be done
--- subsequently where necessary. This also does not update the seed state to account for the block
--- nonce, which should also be done subsequently.
+--  * Of the new epoch is the epoch before the next payday, take a snapshot of bakers and
+--    delegators, allowing for cooldowns that are set to elapse at by the trigger block time for
+--    this epoch.
+--
+-- Note: If the baker or delegator cooldown period is ever less than the duration of an epoch, then
+-- it would be possible to have a baker not in cooldown when the baker snapshot is taken, but be
+-- removed when the cooldowns are processed at the payday. This bad, because the baker/delegator
```
abizjak, line 129:
```suggestion
-- removed when the cooldowns are processed at the payday. This is bad, because the baker/delegator
```

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -1,45 +1,200 @@
...
+                -- rewards, because we will overwrite them with the parameters for the new payday.
+                paydayCapitalDistribution <- bsoGetCurrentCapitalDistribution theState
+                paydayBakers <- bsoGetCurrentEpochFullBakersEx theState
+                paydayPoolRewards <- bsoGetBakerPoolRewardDetails theState
+                paydayMintRate <- bsoGetPaydayMintRate theState
+                let paydayParams = PaydayParameters{..}
+                -- Rotate the capital distribution and bakers so that the current values are
+                -- replaced by the next values.
+                theState <- bsoRotateCurrentCapitalDistribution theState
+                theState <- bsoRotateCurrentEpochBakers theState
```
abizjak, line 161: I think it is a really bad idea to reuse variable names in Haskell. We don't do this anywhere else so please don't do it here either. 

It's so easy to get wrong.

concordium-consensus/src/Concordium/KonsensusV1/Transactions.hs
```diff
@@ -186,50 +201,57 @@ processBlockItems ::
...
                 -- If that is the case then we add it to the transaction table and pending transactions.
                 -- If it is NotOk then we stop verifying the transactions as the block can never be valid now.
-                !verRes <- runTransactionVerifierT (TVer.verify theTime bi) =<< getCtx
+                !verRes <- verifyBlockItem theTime bi verificationContext
                 case verRes of
                     -- The transaction was deemed non verifiable i.e., it can never be
-                    -- valid. We short circuit the recursion here and return 'False'.
-                    (TVer.NotOk _) -> return False
+                    -- valid. We short circuit the recursion here and return 'Nothing'.
+                    (TVer.NotOk _) -> empty
```
abizjak, line 245: Can we please just return Nothing here. This `empty` is needlessly general here.

concordium-consensus/src/Concordium/KonsensusV1/Transactions.hs
```diff
@@ -186,50 +201,57 @@ processBlockItems ::
...
     theTime = bbTimestamp bb
-    -- Process the list of transactions recursively.
-    process :: [BlockItem] -> m Bool
-    process [] = return True
-    process (bi : bis) = do
+    -- Process a transaction
+    process ::
+        Context (PBS.HashedPersistentBlockState pv) ->
+        BlockItem ->
+        MaybeT m (BlockItem, TVer.VerificationResult)
```
abizjak, line 225: Generally MaybeT is not ideal in a tight loop, which this partially is. The bind is rather expensive due to constant checking of Just/Nothign.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -365,6 +366,18 @@ checkQuorumCertificate qsmGenesis sigThreshold FinalizationCommittee{..} QuorumC
             Nothing -> False
             Just FinalizerInfo{..} -> check (finalizerWeight + accumWeight) (finalizerBlsKey : keys) is
 
+-- |Determine the 'BakerId's of the finalizers that signed a 'QuorumCertificate'.
+-- This assumes that all of the signatories are valid indexes into the finalization committee
+-- (which is checked by 'checkQuorumCertificate'). Any invalid finalization indexes will be
+-- omitted.
```
EmilLa1, line 372: Maybe we should document the order of the list returned, I suppose their ordered by ascending baker id.

And a single fold is sufficient since the finalizers are in ascending order based on their respective baker id.

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
```diff
@@ -188,3 +190,112 @@ purgeTables lastFinCommitPoint oldestArrivalTime currentTime TransactionTable{..
...
+    commitAdded' _ [] tt = tt
+    commitAdded' i (bi : bis) tt =
+        commitAdded' (i + 1) bis $!
+            tt & ttHashMap . ix (getHash bi) . _2 %~ addResult commitBlock commitAt i
+    -- Update the pending transaction table
+    forwarded = forwardPTT addedTransactions ptt0
+    -- Determine if a block item is committed to a potentially-live block.
+    isCommitted :: TransactionTable -> WithMetadata a -> Bool
+    isCommitted tt t = case tt ^? ttHashMap . ix (wmdHash t) . _2 . tsCommitPoint of
+        Just transactionCP -> commitPoint lastFinCommit < commitPoint transactionCP
```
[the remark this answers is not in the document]
EmilLa1: It's with the perspective on whether it's comitted to a **live** block or not. So the commit point must be strictly larger than the last finalized commit point?

But yes you are right, if it was `<=` it would still hit the `Nothing` case as you are correct regarding whether the finalized transactions are here or not, so the result of the function will be the same.

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
```diff
@@ -188,3 +190,112 @@ purgeTables lastFinCommitPoint oldestArrivalTime currentTime TransactionTable{..
...
+        Nothing -> False
+    -- Purge a transaction from the transaction table if it is not committed.
+    purgeTransaction tt ((t, _), _)
```
EmilLa1, line 236: should the arguments be forced here?

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
```diff
@@ -188,3 +190,112 @@ purgeTables lastFinCommitPoint oldestArrivalTime currentTime TransactionTable{..
...
+    purgeCredentials tables = foldl' purgeCred tables ftFailedCredentials
+      where
+        purgeCred ts@(!tt, !ptt) ((cred, _), _)
+            | isCommitted tt cred = ts
+            | otherwise =
+                ( tt & ttHashMap . at' (wmdHash cred) .~ Nothing,
+                  ptt & pttDeployCredential %~ HS.delete (getHash cred)
+                )
+    -- Purge a chain update from the transaction table where it is not already committed to a block.
+    purgeUpdate tt ((upd, _), _)
```
EmilLa1, line 276: Same comment as above here.

EmilLa1 (commented):
A few comments for now.

EmilLa1 (approved):

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
td202: I will make the change, but I disagree. Not reusing names in this case makes it much easier to get things wrong.

eb-concordium (approved):
Looks good!

abizjak (approved):

PR #834, opened 2023-05-04
Update readme

lassemand:
## Purpose

Currently the documentation does not correlate with the product, if one is to plainly follow the guide. Meaning it is not explicitly stated that certain variables has to be set in order to execute `cargo run --`.

## Changes

Add documentation about required variables which has to be set in order to start the node

### DIFF
--- a/concordium-node/README.md
+++ b/concordium-node/README.md
@@ -97,7 +97,7 @@ Environment variables only apply to the default build. This links with shared Ha
   Once the node is built it can be run as
 
   ```console
-  cargo run -- --config-dir <CONFIG_DIR> --data-dir <DATA_DIR>
+  cargo run -- --config-dir <CONCORDIUM_NODE_CONFIG_DIR> --data-dir <CONCORDIUM_NODE_DATA_DIR>
   ```
 
   or
@@ -105,7 +105,6 @@ Environment variables only apply to the default build. This links with shared Ha
   ```console
   cargo run --release -- --config-dir <CONCORDIUM_NODE_CONFIG_DIR> --data-dir <CONCORDIUM_NODE_DATA_DIR>
   ```
-
 
   to be run in release mode for improved performance. Note that the
   [concordium-consensus](../concordium-consensus/) dependency is the same regardless of how the
@@ -115,7 +114,6 @@ Documentation about what <CONCORDIUM_NODE_CONFIG_DIR> and <CONCORDIUM_NODE_DATA_
 
 - The node built with Haskell library auto-discovery is not suitable for distribution to other
   machines. It is a dynamically linked binary with a large number of shared library dependencies.
-
 
 ### MacOS Specific
 
--- a/concordium-node/README.md
+++ b/concordium-node/README.md
@@ -110,7 +110,7 @@ Environment variables only apply to the default build. This links with shared Ha
   [concordium-consensus](../concordium-consensus/) dependency is the same regardless of how the
   node itself is built, the `--release` only applies to the optimization of the rust node xcomponents.
   
-Documentation about what <CONCORDIUM_NODE_CONFIG_DIR> and <CONCORDIUM_NODE_DATA_DIR> is seen at https://github.com/Concordium/concordium-node/blob/main/scripts/distribution/ubuntu-packages/README.md.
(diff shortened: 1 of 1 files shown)


concordium-node/README.md
```diff
@@ -97,18 +97,20 @@ Environment variables only apply to the default build. This links with shared Ha
...
 
   ```console
-  cargo run --release --
+  cargo run --release -- --config-dir <CONCORDIUM_NODE_CONFIG_DIR> --data-dir <CONCORDIUM_NODE_DATA_DIR>
   ```
 
   to be run in release mode for improved performance. Note that the
   [concordium-consensus](../concordium-consensus/) dependency is the same regardless of how the
-  node itself is built, the `--release` only applies to the optimization of the Rust node xcomponents.
+  node itself is built, the `--release` only applies to the optimization of the rust node xcomponents.
```
abizjak, line 111: Well you could fix xcomponents as well.

concordium-node/README.md
```diff
@@ -125,7 +127,7 @@ extra-include-dirs:
 - /opt/homebrew/include/
 ```
 
-### M1 MacOS Specific
+### Apple silicon specific
```
abizjak, line 130: This is less clear.

abizjak (changes requested):

concordium-node/README.md
```diff
@@ -97,18 +97,20 @@ Environment variables only apply to the default build. This links with shared Ha
...
-  cargo run --release --
+  cargo run --release -- --config-dir <CONCORDIUM_NODE_CONFIG_DIR> --data-dir <CONCORDIUM_NODE_DATA_DIR>
   ```
 
   to be run in release mode for improved performance. Note that the
   [concordium-consensus](../concordium-consensus/) dependency is the same regardless of how the
-  node itself is built, the `--release` only applies to the optimization of the Rust node xcomponents.
+  node itself is built, the `--release` only applies to the optimization of the rust node xcomponents.
+  
+Documentation about what <CONCORDIUM_NODE_CONFIG_DIR> and <CONCORDIUM_NODE_DATA_DIR> is seen at https://github.com/Concordium/concordium-node/blob/main/scripts/distribution/ubuntu-packages/README.md.
```
abizjak, line 113: This is not the place to refer to. This is https://github.com/Concordium/concordium-node/blob/main/VARIABLES.md#common

concordium-node/README.md
lassemand: It is however more correct. Would you rather have something like M1/M2 MacOS Specific?

concordium-node/README.md
lassemand: Sorry I do not understand that comment

concordium-node/README.md
lassemand: I would much rather keep it as documented in this PR. The generalised term for the individual processors are indeed Apple silicon, and it is just annoying to repeatedly add individual product names.

Right now my chip (M2) does not go in under this header however it should.

Furthermore the term M1 MacOS is simply invalid and does not exist. My MacOS is Ventura while my chip is an M2.

concordium-node/README.md
abizjak: `xcomponents` should be `components`. It's in the same line.

concordium-node/README.md
abizjak: The instructions are related to the operating system + CPU together. If I ran linux on M1 they would not apply.

concordium-node/README.md
lassemand, line 111:
```suggestion
  node itself is built, the `--release` only applies to the optimization of the rust node components.
```

concordium-node/README.md
lassemand: Okay based on how the statement is specified it sounds a lot like the MacOS is M1. 

_Apple silicon MacOS specific_ or _M1/M2 MacOS specific_? or would you rather we enter each individual processor in the title?

concordium-node/README.md
```diff
@@ -125,7 +127,7 @@ extra-include-dirs:
 - /opt/homebrew/include/
 ```
 
-### M1 MacOS Specific
+### Apple silicon + MacOS specific
```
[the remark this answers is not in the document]
lassemand:
```suggestion
### M1 and M2 MacOS specific
```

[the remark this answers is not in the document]
lassemand:
```suggestion
### M1 and M2 MacOS specific
```

abizjak (approved):

Issue #837, opened 2023-05-15
Support querying using block height in gRPC V2 API


limemloh:
Allow block related queries to use block height as a block identifier, useful when iterating a range of blocks.

PR #838, opened 2023-05-15
Support block height as block identifier in gRPC v2 block queries

limemloh:
## Purpose

Closes https://github.com/Concordium/concordium-node/issues/837.
Should be merged with https://github.com/Concordium/concordium-grpc-api/pull/36 and https://github.com/Concordium/concordium-base/pull/374.

## Changes

- Support using block height as block identifiers in gRPC v2 API.
- Extend gRPC v2 API call `GetBlockInfo` with the protocol version of the block.

### DIFF
--- a/concordium-consensus/src/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src/Concordium/External/GRPC2.hs
@@ -48,6 +48,7 @@ import Concordium.GlobalState.Parameters (CryptographicParameters)
 import Concordium.ID.Parameters (withGlobalContext)
 import qualified Concordium.Types.InvokeContract as InvokeContract
 import qualified Concordium.Wasm as Wasm
+import Data.Either (fromRight)
 import Data.Text (Text)
 import qualified Data.Text.Encoding as Text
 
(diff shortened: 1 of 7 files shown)


concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -396,10 +418,10 @@ getNextAccountNonce accountAddress = liftSkovQueryLatest $ do
 -- ** Block indexed
 
 -- |Get the basic info about a particular block.
-getBlockInfo :: BlockHashInput -> MVR finconf (BlockHash, Maybe BlockInfo)
+getBlockInfo :: BlockHashInput -> MVR finconf (Maybe BlockHash, Maybe BlockInfo)
```
abizjak, line 421: With all these maybes, can we document that 
(1) in case the second response is Just the first one always is
(2) the first response is Just if and only if the block hash input could be uniquely resolved.

Ideally all of these would have a more precise response type since really only 3 options are possible
Nothing, Nothing
Just, Nothing
Just Just

but that made for more changes so I'm not sure it's worth doing that. But documentation should be added.

concordium-consensus/src/Concordium/External/GRPC2.hs
```diff
@@ -170,8 +182,9 @@ getAccountInfoV2 cptr blockType blockHashPtr accIdType accIdBytesPtr outHash out
     res <- runMVR (Q.getAccountInfo bhi ai) mvr
     returnMessageWithBlock (copier outVec) outHash res
 
-copyHashTo :: Ptr Word8 -> BlockHash -> IO ()
-copyHashTo dest (BlockHash (Hash h)) = FBS.withPtrReadOnly h $ \p -> copyBytes dest p 32
+copyHashTo :: Ptr Word8 -> Maybe BlockHash -> IO ()
```
Bargsteen, line 185: Docs

concordium-consensus/src/Concordium/Queries.hs
Bargsteen: I think it is worth doing, as I find the multiple Maybe setup harder to read and it affects the whole API here.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -2439,14 +2440,14 @@ impl ConsensusContainer {
         let sender = Box::new(sender);
         let consensus = self.consensus.load(Ordering::SeqCst);
         let mut buf = [0u8; 32];
-        let (block_id_type, block_hash) =
-            crate::grpc2::types::block_hash_input_to_ffi(block_hash).require()?;
+        let bhi = crate::grpc2::types::block_hash_input_to_ffi(block_hash).require()?;
+        let (block_id_type, block_hash) = bhi.to_ptr();
```
Bargsteen, line 2444: Consider making this into a helper function as it is repeated a lot.

Bargsteen (commented):
Nice work! 
A few minor issues, where the main issue is the readability of the double `Maybe`s. I would prefer a more descriptive data type.

concordium-consensus/src/Concordium/Queries.hs
limemloh: I have made an attempt at improving this.

concordium-node/src/consensus_ffi/ffi.rs
limemloh: I don't think it will become easier to read if I factor this out, since the function calling this needs ownership of `bhi` to ensure it is kept alive while consensus processes the query.

abizjak: This is repeated a lot, but it is also necessary due to lifetimes. The original `bhi` must be kept alive long enough that is why we cannot simplify this.

abizjak (approved):

concordium-consensus/src/Concordium/Queries.hs
abizjak: I think it looks good.

PR #840, opened 2023-05-16
Identify nodes in buckets that have the same IP + port.

### LINKED ISSUE #839: Peers are not correctly identified in buckets
**Bug Description**

**Task description**

When peers are maintained in buckets they are identified via their "local id". This is a fresh ID that is assigned on each new connection.

This is incorrect since even if the same node attempts connection multiple times, and becomes a peer multiple times as a result, then it will be assigned different local node ids. As a result, the same IP + port combination is in the buckets multiple times.

This can lead to more problems with bootstrapping since peers are not served efficiently.

**Versions**
- Software Version: All node versions since version 1, until 5.3.*.

### LINKED ISSUE #841: When node is running as a "Node" it adds peers to buckets which it never cleans
**Bug Description**

When a node is running as a normal node (not as a bootstrapper) it adds peers it connects with to the buckets (used by the bootstrappers), but it never uses them or removes them.

This can be observed indirectly by looking at the prometheus metrics.

abizjak:
## Purpose

Fixes #839 

Fixes #841 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased changes
 
+- Do not use peer buckets when running as a normal node (not a bootstrapper).
 - Enable CORS support in grpc-web. This only applies when grpc-web is enabled.
 
 ## 5.3.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased changes
 
-- Do not use peer buckets when running as a normal node (not a bootstrapper).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased changes
 
+- Do not use peer buckets when running as a normal node (not a bootstrapper).
```
EmilLa1, line 5:
```suggestion
- Do not use peer buckets when running as a normal node (not as a bootstrapper).
```

EmilLa1 (approved):

PR #865, opened 2023-05-31
New consensus timeout bugfixes

EmilLa1:
## Purpose

Address a few outstanding comments with respect to mainly timeout message handling. 

https://github.com/Concordium/concordium-node/pull/751#discussion_r1210363933
https://github.com/Concordium/concordium-node/pull/751#discussion_r1210380258
https://github.com/Concordium/concordium-node/pull/751#discussion_r1211117324
https://github.com/Concordium/concordium-node/pull/751#discussion_r1211131183

## Changes

Corrected the code according to the above mentioned comments. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
@@ -155,7 +155,7 @@ receiveTimeoutMessage tm@TimeoutMessage{tmBody = TimeoutMessageBody{..}} skovDat
                                 if qcEpoch tmQuorumCertificate == certEpoch
                                     then received finInfo finalizationCommittee Absent
                                     else return $ Rejected BadQCEpoch
-                            -- If we have not already checked a QC for the round of the quorum certificate assoicated
+                            -- If we have not already checked a QC for the round of the quorum certificate associated
                             -- with the timeout message.
                             _ -> do
                                 getRecentBlockStatus (qcBlock tmQuorumCertificate) skovData >>= \case
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
@@ -288,8 +288,7 @@ executeTimeoutMessage (PartiallyVerifiedTimeoutMessage{..})
     -- - that @tmEpoch >= qcEpoch tmQuorumCertificate@, so we do not need to check that here.
     | Absent <- pvtmBlock = do
         -- In this case, we have already checked a valid QC for the round and epoch of the timeout
-        -- message, but the message is on either a 'BlockPending' or 'BlockUnknown' block.
-        -- We just need to process the timeout.
+        -- message. We just need to process the timeout.
         processTimeout pvtmTimeoutMessage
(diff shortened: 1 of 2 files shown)


chportma (approved):
It looks good! There is just one comment in the code which is now inaccurate. Instead of
        -- In this case, we have already checked a valid QC for the round and epoch of the timeout
        -- message, but the message is on either a 'BlockPending' or 'BlockUnknown' block.
        -- We just need to process the timeout.

We could write
        -- In this case, we have already checked a valid QC for the round and epoch of the timeout
        -- message. We just need to process the timeout.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
```diff
@@ -143,41 +143,38 @@ receiveTimeoutMessage tm@TimeoutMessage{tmBody = TimeoutMessageBody{..}} skovDat
...
+                        -- our highest certified block.
+                        let existingQC = skovData ^. roundExistingQuorumCertificate (qcRound tmQuorumCertificate)
+                            oldOrCurrentHighestCertifiedQCRound = qcRound tmQuorumCertificate <= skovData ^. roundStatus . rsHighestCertifiedBlock . to cbRound
+                        case (existingQC, oldOrCurrentHighestCertifiedQCRound) of
+                            (Just (QuorumCertificateCheckedWitness certEpoch), True) -> do
+                                -- The epochs must match otherwise we reject.
+                                if qcEpoch tmQuorumCertificate == certEpoch
+                                    then received finInfo finalizationCommittee Absent
+                                    else return $ Rejected BadQCEpoch
+                            -- If we have not already checked a QC for the round of the quorum certificate assoicated
```
abizjak, line 158:
```suggestion
                            -- If we have not already checked a QC for the round of the quorum certificate associated
```

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
```diff
@@ -143,41 +143,38 @@ receiveTimeoutMessage tm@TimeoutMessage{tmBody = TimeoutMessageBody{..}} skovDat
...
+                            -- with the timeout message.
+                            _ -> do
```
abizjak, line 160: Does it ever make sense to log the distinction between the 3 different scenarios here?

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
```diff
@@ -319,7 +315,7 @@ executeTimeoutMessage (PartiallyVerifiedTimeoutMessage{..})
                 then -- Advance the round with the new certified block.
                     advanceRoundWithQuorum newCertifiedBlock
                 else -- Otherwise we just update the highest certified block.
```
abizjak, line 317: I suppose the comment isn't exactly accurate anymore.

abizjak (approved):

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Timeout.hs
EmilLa1: Fixed.

Issue #792, opened 2023-03-17
New Consensus: Consolidate logging


EmilLa1:
Figure out what is appropriate for logging and at what level it should be in the new consensus protocol.

In particular the following modules should be checked whether the logging is appropriate. 

- [x] Timeout module (added WARN logging to flaggable offences)
- [x] Quorum module (added WARN logging to flaggable offences)
- [x] Blocks module (Changed all log stmts for flaggable offences to be at WARN level.)

abizjak:
@EmilLa1 All the tasks in this issue have been checked off. Is this to be closed?

EmilLa1:
> @EmilLa1 All the tasks in this issue have been checked off. Is this to be closed?

The checkboxes refer to the progress of the corresponding PR https://github.com/Concordium/concordium-node/pull/866

This is not merged, so in that regard we can't close this issue. 

However, as stated in the PR. It's not clear if we actually want to do anything about it right now - perhaps it should be done at a later stage when we will actually make use the flagging mechanism. 
So in that regard, I think we should close this issue and the PR now. @abizjak Agree?

PR #866, opened 2023-05-31
Flaggable offences are now logged at WARN level.

EmilLa1:
## Purpose

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
@@ -252,7 +252,7 @@ receiveBlockKnownParent parent pendingBlock = do
                     return BlockResultInvalid
         | otherwise = do
             -- The block's epoch is not valid.
-            logEvent Konsensus LLTrace $
+            logEvent Konsensus LLWarning $
                 "Block "
                     <> show pbHash
                     <> " is for an invalid epoch ("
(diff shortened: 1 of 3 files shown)


EmilLa1:
After discussion with @td202 maybe INFO is more suitable as an operator can't really act on these warnings.
@abizjak thoughts?

EmilLa1 (commented):
Don't print the round, epoch and finalizer index just using the show instance as these are stock instances. 

Instead only print out the wrapped value.

abizjak:
> After discussion with @td202 maybe INFO is more suitable as an operator can't really act on these warnings. @abizjak thoughts?

I am fine with INFO, but I am not exactly sure why this change was rushed in at this time, it was certainly not my expectation this would be done now.

It is unclear to me whether some of these situations lead to any punishment or disconnect of the peer sending the bad message. If not, then logging things at info or warn means that there is a relatively easy way to cause the node runners to run out of disk space.

Perhaps it is clear to you, but I have not gone through all the cases to understand that this change is safe.

abizjak (commented):
As discussed in the office let's wait with this until we have time to properly think it through.

EmilLa1:
Closing this for now as per https://github.com/Concordium/concordium-node/issues/792. It is not clear it is something we want to work on at the moment.

Issue #869, opened 2023-06-02
New Consensus: Out of band catch up


EmilLa1:
Implement block export and block import for the new skov.

PR #875, opened 2023-06-02
Export consensusv1 database

EmilLa1:
## Purpose

https://github.com/Concordium/concordium-node/issues/869

## Changes

- Expanded checkDatabase such that it also handles blocks for the new consensus protocol.
- Expand the exporting functionality such that it also exports a v1 low level database.
- Populate the metadata store when inititalizing consensusv1 database.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/ImportExport.hs
+++ b/concordium-consensus/src/Concordium/ImportExport.hs
@@ -1,7 +1,7 @@
-{-# LANGUAGE MonoLocalBinds #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE GADTs #-}
 
 -- |Functionality for importing and exporting the block database.
 --
@@ -39,6 +39,7 @@
 -- genesis block, and all finalization records that are not already included in blocks.
 module Concordium.ImportExport where
 
+import Control.Monad.Trans.Reader
 import Control.Monad
 import Control.Monad.Catch
 import Control.Monad.IO.Class
@@ -70,6 +71,9 @@ import Concordium.Logger
 import Concordium.Types
 import Concordium.Utils.Serialization.Put
 import Lens.Micro.Platform
+import qualified Concordium.KonsensusV1.TreeState.LowLevel.LMDB as KonsensusV1
+import qualified Concordium.KonsensusV1.TreeState.LowLevel as KonsensusV1
+import Concordium.KonsensusV1.TreeState.LowLevel (MonadTreeStateStore(lookupLastBlock))
 
 -- |State used for exporting the database.
 data DBState pv = DBState
@@ -367,6 +371,17 @@ exportDatabaseV3 dbDir outDir chunkSize = do
             return exportError
         Nothing -> return exportError
 
+exportConsensusV1Blocks ::
+  (
+    MonadLogger m,
+    KonsensusV1.MonadTreeStateStore m
+  ) =>
+  m (Bool, BlockIndex)
+exportConsensusV1Blocks = do
+    lookupLastBlock >>= \case
+        Nothing -> return (True, Empty)
+        Just sb -> undefined
+
 -- |Export database sections corresponding to blocks with genesis indices >= genIndex
 -- and of height >= startHeight.
 -- Returns a @Bool@ and a @BlockIndex@ where the former indicates whether an error occurred,
@@ -397,8 +412,15 @@ exportSections dbDir outDir chunkSize genIndex startHeight blockIndex lastWritte
         then
             openReadOnlyDatabase treeStateDir >>= \case
                 Nothing -> do
-                    putStrLn "Tree state database could not be opened"
-                    return (True, Empty)
+                    -- If we cannot open the database as a 'ConsensusV0', then
+                    -- we try open up the 'ConsensusV1' database.
+                    KonsensusV1.openReadOnlyDatabase treeStateDir >>= \case
+                        -- If the database is unrecognized we stop here.
+                        Nothing -> do
+                            putStrLn "Tree state database could not be opened"
+                            return (True, Empty)
+                        Just (KonsensusV1.VersionDatabaseHandlers (dbh :: KonsensusV1.DatabaseHandlers pv)) -> 
+                            runSilentLogger $ runReaderT (KonsensusV1.runDiskLLDBM @pv exportConsensusV1Blocks) dbh
                 Just (VersionDatabaseHandlers (dbh :: DatabaseHandlers pv ())) -> do
                     (exportError, sectionData) <-
                         liftIO $
@@ -420,17 +442,7 @@ exportSections dbDir outDir chunkSize genIndex startHeight blockIndex lastWritte
                                                     -- database so we do not accidentally export one chain on top of
                                                     -- another.
                                                     let genHash = finalizationBlockPointer genFinRec
-                                                    let exportedGenHash = case blockIndex of
-                                                            -- nothing was previously exported for the
-                                                            -- section corresponding to the current
-                                                            -- genesis index.
-                                                            Empty -> genHash
-                                                            -- blocks were previously exported for the
-                                                            -- current genesis index, so use the genesis
-                                                            -- block hash that was previously exported
-                                                            -- for that section and compare it with that
-                                                            -- of the database.
-                                                            _ :|> (gh, _) -> gh
+                                                    let exportedGenHash = exportedGenHashOr genHash
                                                     if genHash /= exportedGenHash
                                                         then do
                                                             liftIO . putStrLn $
@@ -481,6 +493,19 @@ exportSections dbDir outDir chunkSize genIndex startHeight blockIndex lastWritte
             -- the condition for terminating the export.
             putStrLn $ "The tree state database does not exist at " ++ treeStateDir
             return (False, Empty)
+  where
+    -- Return the last exported genesis hash or the provided one.
+    exportedGenHashOr bh = case blockIndex of
+        -- nothing was previously exported for the
+        -- section corresponding to the current
+        -- genesis index.
+        Empty -> bh
+        -- blocks were previously exported for the
+        -- current genesis index, so use the genesis
+        -- block hash that was previously exported
+        -- for that section and compare it with that
+        -- of the database.
+        _ :|> (gh, _) -> gh
 
 -- |Write a database section as a collection of chunks in the specified directory. The last exported
 -- chunk (i.e. the one containing the block with the greatest height in the section) also contains
--- a/concordium-consensus/src/Concordium/ImportExport.hs
+++ b/concordium-consensus/src/Concordium/ImportExport.hs
@@ -773,12 +773,16 @@ exportBlocksToChunk hdl firstHeight chunkSize lastFinalizationRecordIndex getBlo
                 liftIO $ do
                     BS.hPut hdl $ runPut $ putWord64be len
                     BS.hPut hdl serializedBlock
+                -- For 'ConsensusV0' this will yield a result (as we are looking up finalized blocks),
+                -- while for 'ConsensusV1' this will always return @Nothing@, and as such
+                -- the @lastFinRecIdx@ will remain zero throughout the export.
                 let lastFinRecIdx' = fromMaybe lastFinRecIdx finalizationRecordIndexM
                 if count < chunkSize
                     then ebtc (height + 1) (count + 1) lastFinRecIdx'
                     else return (count, lastFinRecIdx')
 
 -- |Export all finalization records with indices above `dbsLastFinIndex` to a chunk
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -1,4 +1,5 @@
-{-# LANGUAGE MonoLocalBinds #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
abizjak, line 1: We don't usually want this. Why is it necessary?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -367,6 +381,85 @@ exportDatabaseV3 dbDir outDir chunkSize = do
...
+    -- |Last written chunk in previous export
+    Maybe String ->
+    -- |Returns a @Bool@ which indicates whether anything went wrong,
+    -- i.e. it is 'True' if an error occured and otherwise 'False,
+    -- and the resulting 'BlockIndex'.
+    m (Bool, BlockIndex)
+exportConsensusV1Blocks outDir chunkSize genIndex startHeight blockIndex lastWrittenChunkM = do
+    KonsensusV1.lookupFirstBlock >>= \case
+        Nothing -> do
+            liftIO . putStrLn $ "Could not read from database."
```
abizjak, line 424: Could we not instead use the logging infrastructure here? It's a bit odd we have runSilentLogger and then here we use println.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -367,6 +381,85 @@ exportDatabaseV3 dbDir outDir chunkSize = do
...
+            return (True, Empty)
+        Just genesisBlock -> do
+            let exportedGenHash = exportedGenHashOr genHash blockIndex
+                genHash = getHash genesisBlock
+            if genHash /= exportedGenHash
+                then do
+                    liftIO . putStrLn $ "Genesis hash does not match the recently exported block index."
```
abizjak, line 431: also here.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -397,8 +490,20 @@ exportSections dbDir outDir chunkSize genIndex startHeight blockIndex lastWritte
         then
             openReadOnlyDatabase treeStateDir >>= \case
                 Nothing -> do
-                    putStrLn "Tree state database could not be opened"
-                    return (True, Empty)
+                    KonsensusV1.openReadOnlyDatabase treeStateDir >>= \case
```
abizjak, line 493: Could you comment here that you do "try and catch" approach and note the reasons a bit, so it's easier to maintain in the future?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -482,6 +596,14 @@ exportSections dbDir outDir chunkSize genIndex startHeight blockIndex lastWritte
             putStrLn $ "The tree state database does not exist at " ++ treeStateDir
             return (False, Empty)
 
+-- |Action for getting a block (and possibly a finalization index) at a particular height.
+-- Note that for consensus v1 the finalization index is simply @Nothing@ as they do not
+-- exist on this version.
+type GetBlockAt m = BlockHeight -> m (Maybe (BS.ByteString, Maybe FinalizationIndex))
```
abizjak, line 602: Could you provide a bit of context as to why we do this slightly hacky version with  an optional FinalizationIndex as opposed to a more "type safe" version. So we can maintain it better in the future.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -367,6 +381,85 @@ exportDatabaseV3 dbDir outDir chunkSize = do
...
+                                writeChunks
+                                    genIndex
+                                    (demoteProtocolVersion (protocolVersion @pv))
+                                    genHash
+                                    startHeight
+                                    (KonsensusV1.bmHeight . KonsensusV1.stbInfo $ sb)
+                                    outDir
+                                    chunkSize
+                                    lastWrittenChunkM
+                                    0
```
abizjak, line 458: So this 0 here is a dummy value, right? Can we document that please?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -454,9 +565,12 @@ exportSections dbDir outDir chunkSize genIndex startHeight blockIndex lastWritte
                                                                     outDir
                                                                     chunkSize
                                                                     lastWrittenChunkM
+                                                                    0
```
abizjak, line 568: Why is this correct? What is the meaning of "lastFinalizationIndex" here?

concordium-consensus/src/Concordium/ImportExport.hs
abizjak: Can you elaborate on when the finalization index will be present? Is it when the block is finalized in <= P5?

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -586,63 +719,64 @@ writeChunks
...
+                let len = fromIntegral $ BS.length serializedBlock
                 liftIO $ do
                     BS.hPut hdl $ runPut $ putWord64be len
                     BS.hPut hdl serializedBlock
-                forM_ (blockFields (sbBlock sb)) $ \bf ->
-                    case blockFinalizationData bf of
-                        BlockFinalizationData fr ->
-                            dbsLastFinIndex .= finalizationIndex fr
-                        NoFinalizationData -> return ()
+                let lastFinRecIdx' = fromMaybe lastFinRecIdx finalizationRecordIndexM
```
abizjak, line 755: I'm not sure I understand this. So you are looping here over finalized blocks in the database.

So if at the end you'd get a Nothing for finalization index (why does that happen?) then you'd return `0` from this function. 

I guess that in V0 consensus you'd always have a finalization index, so this will be fine.

And in consensus V1 you'd never have it, and it also does not matter since it's not used.

I understand there is a lack of time to do it more cleanly, but could we explain this a bit more at least since it's not so clear from the code itself.

abizjak: We could probably relatively easily change the API a bit so that 
```
GetBlockAt
```
would take an additional type parameter, the finalizationindex.

But I think that'll take you a bit too much time, so this is just a thought for now.

abizjak (approved):

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Yes. Documented.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Documented this.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Documented this.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Introduced a logger for all output.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Fixed it.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Elaborated a bit.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: I have added some documentation.

EmilLa1: .. And refactored a bit.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -586,67 +760,91 @@ writeChunks
...
-                    len = fromIntegral $ BS.length serializedFr
-                liftIO $ do
-                    BS.hPut hdl $ runPut $ putWord64be len
-                    BS.hPut hdl serializedFr
-                exportFinRecsFrom (count + 1) (finRecIndex + 1)
+        case getFinalizationRecordAt of
+            -- Write out the @FinalizationRecord@s above the last @FinalizationIndex@.
+            -- We terminate the loop when there is the 'GetFinalizationRecordAtV0' returns
+            -- 'Nothing'.
+            GetFinalizationRecordAtV0 f -> do
```
abizjak, line 837: Something is a bit off in this sentence.

abizjak (approved):

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -1,7 +1,9 @@
-{-# LANGUAGE MonoLocalBinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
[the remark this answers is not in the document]
EmilLa1: It isn't anymore.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -357,23 +376,110 @@ exportDatabaseV3 dbDir outDir chunkSize = do
...
+    -- |Chunk size
+    Word64 ->
+    -- |The genesis index.
+    GenesisIndex ->
+    -- |Height to start export from.
+    BlockHeight ->
+    -- |The block index of the previous export.
+    BlockIndex ->
+    -- |Last written chunk in previous export
+    Maybe String ->
```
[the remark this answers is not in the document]
EmilLa1: It was a `Maybe String` before - changed it now to be `Maybe FilePath.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: fixed.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -586,67 +762,89 @@ writeChunks
...
+    writeBlockOut :: BS.ByteString -> m ()
+    writeBlockOut serializedBlock = do
+        let len = fromIntegral $ BS.length serializedBlock
+        liftIO $ do
+            BS.hPut hdl $ runPut $ putWord64be len
+            BS.hPut hdl serializedBlock
+    continue count height finalizationRecordIndex =
+        if count < chunkSize
+            then ebtc (height + 1) (count + 1) finalizationRecordIndex
+            else return (count, finalizationRecordIndex)
```
td202, line 818: I would suggest making separate V0 and V1 versions of this function, rather than combining them together as currently done. In particular, this avoids casing on `getBlockAt` inside the loop, and we can do without finalization indexes altogether in V1 (where they don't exist as a concept). Of course, the likes of `writeBlockOut` can be factored out.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -586,67 +762,89 @@ writeChunks
...
 
 -- |Export all finalization records with indices above `dbsLastFinIndex` to a chunk
+-- Note. For 'ConsensusV1' this function will not write anything to the file.
 exportFinRecsToChunk ::
```
td202, line 822: For this, as we don't really need the consensus V1 version, I would just make it the V0 version, and at the call-site decide whether it should be called or not.

td202 (approved):
This looks fine. We will need to return to include (some) finalization entries in the consensus V1 export.

concordium-consensus/src/Concordium/ImportExport.hs
```diff
@@ -357,23 +375,110 @@ exportDatabaseV3 dbDir outDir chunkSize = do
...
+    GenesisIndex ->
+    -- |Height to start export from.
+    BlockHeight ->
+    -- |The block index of the previous export.
+    BlockIndex ->
+    -- |Last written chunk in previous export
+    Maybe FilePath ->
+    -- |Returns a @Bool@ which indicates whether anything went wrong,
+    -- i.e. it is 'True' if an error occured and otherwise 'False,
+    -- and the resulting 'BlockIndex'.
```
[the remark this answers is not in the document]
EmilLa1: Yes. Added that.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Same as above.

concordium-consensus/src/Concordium/ImportExport.hs
EmilLa1: Same as above.

PR #878, opened 2023-06-06
Consensus catchup

EmilLa1:
## Purpose

This implements catch-up for protocol version 6.

## Changes

- Define data types for catch up in P6:
    - `CatchUpStatus`: summary of the current state of the node's consensus.
    - `CatchUpTerminalData`: collection of QCs, TCs, and quorum and timeout messages used to complete catch-up.
    - `CatchUpMessage`: a request, response or status message. All contain `CatchUpStatus`, and a response may contain `CatchUpTerminalData`.
- Functions for constructing and processing these datatypes.
- Integration with multiversion runner.
- `bufferedHandlePendingLive` is renamed to `bufferedSendCatchUpStatus` and changed to always send a catch-up status for the latest consensus version. This is used when a pending block becomes live, but is also used when a direct message block is received and it causes progress.
- The `NoGenesis` case is split out of the version 0 catch up status and promoted to its own constructor for `VersionedCatchUpStatus`. This is more convenient, as the code is used independent of the version.
- The result code `Unverifiable` is treated as indicating a peer should catch up.
- Successfully processed direct finalization messages are relayed to peers (as well as blocks and finalization records).

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
```diff
@@ -13,3 +54,490 @@ module Concordium.KonsensusV1.Consensus.CatchUp where
...
+            <$> sd ^. currentQuorumMessages . smBlockToWeightsAndSignatures
+    cusCurrentRoundTimeouts = makeTimeoutSet <$> sd ^. currentTimeoutMessages
+    getLeavesBranches _ curLeaves curBranches Seq.Empty = (curLeaves, curBranches)
+    getLeavesBranches parentBranches curLeaves curBranches (rest Seq.:|> cur) =
+        getLeavesBranches newParentBranches newLeaves newBranches rest
+      where
+        (newLeaves, newBranches, newParentBranches) = foldl' up (curLeaves, curBranches, Set.empty) cur
+        up (l, b, p) blk
+            | Set.member hsh parentBranches = (l, hsh : b, p')
+            | otherwise = (hsh : l, b, p')
```
abizjak, line 121: It would help readability if you provided a bit of a description of this. A type and description of arguments would suffice I think.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
```diff
@@ -13,3 +54,490 @@ module Concordium.KonsensusV1.Consensus.CatchUp where
...
+                            _ -> (id, id)
+                    return $
+                        filterFirst (Map.elems tmFirstEpochTimeouts)
+                            <> filterSecond (Map.elems tmSecondEpochTimeouts)
+        | otherwise =
+            return []
+
+data TerminalDataResult
+    = TerminalDataResultValid {tdrProgress :: Bool}
+    | TerminalDataResultInvalid {tdrProgress :: Bool}
```
abizjak, line 362: I'm struggling to understand this type, and the meaning of `tdrProgress`. Can you please add documentation explaining it.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
```diff
@@ -13,3 +54,490 @@ module Concordium.KonsensusV1.Consensus.CatchUp where
...
+                    | otherwise -> do
+                        -- Since 0 < hgt, we can be sure that 'fb' is not a genesis block.
+                        return
+                            CatchUpPartialResponseBlock
+                                { cuprNextBlock = toSignedBlock (LowLevel.stbBlock fb),
+                                  cuprContinue = sendBlocksFromHeight False (hgt + 1),
+                                  cuprFinish = return Absent
+                                }
+    catchUpAfterLastFinalized checkKnown = do
+        catchUpBranches checkKnown checkKnown [] (skovData ^. branches)
```
abizjak, line 286: Since this function is called without acquiring any locks, it's rather subtle that this will work correctly and send all the blocks, in the case that finalization happens between, say, `sendBlocksFromHeight` is called and then the continuation is called.

That does appear to be the case since the branches are just kept in memory, and the state is pure, not behind mvars or iorefs, but this is at all not clear just from looking at this function. It also crucially relies on `skovData` being passed in at the top-level.

So I would appreciate if a bit more explanation was given on why state changes in between invoking continuations are correctly handled here. Otherwise it's very hard to maintain this in the future.

abizjak (commented):
Looks good.

I have some questions and concerns, but I think they can easily be addressed.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
```diff
@@ -13,3 +54,490 @@ module Concordium.KonsensusV1.Consensus.CatchUp where
...
+A node N is up to date with a peer P if:
+    * N considers P's last finalized block to be finalized.
+    * The round of N's highest certified block is at least the round of P's highest certified block.
+    * N's current round (round(N)) is at least P's current round.
+    * N's current epoch is at least P's current epoch.
+    * N has all timeout signatures on round(N) that P has, and considers them valid if they are.
+    * N has all quorum signatures on round(N) that P has, and considers them valid if they are.
+    * If P has only one block in round(N), then N also has that block.
+
+If P is honest and considers a timeout signature on round(N) valid, then
```
EmilLa1, line 68: then what?

td202: I have removed this as I think the notes are not so useful any more.

td202 (approved):

Issue #910, opened 2023-06-22
"Unsuccesful decryption" results in peer being banned


td202:
**Bug Description**

A sporadic (and generally rare) error occurs where a peer gets banned as a result of an error in decrypting a message from a peer (immediately) after handshake.

```
2023-06-22T12:50:40.926751+02:00 DEBUG: Got a Handshake request from peer 9911a8b1b393f40e
2023-06-22T12:50:40.978805800+02:00 TRACE: External: Received finalization message, size = 101.
2023-06-22T12:50:41.026414800+02:00 DEBUG: Concluded handshake with peer 0000000000000003 (their id 9911a8b1b393f40e); wire protocol version 1
2023-06-22T12:50:41.077809900+02:00 DEBUG: Processed a finalization message from 0000000000000002
2023-06-22T12:50:41.127652700+02:00 ERROR: [receiving from peer 9911a8b1b393f40e] Unsuccesful decryption; the sender was not authenticated.
2023-06-22T12:50:41.175018200+02:00 DEBUG: Sent a broadcast containing a finalization message
2023-06-22T12:50:41.217003700+02:00 DEBUG: Closing the connection to peer 9911a8b1b393f40e
2023-06-22T12:50:41.253510800+02:00 WARN: Soft-banning 127.0.0.1 due to a breach of protocol
```

**Steps to Reproduce**

It is unclear what exactly triggers this, but it seems to happen after handshake with quite low frequency. Repeatedly starting networks of nodes is likely show the error occasionally.

**Expected Result**

The "Unsuccesful decryption" error should not occur, and consequently the peer should not be banned.

**Actual Result**

See above log.

**Versions**
- Software Version: 6.0.0 and earlier
- OS: Windows, Unix

PR #916, opened 2023-06-25
Don't drop messages just after handshake completes.

abizjak:
## Purpose

If we are not the initiator of the handshake then
- we receive message A
- we respond with a B message -> when they process this, they send a C message, and they promote us to a peer

If at this point they'd have any broadcast messages, they would send them to us.

The way we read from the underlying socket means we can in principle read multiple messages at the same time if they happen to be received already by the network stack (in low_level::read_from_socket).

In particular this means that the C message can be in the network stack's buffer together with a number of followup messages.

The way reading from the socket works is that we read into an internal buffer. And that buffer is consumed as we read from it, until we have to refresh it from the network socket again.

Previously we cleared the buffer after reading message C, which made us drop some messages.

As a result the nonces kept track of by the noise session are incorrect. In particular we can observe the following in the `encrypt_with_ad` function in noise
```
ENCRYPT KEY = "5d5322c66db62e375d8ffae7ca32fc0104f6554de384c72e82c1089b5fd4c2f7", NONCE = 2, AD = [], IN_OUT = 8a32f36233773e3ebdcb69db3e9ca4d33051fdaf047cc859ca675a7e77830e0533e1c6724bcea50443e7278ed4b82585180007c8f8eb2ead44244f542c5b498ef6fa3a53c4c3ad849f9953bb13323a030f08df5a780d8282c3e2a4f2f818c80432c4358597046ee1702041779099d6401aded9eab79f21378b07a9b4fde2eb3acf293a730f3e9ed19192736799cd86d8cf29f9b4e407a0aaa1f2ba88977faadcb0afe214b949ba4ac8b9a849a52478e273ea2958b597d147a401171ddda7b26dde242f0cb2b91439e9b208342e93004585d5fad0d9323317620029c56965279a5cc525356dfd96090c64732eab59ae76e1161ce461b2989a7f06ba5f6c6c166188efaec12ee5653a6936ff3e7d8a60a480127213f7fc842f70b89bce6c7508c171fe183931fe2d12449acc0a79b3bfeab91d464ee7a17626a0cec5560c145d8951c916ed6d7b3d64a137ec5da759c2a2d616f05fae31cf27c9f81acf99204e80db96af9a5cb957dd74b4df83534ddbd39debe9d02a296f26533843a10198670aa3be1cccf3546cdd0ac74f0bd2c70cbcc693ba8b28426499b42953774d418cb0fe7e437173a2f8d6c0c59a04de8722dbeb1a0fb7f6d44e812aeb3d3f8879d47724c1e545da9e5846201406010df101ba025953eb257629d0ab92efa678979189, MAC = 0460512e509b5e3bbdea8a302f35f408
```

but on the receiving end when decrypting,

```
KEY = "5d5322c66db62e375d8ffae7ca32fc0104f6554de384c72e82c1089b5fd4c2f7", NONCE = 0, AD = [], IN_OUT = 8a32f36233773e3ebdcb69db3e9ca4d33051fdaf047cc859ca675a7e77830e0533e1c6724bcea50443e7278ed4b82585180007c8f8eb2ead44244f542c5b498ef6fa3a53c4c3ad849f9953bb13323a030f08df5a780d8282c3e2a4f2f818c80432c4358597046ee1702041779099d6401aded9eab79f21378b07a9b4fde2eb3acf293a730f3e9ed19192736799cd86d8cf29f9b4e407a0aaa1f2ba88977faadcb0afe214b949ba4ac8b9a849a52478e273ea2958b597d147a401171ddda7b26dde242f0cb2b91439e9b208342e93004585d5fad0d9323317620029c56965279a5cc525356dfd96090c64732eab59ae76e1161ce461b2989a7f06ba5f6c6c166188efaec12ee5653a6936ff3e7d8a60a480127213f7fc842f70b89bce6c7508c171fe183931fe2d12449acc0a79b3bfeab91d464ee7a17626a0cec5560c145d8951c916ed6d7b3d64a137ec5da759c2a2d616f05fae31cf27c9f81acf99204e80db96af9a5cb957dd74b4df83534ddbd39debe9d02a296f26533843a10198670aa3be1cccf3546cdd0ac74f0bd2c70cbcc693ba8b28426499b42953774d418cb0fe7e437173a2f8d6c0c59a04de8722dbeb1a0fb7f6d44e812aeb3d3f8879d47724c1e545da9e5846201406010df101ba025953eb257629d0ab92efa678979189, MAC = 0460512e509b5e3bbdea8a302f35f408
```

Observe that the message, MAC, and public key are all the same, but the NONCE is different (the nonce is tracked per noise session).

Fixes https://github.com/Concordium/concordium-node/issues/910

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/connection/low_level.rs
+++ b/concordium-node/src/connection/low_level.rs
@@ -285,7 +285,7 @@ impl ConnectionLowLevel {
     fn process_msg_a(&mut self, len: usize) -> anyhow::Result<Vec<u8>> {
         recv_xx_msg!(self, len, "A");
         let pad = 16;
-        let payload_in = self.socket_buffer.slice(len)[DHLEN..][..len - DHLEN - pad].try_into()?;
+        let payload_in = self.socket_buffer.slice(len)[DHLEN..][..len - DHLEN - pad].to_vec();
         let payload_out = self.handler.upgrade().unwrap().produce_handshake_request()?; // safe
         send_xx_msg!(self, DHLEN * 2 + MAC_LENGTH, &payload_out, MAC_LENGTH, "B");
 
@@ -296,7 +296,7 @@ impl ConnectionLowLevel {
         recv_xx_msg!(self, len, "B");
         let payload_in = self.socket_buffer.slice(len)[DHLEN * 2 + MAC_LENGTH..]
             [..len - DHLEN * 2 - MAC_LENGTH * 2]
-            .try_into()?;
+            .to_vec();
         let payload_out = self.handler.upgrade().unwrap().produce_handshake_request()?; // safe
         send_xx_msg!(self, DHLEN + MAC_LENGTH, &payload_out, MAC_LENGTH, "C");
         self.socket.set_nodelay(false)?;
@@ -416,6 +416,12 @@ impl ConnectionLowLevel {
         self.incoming_msg.message.write_all(self.socket_buffer.slice(to_read))?;
         self.incoming_msg.pending_bytes -= to_read;
 
+        // When we are post-handshake, i.e., the noise session has been established
+        // we consume the data that was read from the buffer.
+        // Before the handshake completes the messages are handled a bit differently
+        // using `process_msg_*` methods, which read directly from the socket buffer
+        // so we cannot modify it here. Instead we consume the input after successfully
+        // reading the noise messages below.
         if self.is_post_handshake() {
             self.socket_buffer.shift(to_read);
         }
@@ -441,8 +447,7 @@ impl ConnectionLowLevel {
                     }
                 }
 
-                // If we reach this point we have read the entire noise message.
-                // So consume it from the socket buffer.
+                // We have processed an entire noise message, consume it from the socket buffer.
                 self.socket_buffer.shift(to_read);
                 Ok(ReadResult::Complete(payload))
             } else {
--- a/concordium-node/src/connection/low_level.rs
+++ b/concordium-node/src/connection/low_level.rs
@@ -307,7 +307,7 @@ impl ConnectionLowLevel {
         recv_xx_msg!(self, len, "C");
         let payload = self.socket_buffer.slice(len)[DHLEN + MAC_LENGTH..]
             [..len - DHLEN - MAC_LENGTH * 2]
-            .try_into()?;
+            .to_vec();
         self.socket.set_nodelay(false)?;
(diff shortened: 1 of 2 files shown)


concordium-node/src/connection/low_level.rs
```diff
@@ -441,7 +441,9 @@ impl ConnectionLowLevel {
                     }
                 }
 
-                self.socket_buffer.reset();
+                // If we reach this point we have read the entire noise message.
+                // So consume it from the socket buffer.
```
EmilLa1, line 445: You are here describing what is going on in the code below. A better comment would be describing **why**, i.e. there may still messages left in the buffer for us to read.

Also, isn't this only ever going to be a thing for the responder? As the cause of this, is that the initiator sends a message before the responder recognized he is past the handshake. 

Also, this will explain why the `shift` is not happening in the `if i'm responder` clause.

concordium-node/src/connection/low_level.rs
```diff
@@ -307,7 +307,7 @@ impl ConnectionLowLevel {
         recv_xx_msg!(self, len, "C");
         let payload = self.socket_buffer.slice(len)[DHLEN + MAC_LENGTH..]
             [..len - DHLEN - MAC_LENGTH * 2]
-            .try_into()?;
+            .to_vec();
```
EmilLa1, line 310: I suppose the same could be done for `process_msg_a` and `process_msg_b` ?

EmilLa1 (approved):

concordium-node/src/connection/low_level.rs
abizjak: Ok, I'll try to add a bit more.

The thing is, that it is kind of "obvious" that the new behaviour is the correct one. We read a message, we consume it from the buffer.

So in that sense there is not really a need to explain it in terms of the bug that we had, since that was just wrong behaviour. But I'll add more context.

td202 (approved):

concordium-node/src/connection/low_level.rs
abizjak: Changed also in those methods.

concordium-node/src/connection/low_level.rs
abizjak: I have added a bit of documentation above next to where we consume from the socket buffer for post-handshake connections only.

PR #938, opened 2023-06-30
New consensus database recovery

td202:
## Purpose

This changes the low-level tree state database so that certified blocks are stored. The reason for this is to allow the chain to fully recover its state necessary to be live after a restart.

## Changes

- Low-level tree state:
    - Both finalized and certified blocks are stored, indexed by block hash.
    - Finalized blocks are indexed by height.
    - Quorum certificates for certified blocks are stored, indexed by the round number.
    - `writeBlocks` is removed and replaced by `writeFinalizedBlocks`, `writeCertifiedBlock` and `writeCertifiedBlockWithFinalization`
    - `writeFinalizedBlocks` marks a series of certified blocks as finalized, removing the QCs and updating the latest finalization entry.
    - `writeCertifiedBlock` adds a new certified block.
    - `writeCertifiedBlockWithFinalization` adds a certified block and marks blocks as finalized in a single transaction. It is equivalent to doing both `writeCertifiedBlock` and `writeFinalizedBlocks`.
    - `lookupLastBlock` is renamed to `lookupLastFinalizedBlock`.
    - `lookupCertifiedBlocks` is added for getting the non-finalized certified blocks, that is required when loading the database.
    - The most recent timeout certificate is added to the persistent round status.
    - `rollBackBlocksUntil` is removed from the `MonadTreeStateStore` and only provided for the LMDB implementation. Its behaviour is changed to roll back certified blocks as well as finalized block where necessary, but to ensure that the latest finalization record is always valid.
- `finalizingCertifiedBlock` is removed from `SkovData`, and replaced with `latestFinalizationEntry`.
- `loadSkovData`: the QC for the last finalized block is added to the `roundExistingQCs`; the `latestFinalizationEntry` is loaded from the low-level database.
- `loadCertifiedBlocks` is added, which is called as part of `activateSkovV1State` when loading the state. It deals with loading certified blocks into the state (including adding their transactions to the transaction table) and also sets up the previous round timeout based on the stored timeout certificate (where appropriate). It should establish that the current round is the next one from the last we have a certified block are timeout certificate for. (Previously, as certified blocks and timeout certificates were not stored, this could not be guaranteed.)
- Finality:
    - `checkFinality`/`checkFinalityWithBlock` is replaced by `processCertifiedBlock`, which will store the certified block, as well as doing the finality checking and processing that was part of `checkFinality`.
    - `processFinalizationEntry` provides another route to finalization given a finalization entry instead of a certified block. (This is for when we might not have the successor block itself, such as during catch-up or in the case of an epoch transition. `processFinalization` was previously used for this purpose.)
    - `processFinalizationHelper` replaces `processFinalization` as an internal helper for implementing `processCertifiedBlock` and `processFinalizationEntry`.
- Catch up:
    - `CatchUpTerminalData` replaces `cutdQuorumCertificates` with `cutdLatestFinalizationEntry` and `cutdHighestQuorumCertificate`.
    - `cutdLatestFinalizationEntry` is used to send the latest finalization entry whenever the peer is behind on the last finalized block. It is processed as part of `processCatchUpTerminalData`.
    - `cutdHighestQuorumCertificate` is used to send the highest QC when the peer is behind on the highest certified block.
- Block state:
    - `cacheBlockState` and `cacheBlockStateAndGetTransactionTable` are added to the `BlockStateStorage` type class. This change is to make them usable inside the `SkovV1T` monad transformer.
    - `cacheBlockStateAndGetTransactionTable` is implemented by the existing `cacheStateAndGetTransactionTable`
    - `cacheBlockState` has a similar implementation to `cacheBlockStateAndGetTransactionTable`, but without building the transaction table.
- `activateSkovV1State` is revised so that it runs in the `SkovV1T` monad transformer, which slightly simplifies how it is implemented and used. It calls `loadCertfiedBlocks`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
@@ -37,7 +37,9 @@ import Concordium.KonsensusV1.Types
 -- If so, the block and its ancestors are finalized, the tree is pruned to the decendants of the
 -- new last finalized block, and, if applicable, the epoch is advanced.
 -- If the block is already written, then it is assumed that it has already been processed in this
--- manner, and so no further action is taken.
+-- manner, and so no further action is taken. (Normally, `processCertifiedBlock` should not be
+-- called on a block that is already finalized, but it can happen if a roll-back occurred at
+-- start up.)
 --
 -- This function incorporates the functionality of @checkFinality@ from the bluepaper.
 processCertifiedBlock ::
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
@@ -36,10 +36,10 @@ import Concordium.KonsensusV1.Types
 -- Check if the certified block causes its parent to become finalized.
 -- If so, the block and its ancestors are finalized, the tree is pruned to the decendants of the
 -- new last finalized block, and, if applicable, the epoch is advanced.
+-- If the block is already written, then it is assumed that it has already been processed in this
+-- manner, and so no further action is taken.
 --
 -- This function incorporates the functionality of @checkFinality@ from the bluepaper.
---
--- PRECONDITION: the certified block is valid and live (and not finalized).
 processCertifiedBlock ::
     ( MonadState (SkovData (MPV m)) m,
       TimeMonad m,
@@ -60,7 +60,7 @@ processCertifiedBlock cb@CertifiedBlock{..}
     | NormalBlock block <- bpBlock cbQuorumBlock,
       let parentQC = blockQuorumCertificate block,
       qcRound cbQuorumCertificate == qcRound parentQC + 1,
-      qcEpoch cbQuorumCertificate == qcEpoch parentQC = do
+      qcEpoch cbQuorumCertificate == qcEpoch parentQC = unlessStored $ do
         let finalizedBlockHash = qcBlock parentQC
         sd <- get
         unless (finalizedBlockHash == getHash (sd ^. lastFinalized)) $ do
@@ -73,11 +73,13 @@ processCertifiedBlock cb@CertifiedBlock{..}
                         }
             processFinalizationHelper newFinalizedPtr newFinalizationEntry (Just cb)
             shrinkTimeout cbQuorumBlock
-    | otherwise = do
+    | otherwise = unlessStored $ do
+        storedBlock <- makeStoredBlock cbQuorumBlock
+        LowLevel.writeCertifiedBlock storedBlock cbQuorumCertificate
+  where
+    unlessStored a = do
         alreadyStored <- LowLevel.memberBlock (getHash cbQuorumBlock)
-        unless alreadyStored $ do
-            storedBlock <- makeStoredBlock cbQuorumBlock
-            LowLevel.writeCertifiedBlock storedBlock cbQuorumCertificate
+        unless alreadyStored a
 
 -- |Process a finalization entry that finalizes a block that is not currently considered finalized.
 --
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
@@ -32,13 +32,61 @@ import qualified Concordium.KonsensusV1.TreeState.LowLevel as LowLevel
 import Concordium.KonsensusV1.TreeState.Types
 import Concordium.KonsensusV1.Types
 
--- |Check if a valid quorum certificate finalizes a block. If so, the block and its ancestors
--- are finalized, the tree is pruned to the decendants of the new last finalized block, and,
--- if applicable, the epoch is advanced.
+-- |Ensure that the given certified block is written to the low-level database.
+-- Check if the certified block causes its parent to become finalized.
+-- If so, the block and its ancestors are finalized, the tree is pruned to the decendants of the
+-- new last finalized block, and, if applicable, the epoch is advanced.
 --
--- PRECONDITION: the quorum certificate is valid. (If the target of the QC is not live, this
--- function will not make any change to the state, and will not error.)
-checkFinality ::
+-- This function incorporates the functionality of @checkFinality@ from the bluepaper.
+--
+-- PRECONDITION: the certified block is valid and live (and not finalized).
+processCertifiedBlock ::
+    ( MonadState (SkovData (MPV m)) m,
+      TimeMonad m,
+      MonadIO m,
+      LowLevel.MonadTreeStateStore m,
+      BlockStateStorage m,
+      GSTypes.BlockState m ~ PBS.HashedPersistentBlockState (MPV m),
+      MonadThrow m,
+      MonadConsensusEvent m,
+      MonadLogger m,
+      IsConsensusV1 (MPV m),
+      HasCallStack
+    ) =>
+    -- |The newly-certified block.
+    CertifiedBlock (MPV m) ->
+    m ()
+processCertifiedBlock cb@CertifiedBlock{..}
+    | NormalBlock block <- bpBlock cbQuorumBlock,
+      let parentQC = blockQuorumCertificate block,
+      qcRound cbQuorumCertificate == qcRound parentQC + 1,
+      qcEpoch cbQuorumCertificate == qcEpoch parentQC = do
+        let finalizedBlockHash = qcBlock parentQC
+        sd <- get
+        unless (finalizedBlockHash == getHash (sd ^. lastFinalized)) $ do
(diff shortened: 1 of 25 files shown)


concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -504,55 +648,250 @@ instance
...
+                                            (finHash : hashes)
+                                            parent
+                                            (blockQuorumCertificate block)
+                                    else do
+                                        -- The block does not have a timeout certificate or finalization
+                                        -- entry, so the parent block is explicitly finalized. We thus
+                                        -- set the latest finalization entry to finalize the
+                                        -- parent block so that the database invariant is
+                                        -- restored.
+                                        storeRecord
```
abizjak, line 864: Shouldn't this be `storeReplaceRecord` rather than `storeRecord`?

I'd expect that we need to overwrite if it already exists.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/StartUp.hs
```diff
@@ -209,3 +231,88 @@ loadSkovData _runtimeParameters = do
...
+                          rtCertifiedBlock = highCB
+                        }
+            roundStatus . rsCurrentRound .= tcRound lastTimeout + 1
+    rs <- use roundStatus
+    let expectedCurrentRound
+            | Present prevTO <- rs ^. rsPreviousRoundTimeout = 1 + tcRound (rtTimeoutCertificate prevTO)
+            | otherwise = 1 + cbRound (rs ^. rsHighestCertifiedBlock)
+    unless (expectedCurrentRound == rs ^. rsCurrentRound) $
+        throwM . TreeStateInvariantViolation $
+            "The current round does not match the expected round."
```
abizjak, line 285: I don't think this check makes sense in the presence of rollbacks.

Suppose we have no timeouts, and for some reason or database (block state) gets corrupt. Then we'll have deleted a number of certified blocks first in `initialiseExistingSkov`.

But if I understand correctly how `rsCurrentRound` is defined, it is always at least the one stored in the `roundStatusStore`. So these two together cannot really work unless I'm misunderstanding something.

I suppose we don't want to roll back the persistent round status, so we don't send contradicting messages.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -229,8 +229,9 @@ data SkovData (pv :: ProtocolVersion) = SkovData
       _skovPendingBlocks :: !PendingBlocks,
       -- |Pointer to the last finalized block.
       _lastFinalized :: !(BlockPointer pv),
-      -- |Certified block that justifies the last finalized block being finalized.
-      _finalizingCertifiedBlock :: !(Option (CertifiedBlock pv)),
+      -- |A finalization entry that finalizes the last finalized block, unless that is the
```
EmilLa1, line 232: the `_branches` above. Isn't it an invariant that (certified blocks) in the branches must be present also in the lmdb database?

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -448,6 +484,79 @@ asWriteTransaction t = do
         (LMDB_Error _ _ (Right MDB_MAP_FULL)) -> Just ()
         _ -> Nothing
 
+-- |Helper function for implementing 'writeFinalizedBlocks'.
+writeFinalizedBlocksHelper ::
```
EmilLa1, line 488: We could simply pass in the pruned blocks (from the `PrunedResult` at the call site, and delete those to make this a bit simpler)
i.e. 

```suggestion
writeFinalizedBlocksHelper ::writeFinalizedBlocksHelper ::
    (HasDatabaseHandlers dbh pv, IsProtocolVersion pv) =>
    [BlockHash] ->
    [StoredBlock pv] ->
    FinalizationEntry ->
    dbh ->
    MDB_txn ->
    IO [BlockHash]
writeFinalizedBlocksHelper prunedBlocks finBlocks finEntry dbh txn = do    
    -- 1. Delete the pruned blocks from the blocks table. 
    -- 2. Delete the QC's which have a round less than or equal to succQC of the finalization entry.
    mapM_ (deleteRecord txn (dbh ^. blockStore)) prunedBlocks
    withCursor txn (dbh ^. nonFinalizedQuorumCertificateStore) $ \cursor -> do
        let loop' Nothing = return ()
            loop' (Just (Left e)) = throwM (DatabaseInvariantViolation e)
            loop' (Just (Right (rnd, _))) = do
                case compare rnd (qcRound $ feSuccessorQuorumCertificate finEntry) of
                    LT -> do
                        deleteAtCursor cursor
                        loop' =<< getCursor CursorNext cursor
                    GT -> do
                        deleteAtCursor cursor
                        loop' =<< getCursor CursorNext cursor
                    _ -> loop' Nothing                    
        loop' =<< getCursor CursorFirst cursor
```

concordium-consensus/src/Concordium/KonsensusV1/TreeState/StartUp.hs
```diff
@@ -209,3 +231,88 @@ loadSkovData _runtimeParameters = do
...
+            .= CertifiedBlock
+                { cbQuorumCertificate = qc,
+                  cbQuorumBlock = blockPointer
+                }
+
+        curRound <- use $ roundStatus . rsCurrentRound
+        when (blockRound blockPointer >= curRound) $ do
+            roundStatus
+                %=! (rsPreviousRoundTimeout .~ Absent)
+                . (rsCurrentRound .~ blockRound blockPointer + 1)
```
EmilLa1, line 313: Shouldn't this be `qcRound qc + 1`  ?

td202: Since the QC is for the block, these should be the same.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -504,55 +648,250 @@ instance
...
+--   * The database invariants hold.
+--   * No other concurrent accesses are made to the database.
+--
+-- The updates to the database are performed in the following write transactions:
+--   * Removing a single certified block.
+--   * Removing all certified blocks.
+--   * Removing the last finalized block and implicitly finalized blocks, rolling back the
+--     latest finalization entry to the prior explicitly finalized block (or removing it if
+--     it would be for the genesis block).
+rollBackBlocksUntil ::
```
EmilLa1, line 679: I believe we should also clear the "last timeout certificate" when rolling back blocks.

td202: We won't clear it, but if it's not coherent with the new highest certified block we will warn instead of crashing.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/StartUp.hs
td202: I have now changed how `rsCurrentRound` is defined, so it is the highest round for which we have a valid QC or TC for the previous round (i.e. it is justified in the current state). Additional checks should be used to prevent baking and/or signing for "old" rounds.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
```diff
@@ -48,23 +50,45 @@ checkFinality ::
...
+                          feSuccessorQuorumCertificate = cbQuorumCertificate,
+                          feSuccessorProof = getHash (sbBlock block)
+                        }
+            processFinalizationHelper newFinalizedPtr newFinalizationEntry (Just cb)
+            shrinkTimeout cbQuorumBlock
+    | otherwise = unlessStored $ do
+        storedBlock <- makeStoredBlock cbQuorumBlock
+        LowLevel.writeCertifiedBlock storedBlock cbQuorumCertificate
+  where
+    unlessStored a = do
```
EmilLa1, line 80: Please document that this is to cover the case where we have rolled back blocks (and we don't "demote" a finalized block as part of this process).

EmilLa1 (approved):

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
td202: This is not really an invariant of `_branches`. I'll add a comment on the block table.

abizjak (approved):
Approving with the expectation that more work will be done to get to the bottom of the issues that were discovered.

eb-concordium (approved):

PR #957, opened 2023-07-06
Avoid busy-looping when minimum block time is set to 0.

### LINKED ISSUE #951: Node does not properly start if single baker with no min block time
**Bug Description**

When `minBlockTime` is set to 0, or close to it, e.g., 1ms, then
```
startEvents = do
    isShutdown <- use isConsensusShutdown
    unless isShutdown $ do
        genesisTime <- timestampToUTCTime . BaseV1.genesisTime . gmParameters <$> use genesisMetadata
        doAfter genesisTime $ do
            resetTimerWithCurrentTimeout
            makeBlock
```

does not ever terminate if the node is the only baker so the node does not properly start, in particular the rpc services do not start, nor is the signal handler properly installed.

In particular the processing of network messages is not started, meaning that the outbound consensus queues get full, leading the node to stop producing blocks since `sendBlock` blocks.

**Versions**
- Software Version: 6.0.{0,1}.

td202:
## Purpose

Fixes #951

When the minimum block time is 0, or close to it, and when there is only one baker, the baker would loop at start up and never yield. This changes the behaviour so that the baker yields control (releasing the write lock) at least once per block.

## Changes

- On start up, always wait on a timer before starting baking, even if the genesis time has elapsed. This returns control to the P2P layer sooner to allow start-up operations to complete.
- After baking a block, always wait on a timer before sending (and validating) the block. This ensures that we do not hold the write lock in a tight loop in the 1-baker 0-minimum-block-time case.
- Bypass the delay on a timer thread if it is negative. It seems this may have caused misbehaviour.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 
+- Fix a bug where a setup with a single baker and a minimum block time of 0s would result in an
+  unresponsive node in protocol version 6.
+
 ## 6.0.1
 
 - Remove configuration option `no-rpc-server` and environment variable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fix a bug where a setup with a single baker and a minimum block time of 0s would result in an
   unresponsive node in protocol version 6.
+- Fix a bug where receiving a duplicate of an invalid block could be spuriously reported as double
+  signing.
 
 ## 6.0.1
 
--- a/concordium-consensus/src/Concordium/KonsensusV1.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1.hs
@@ -113,6 +113,7 @@ startEvents = do
     isShutdown <- use isConsensusShutdown
     unless isShutdown $ do
         genesisTime <- timestampToUTCTime . BaseV1.genesisTime . gmParameters <$> use genesisMetadata
-        doAfter genesisTime $ do
+        void . onTimeout (DelayUntil genesisTime) $ do
+            logEvent Konsensus LLInfo "Starting consensus operations."
             resetTimerWithCurrentTimeout
             makeBlock
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
@@ -278,16 +278,25 @@ receiveBlockKnownParent parent pendingBlock = do
             let blockWitness = toBlockSignatureWitness (pbBlock pendingBlock)
             -- Check if we've already seen a signed block from this baker in this round.
             use (roundBakerExistingBlock (blockRound pendingBlock) (blockBaker pendingBlock)) >>= \case
-                Just w -> do
-                    -- If the baker has already signed a block in this round then we flag it.
-                    logEvent Konsensus LLDebug $
-                        "Baker "
-                            <> show (blockBaker pendingBlock)
-                            <> " signed multiple blocks in round "
-                            <> show (blockRound pendingBlock)
-                            <> "."
-                    flag (DuplicateBlock w blockWitness)
-                    return $ BlockResultDoubleSign verifiedBlock
+                Just w
+                    | bswBlockHash w == pbHash -> do
+                        -- This case is uncommon, as duplicates are typically detected earlier, but
+                        -- it can happen if:
+                        --   1. the block bypasses de-duplication (i.e. as a direct message), and
+                        --   2. the block is invalid, but has a valid baker signature.
+                        -- We thus check if the block is actually distinct so that we do not
+                        -- spuriously report it as a double signing.
+                        return BlockResultDuplicate
+                    | otherwise -> do
+                        -- If the baker has already signed a block in this round then we flag it.
+                        logEvent Konsensus LLDebug $
+                            "Baker "
+                                <> show (blockBaker pendingBlock)
+                                <> " signed multiple blocks in round "
+                                <> show (blockRound pendingBlock)
+                                <> "."
+                        flag (DuplicateBlock w blockWitness)
+                        return $ BlockResultDoubleSign verifiedBlock
                 Nothing -> do
                     -- If the baker has not already signed a block in this round, we record that
                     -- it has now and return control to the caller for retransmitting the block.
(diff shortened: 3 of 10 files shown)


eb-concordium (approved):

abizjak:
> Just some comments from somebody who doesn't understand Haskell: I don't understand why the old code would cause this issue and how the new code fixes it. The change avoids the function `doAfter`. If that function does what its name suggests, there should not be an issue. Is there a bug in that function? Could it cause issues elsewhere and we should generally avoid it? This is worrying me because it seems similar issues could have catastrophic consequences and it is completely unexpected to happen here (again, for somebody who doesn't understand Haskell).

`doAfter` does not spawn a background computation in case the time has already passed. Whereas onTimeout does regardless if the time has already passed.

`doAfter` works correctly, but there is an issue in the sense that we have a bunch of other services that must be given opportunity to start/do work. In particular the way the node is set up. When starting up we call startConsensus which we expect to terminate and give us back a pointer to the consensus component so we can start other services (such as the grpc server, network layer) and give them access to the consensus component. But since `doAfter` never yields in the cases where blocks come too quickly, this never happens. Which then causes odd node behaviour.

Did this help in any way?

So if we are just looking at consensus in isolation then `doAfter` is still correct.

abizjak (approved):
Looks good.

there are still some cases where the node cannot be properly killed but they are not reproducible. This certainly improves things.

chrmatt:
> > Just some comments from somebody who doesn't understand Haskell: I don't understand why the old code would cause this issue and how the new code fixes it. The change avoids the function `doAfter`. If that function does what its name suggests, there should not be an issue. Is there a bug in that function? Could it cause issues elsewhere and we should generally avoid it? This is worrying me because it seems similar issues could have catastrophic consequences and it is completely unexpected to happen here (again, for somebody who doesn't understand Haskell).
> 
> `doAfter` does not spawn a background computation in case the time has already passed. Whereas onTimeout does regardless if the time has already passed.
> 
> `doAfter` works correctly, but there is an issue in the sense that we have a bunch of other services that must be given opportunity to start/do work. In particular the way the node is set up. When starting up we call startConsensus which we expect to terminate and give us back a pointer to the consensus component so we can start other services (such as the grpc server, network layer) and give them access to the consensus component. But since `doAfter` never yields in the cases where blocks come too quickly, this never happens. Which then causes odd node behaviour.
> 
> Did this help in any way?
> 
> So if we are just looking at consensus in isolation then `doAfter` is still correct.

Thanks for the explanation! I think I now partially understand what's going on. But it seems still a bit strange to me. If we need the block production to be executed in a background task, why don't we explicitly start one for doing so instead of relying on the behavior of some wait-function to implicitly do so?

abizjak:
@chrmatt Well in the two places changed in this PR we need both to wait and to then do in background. So the `onTimeout` function fits. Not sure exactly what you would change.

EmilLa1 (approved):

EmilLa1:
> > > Just some comments from somebody who doesn't understand Haskell: I don't understand why the old code would cause this issue and how the new code fixes it. The change avoids the function `doAfter`. If that function does what its name suggests, there should not be an issue. Is there a bug in that function? Could it cause issues elsewhere and we should generally avoid it? This is worrying me because it seems similar issues could have catastrophic consequences and it is completely unexpected to happen here (again, for somebody who doesn't understand Haskell).
> > 
> > 
> > `doAfter` does not spawn a background computation in case the time has already passed. Whereas onTimeout does regardless if the time has already passed.
> > `doAfter` works correctly, but there is an issue in the sense that we have a bunch of other services that must be given opportunity to start/do work. In particular the way the node is set up. When starting up we call startConsensus which we expect to terminate and give us back a pointer to the consensus component so we can start other services (such as the grpc server, network layer) and give them access to the consensus component. But since `doAfter` never yields in the cases where blocks come too quickly, this never happens. Which then causes odd node behaviour.
> > Did this help in any way?
> > So if we are just looking at consensus in isolation then `doAfter` is still correct.
> 
> Thanks for the explanation! I think I now partially understand what's going on. But it seems still a bit strange to me. If we need the block production to be executed in a background task, why don't we explicitly start one for doing so instead of relying on the behavior of some wait-function to implicitly do so?

I suppose, in principle we could have a thread that listened checked for the current round of the tree state, and then called "makeBlock" if it was progressed, but I think the current solution is much simpler as it can only ever be triggered when advancing a round either by a quorum certificate or a timeout certificate (hence we're calling `makeBlock` after doing so). 

And for the first startEvents (that calls `makeBlock` is to sort of bootstrap the chain in the first round (i.e. give the nodes a block to sign off), as otherwise we would always timeout the first round.

chrmatt:
I think my main issue is that things seem to happen implicitly. It is apparently very important that things happen in the background, and `onTimeout` does that always, while `doAfter` also does that, but only in certain cases. I'd expect that, e.g., in the initial setup, we explicitly spawn a thread for making blocks (either using a function that makes clear that it's doing so or using comments), and then once we are inside the baking thread, we can bake more blocks using timeouts, which may or may not spawn further threads, which should not break things either way (or if it does, this should again be explicit).

td202:
> I think my main issue is that things seem to happen implicitly. It is apparently very important that things happen in the background, and `onTimeout` does that always, while `doAfter` also does that, but only in certain cases. I'd expect that, e.g., in the initial setup, we explicitly spawn a thread for making blocks (either using a function that makes clear that it's doing so or using comments), and then once we are inside the baking thread, we can bake more blocks using timeouts, which may or may not spawn further threads, which should not break things either way (or if it does, this should again be explicit).

Our design uses more of an event-based model, rather than having a worker thread. So baking is triggered by advancing the round. Instead of signalling a background thread to wake up and bake a block, we just bake on the current thread when we advance the round. This is generally a bit more efficient*, because it avoids context switching, and we already hold the lock on the state. I think it's also a bit more explicit, because with a thread the coordination can obscure the ordering.

Another way of thinking of it is that the runtime system itself is maintaining the worker thread(s), and when we `onTimeout` or `doAfter`, those threads will pick up the timer event and execute the action.

For reference, `doAfter` is implemented using `onTimeout`:
```
-- |If the given time has elapsed, perform the supplied action. Otherwise, start a timer to
-- asynchronously perform the action at the given time.
doAfter :: (TimeMonad m, TimerMonad m) => UTCTime -> m () -> m ()
doAfter time action = do
    now <- currentTime
    if time <= now
        then action
        else void $ onTimeout (DelayUntil time) action
```
Typically, it's fine to just run the action in the current thread - as long as it's not going to take too long. The problem arises in the case of 0s minimum block time that we end up with an infinite chain of baking that all ends up getting run immediately.

chrmatt:
@td202 Thanks for the explanation, I think that makes sense.

Do we have a place where the overall architecture is described? It's quite hard for an outsider to understand what is going on and the discussion we are having here could also be useful for others looking at the code at some later point.

Issue #947, opened 2023-07-04
Block state database truncation may fail on Windows


td202:
**Bug Description**

If the block state file is shorter than expected, then it should be truncated to a successfully loadable block state at start up. On Windows, this can sometimes fail with an "Invalid argument" error.

**Steps to Reproduce**

1. Run a node on Windows.
2. Shut down the node.
3. Truncate the block file by a few bytes.
4. Restart the node.

**Expected Result**

The node should start, truncate the file, and continue to operate normally.

**Actual Result**

The node fails with an "Invalid argument" error.

**Versions**
- Software Version 6.0.0 (and likely earlier)
- OS Windows

PR #960, opened 2023-07-07
Database and catch-up fixes

### LINKED ISSUE #958: Catch up loops or soft-bans peer
**Bug Description**

In-band catch-up can be observed to fail in the following ways:
- After receiving the catch-up response, the node soft-bans the peer, citing "Catching up with peer {} resulted in incompatible globalstates, dropping and soft-banning".
- After receiving the catch-up response, the node repeatedly attempts to catch-up with the peer. This ends when a new block is produced and distributed.

In both cases, this is caused by the node failing to properly process the catch-up response. In particular, the highest quorum certificate is not processed where the QC is also part of the finalization entry. If there is also a timeout certificate or quorum/timeout messages, these will cause the response to be rejected resulting in a ban. (This is since failing to process the QC results in a state where they cannot be processed.) Otherwise, the node will detect that it has still not caught up successfully and retry, resulting in the looping behaviour. When a block is eventually produced, this (should) lead to the highest QC actually getting updated, ending the loop.

**Steps to Reproduce**

Run a chain with multiple nodes. (Minimum block time should be at least 1 second to give a chance of observing the behaviour.) Have a node catch-up. To observe the soft-ban behaviour, it is easiest to have some bakers offline, so that timeouts occur - especially if the catching-up node is required to make the 2/3 committee online.

**Expected Result**

The node should not loop in catch-up or ban its (well-behaved) peers.

**Actual Result**

The node may loop in catch-up or ban peers.

**Versions**
- Software Version: 6.0.0, 6.0.1
- OS: All

td202:
## Purpose

This addresses:

#947: Database truncation fails during roll-back on Windows
#958: Catch-up loops or soft-bans peers

## Changes

- When attempting to read beyond the end of the block state database, the file is remapped. This is changed so that the file is not remapped unless it is longer than the existing map. This addresses #947 because the file cannot be truncated on Windows while it is memory-mapped. The roll-back process can attempt to read beyond the end of the database multiple times, causing the file to be mapped repeatedly (prior to this change). When truncating the file, only the latest mapping is forced to close, and thus the other mappings can cause failure.
- Process the quorum certificate for the highest certified block before the finalization entry. Also, change the criterion for doing so to be that it will advance the round (rather than we have already checked such a QC). This resolves #958.
- Fix an issue with accounting for which blocks are rolled back. This mainly affects logging, but could cause some misbehaviour if only one block is rolled back.
- Additional trace-level logging of catch-up messages.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
@@ -490,9 +490,10 @@ processCatchUpTerminalData CatchUpTerminalData{..} = flip runContT return $ do
                         currentProgress
                         "quorum certificate is inconsistent with the block it certifies."
                     )
-                use (roundExistingQuorumCertificate (qcRound qc)) >>= \case
-                    Just _ -> return currentProgress
-                    Nothing -> do
+                -- We only care if the QC will advance our current round.
+                currentRound <- use $ roundStatus . rsCurrentRound
+                if currentRound == qcRound qc
+                    then do
                         gets (getBakersForEpoch (qcEpoch qc)) >>= \case
                             Nothing -> return currentProgress
                             Just BakersAndFinalizers{..} -> do
@@ -520,6 +521,7 @@ processCatchUpTerminalData CatchUpTerminalData{..} = flip runContT return $ do
                                         advanceRoundWithQuorum
                                             newCertifiedBlock
                                     return True
+                    else return currentProgress
             Nothing -> return currentProgress
     processTC currentProgress Absent = return currentProgress
     processTC currentProgress (Present tc) = do
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
@@ -435,8 +435,8 @@ processCatchUpTerminalData ::
     CatchUpTerminalData ->
     m TerminalDataResult
 processCatchUpTerminalData CatchUpTerminalData{..} = flip runContT return $ do
-    progress0 <- processFE False cutdLatestFinalizationEntry
-    progress1 <- foldM processQC progress0 cutdHighestQuorumCertificate
+    progress0 <- foldM processQC False cutdHighestQuorumCertificate
+    progress1 <- processFE progress0 cutdLatestFinalizationEntry
     progress2 <- processTC progress1 cutdTimeoutCertificate
     progress3 <- foldM processQM progress2 cutdCurrentRoundQuorumMessages
     progress4 <- foldM processTM progress3 cutdCurrentRoundTimeoutMessages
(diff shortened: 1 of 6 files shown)


concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -851,7 +851,20 @@ rollBackBlocksUntil checkState = do
             throwM . DatabaseRecoveryFailure $
                 "Genesis block state could not be recovered."
         (count', hashes, newLastFin) <- asWriteTransaction $ \dbh txn -> do
-            let loop !c hashes fin finQC = case stbBlock fin of
+            let loop ::
+                    -- Current count of blocks rolled back
+                    Int ->
+                    -- Accumulated list of rolled-back finalized blocks
```
abizjak, line 857: Can you elaborate on the order of this list please.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -851,7 +851,20 @@ rollBackBlocksUntil checkState = do
...
+                    [BlockHash] ->
+                    -- Block to roll back
+                    StoredBlock pv ->
+                    -- Quorum certificate on the block
+                    QuorumCertificate ->
+                    -- Total number of blocks rolled back,
+                    -- List of hashes of rolled-back blocks,
```
abizjak, line 864: And here.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
```diff
@@ -435,8 +435,8 @@ processCatchUpTerminalData ::
     CatchUpTerminalData ->
     m TerminalDataResult
 processCatchUpTerminalData CatchUpTerminalData{..} = flip runContT return $ do
-    progress0 <- processFE False cutdLatestFinalizationEntry
-    progress1 <- foldM processQC progress0 cutdHighestQuorumCertificate
+    progress0 <- foldM processQC False cutdHighestQuorumCertificate
+    progress1 <- processFE progress0 cutdLatestFinalizationEntry
```
abizjak, line 439: Seeing that there seemed to have been a bug here, can you please add documentation as to why this order is correct, and not the previous one?

abizjak (approved):

concordium-consensus/src/Concordium/KonsensusV1/Consensus/CatchUp.hs
td202: The old ordering isn't wrong per se, but this is more efficient. The bug was caused by relying on `roundExistingQCs` to determine if the QC was worth processing. While the finalization entry was generally the trigger for the bug (because it was processed first), the bug could still potentially be triggered even with the reordering.

abizjak: Thanks. But if this is more efficient than that is still useful insight that should be documented.

PR #962, opened 2023-07-11
removed rpc_server

lassemand:
## Purpose

Remove unused map as well as ensure we have default params for flynet (Should be overwritten anyways from within ansible though)

## Changes

Add default parameters for flynet ports
Remove rpc server map

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/scripts/distribution/ubuntu-packages/template/debian/concordium-node.service
+++ b/scripts/distribution/ubuntu-packages/template/debian/concordium-node.service
@@ -45,10 +45,6 @@ Environment=CONCORDIUM_NODE_CONNECTION_BOOTSTRAP_NODES=${build_bootstrap}
 Environment=CONCORDIUM_NODE_CONNECTION_DESIRED_NODES=5
 # maximum number of __nodes__ the node will be connected to.
 Environment=CONCORDIUM_NODE_CONNECTION_MAX_ALLOWED_NODES=10
-# address of the GRPC server
-Environment=CONCORDIUM_NODE_RPC_SERVER_ADDR=0.0.0.0
-# and its port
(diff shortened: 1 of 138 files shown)


scripts/distribution/ubuntu-packages/template/debian/concordium-node.service
```diff
@@ -45,10 +45,6 @@ Environment=CONCORDIUM_NODE_CONNECTION_BOOTSTRAP_NODES=${build_bootstrap}
 Environment=CONCORDIUM_NODE_CONNECTION_DESIRED_NODES=5
 # maximum number of __nodes__ the node will be connected to.
 Environment=CONCORDIUM_NODE_CONNECTION_MAX_ALLOWED_NODES=10
-# address of the GRPC server
-Environment=CONCORDIUM_NODE_RPC_SERVER_ADDR=0.0.0.0
```
lassemand, line 49: So this is what we want to remove in general. We are trying to move away from this API, and we want to make this environment variable conditional. If we do not specify it, will still be present from here. Even worse the build rpc_server:port will be null. But because we overwrite from within the ansible script it has so far been working.

abizjak (approved):

scripts/distribution/ubuntu-packages/template/debian/concordium-node.service
abizjak: What is this comment about? THis has been removed some time ago.

lassemand: Yes. This PR now only removes the map

PR #963, opened 2023-07-12
Enable out-of-band catch-up on Windows by default

td202:
## Purpose

Update the node configuration set up by the Windows installer to enable out-of-band catch-up by default.

## Changes

- Set the `CONCORDIUM_NODE_CONSENSUS_DOWNLOAD_BLOCKS_FROM` environment variable for the node for mainnet and testnet configurations.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/service/windows/installer/resources/nodes.toml
+++ b/service/windows/installer/resources/nodes.toml
@@ -48,6 +48,8 @@ collector.enabled = false
 # non-baker nodes. It significantly reduces CPU use when the node's database is
 # large and seems to have little downside.
 node.env.CONCORDIUM_NODE_RUNTIME_HASKELL_RTS_FLAGS = '-I0'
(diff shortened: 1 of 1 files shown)


abizjak (changes requested):
I don't want to add this to the 6.0 release. We should add this to all or none of the distributions, and then we need to make sure we don't have problems with it during testing.

It can go into the following release in October, where also we can make sure that there are no issues on any of t he distributions.

limemloh (approved):

abizjak:
@td202 Please make a task for this for the next 6.1 release to cover
- updating all distributions
- making sure we update all our documentation to reflect the new defaults

abizjak (approved):
Rasmus will do other distributions and update the documentation. This can be merged now.

PR #964, opened 2023-07-12
Prevent the baker thread from starting when it is not needed.

### LINKED ISSUE #952: Confusing logging when starting at P6.
**Bug Description**

The logging when the node starts in P6 is not great

```
2023-07-05T20:38:38.007348457Z: INFO: Runner: Starting baker thread
2023-07-05T20:38:38.007361291Z: INFO: Runner: Baking thread is not required and will shut down.
2023-07-05T20:38:38.007364386Z: INFO: Runner: Exiting baker thread
```

is confusing since it might look like something is off with baking, but that's not the case.

**Steps to Reproduce**

Start node in protocol 6.

**Versions**
- Software Version: 6.0.{0,1}

td202:
## Purpose

Closes #952 

This prevents the baker thread from starting when it is not needed by the consensus version, and thus avoid the attendant logging. A DEBUG-level log message does note that the baker thread is not required.

## Changes

- Suppress starting the baker thread when it is not required for the current consensus version.
- Make the log message for when the baker thread is shut down more reassuring.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/MultiVersion.hs
+++ b/concordium-consensus/src/Concordium/MultiVersion.hs
@@ -1222,6 +1222,8 @@ startTransactionPurgingThread mvr@MultiVersionRunner{..} =
 
 -- |Start a baker thread associated with a 'MultiVersionRunner'.
 -- This will only succeed if the runner was initialised with baker credentials.
+-- A baker thread will only be started if the current consensus version is 0, since baking
+-- is not handled as a separate thread in consensus version 1.
 startBaker :: MultiVersionRunner finconf -> IO ()
 startBaker MultiVersionRunner{mvBaker = Nothing, ..} =
     mvLog
@@ -1249,14 +1251,7 @@ startBaker mvr@MultiVersionRunner{mvBaker = Just Baker{..}, ..} = whenBakerRequi
             return $! case Vec.last versions of
                 EVersionedConfigurationV0{} -> True
                 EVersionedConfigurationV1{} -> False
-        if required
-            then a
-            else
-                mvLog
-                    Runner
-                    LLDebug
-                    "The current consensus version does not require a baker thread, so it will not \
-                    \be started"
+        when required a
     -- The baker loop takes the current genesis index and last known slot that we baked for, and
     -- will continually attempt to bake until the consensus is shut down.
     bakerLoop :: GenesisIndex -> Slot -> IO ()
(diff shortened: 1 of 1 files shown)


concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -1242,6 +1242,21 @@ startBaker mvr@MultiVersionRunner{mvBaker = Just Baker{..}, ..} = do
...
+                EVersionedConfigurationV0{} -> True
+                EVersionedConfigurationV1{} -> False
+        if required
+            then a
+            else
+                mvLog
+                    Runner
+                    LLDebug
+                    "The current consensus version does not require a baker thread, so it will not \
+                    \be started"
```
abizjak, line 1259: I think this log message isn't really useful, and I think it is confusing, to anybody apart from somebody who knows the code and its history. It's a bit hard to appreciate the distinction that a baker **thread** will not be started, but that baking will still happen, just via a different mechanism.

The code is so simple here that I don't think we need to log not doing the action.

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -1228,7 +1228,7 @@ startBaker MultiVersionRunner{mvBaker = Nothing, ..} =
         Runner
         LLError
         "Attempted to start baker thread, but consensus was started without baker credentials."
-startBaker mvr@MultiVersionRunner{mvBaker = Just Baker{..}, ..} = do
+startBaker mvr@MultiVersionRunner{mvBaker = Just Baker{..}, ..} = whenBakerRequired $ do
```
limemloh, line 1231: Consider adding to the doc comment above that this function is only interesting for older protocol versions and does nothing for newer versions

concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -1228,7 +1228,7 @@ startBaker MultiVersionRunner{mvBaker = Nothing, ..} =
         Runner
         LLError
         "Attempted to start baker thread, but consensus was started without baker credentials."
```
limemloh, line 1230: I might be wrong, but won't this still be logged in the new consensus? The message seems only relevant for the old consensus.

td202: This should never happen because the Rust code only `startBaker` when there are baker credentials.

abizjak (approved):
Looks good.

limemloh (approved):

PR #973, opened 2023-07-14
Fix processing of chain updates in P6.

### LINKED ISSUE #972: Parameter update side-effects may be missed when scheduled simultaneously
**Bug Description**

When updates to certain chain parameters are scheduled for the same time, it is possible for some of the effects to be missed. This can include the following:

- Changes to the time parameters (mint rate and reward period length) may result in the wrong mint rate and length being applied for one reward period.
- Changes to the commission bounds may not be applied to existing bakers.
- Changes to the mint distribution may not be applied correctly in immediate reward distributions.

**Steps to Reproduce**

1. Schedule a mint distribution update, pool parameters update, or time parameters update.
2. Schedule another update (e.g. root key update) for the same time. (Note that not every type of update will mask every other type - root key update should mask all other types.)

**Expected Result**

The side-effects of the update should take effect in the appropriate block.

**Actual Result**

Some of the side effects do not take effect.

**Versions**
- Software Version: 6.0.3 and earlier
- OS: All

abizjak:
## Purpose

Fixes #972

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## 6.0.3
 
 - Update specification hash for protocol 5 to protocol 6 update.
+- Fix processing of chain parameter updates when they occur at the same time.
 
 ## 6.0.2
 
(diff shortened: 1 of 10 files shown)


eb-concordium (approved):

abizjak:
> This is fine for now. However, I would strongly advocate for removing the buggy behaviour altogether, instead of leaving it in place for P1-P5, unless it actually occurs in practice on mainnet or testnet. It would be reasonable to defer this change until after the P6 protocol update.

I agree with that, but I don't think we should risk that before P6 is in effect on all public networks.

We can fix this in 6.1 once we have confirmed it.

PR #986, opened 2023-08-04
Address comments on database changes.

### LINKED ISSUE #984: New Consensus Database Changes: address review comments
This issue is to address review comments on the database changes from @chportma:

1. I cannot make sense of the comment on line 135 of `LowLevel.hs`, namely “The parent block is the (previous) highest certified block.” As far as I can tell, all certified blocks are added, regardless of whether their parent was the previous highest certified block or not. Furthermore, I do not see what “database invariant” (line 132) is related to the comment on line 135.
2. I think that the condition `cbEpoch highCB <= 2 + tcMinEpoch lastTimeout` on line 329 of `StartUp.hs` will always be satisfied: suppose that cbEpoch highCB > 2 + tcMinEpoch lastTimeout, then the epoch of the LFB > 1 + tcMinEpoch lastTimeout (since the difference between the epochs of the highest CB and the LFB can be at most one), so the epoch of the LFB > tcMaxEpoch lastTimeout, hence the timeout points past the LFB, which is not possible for a valid timeout with `tcRound lastTimeout >= curRound`  (line 324). Maybe this test is there to check for some inconsistent in the stored data, but I don’t see why this test would be relevant. In any case, due to point 3 below, this might be removed altogether.
3. In line 335 of `StartUp.hs` a TC is being paired with the current highest CB, but the epoch might not be the one used to sum up the weights of the signatures in the TC. We have stored all CBs, so in particular we will have stored the one that was the highest CB when this TC was generated, but I don’t see any way to find it without more information. Changing the type of the `prsLatestTimeout` to be `RoundTimeout` should be (more than) enough I think.
4. Near the end of `loadCertifiedBlocks` (no screenshot attached), we are adding `prsLastSignedTimeoutMessage` to `currentTimeoutMessages`. But we aren’t adding `prsLastSignedQuorumMessage` to `currentQuorumMessages`. Is there any reason for that? I don’t think that this is an issue, I just don’t see any reasons for adding the tm, but not the qm.

td202:
## Purpose

Closes #984.

## Changes

- `QuorumMessages` structure is revised to index by baker ID instead of finalizer ID, because the latter is not stable across epochs. Consequently, the baker ID is added to the `VerifiedQuorumMessage` structure.
- In `loadCertifiedBlocks`:
    - When loading the latest timeout certificate, we find an appropriate certified block to pair it. Typically, this is the highest certified block, but it can be that that is invalid and we need to use a block from the previous epoch.
    - Load the last sent quorum message into the quorum messages for the current round if it is appropriate.
- Documentation updates.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix a bug in how the last timeout certificate is recovered at start-up.
+
 ## 6.0.4
 
 - Fix a bug in how timeout certificates across epoch boundaries are handled in catch-up.
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
@@ -1034,6 +1034,7 @@ validateBlock blockHash BakerIdentity{..} finInfo = do
                     VerifiedQuorumMessage
                         { vqmMessage = quorumMessage,
                           vqmFinalizerWeight = finalizerWeight finInfo,
+                          vqmFinalizerBakerId = finalizerBakerId finInfo,
                           vqmBlock = block
                         }
                     makeBlock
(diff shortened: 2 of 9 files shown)


chportma (approved):
It looks good!

abizjak (approved):
Good documentation. Looks good to me.

EmilLa1 (approved):

PR #999, opened 2023-08-15
Support for epoch winning bakers and first block queries.

td202:
## Purpose

Support GRPCv2 endpoints `GetWinningBakersEpoch` (for getting the list of bakers that won in a given epoch) and `GetFirstBlockEpoch` (for getting the first block in a given epoch).

## Changes

- Add a new query result code that represents when a query refers to a future epoch.
- Implement getting the first finalized block of an epoch for both consensus versions.
- Implement the `GetWinningBakersEpoch` endpoint (supports consensus v1 only).
- Implement the `GetFirstBlockEpoch` endpoint (supports both consensus versions).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
@@ -356,9 +356,10 @@ getWinningBakersForEpoch targetEpoch sd = do
         bakers <- BS.getCurrentEpochBakers (bpState lastOfEpoch)
         seedState <- BS.getSeedState (bpState lastOfEpoch)
         let leNonce = seedState ^. currentLeadershipElectionNonce
+        let roundWinner = getLeaderFullBakers bakers leNonce
         let roundWB rnd present =
                 WinningBaker
-                    { wbWinner = getLeaderFullBakers bakers leNonce rnd ^. Accounts.bakerIdentity,
+                    { wbWinner = roundWinner rnd ^. Accounts.bakerIdentity,
                       wbRound = rnd,
                       wbPresent = present
                     }
--- a/concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 
@@ -18,6 +19,7 @@ import Concordium.Types
 import qualified Concordium.Types.Accounts as Accounts
 import Concordium.Types.BakerIdentity
 import Concordium.Types.Parameters hiding (getChainParameters)
+import Concordium.Types.Queries
 import Concordium.Utils
 
 import qualified Concordium.Crypto.BlockSignature as Sig
@@ -25,12 +27,13 @@ import Concordium.GlobalState.BakerInfo
 import qualified Concordium.GlobalState.BlockState as BS
 import qualified Concordium.GlobalState.Persistent.BlockState as PBS
 import qualified Concordium.GlobalState.TreeState as GSTypes
+import Concordium.KonsensusV1.LeaderElection
 import Concordium.KonsensusV1.TreeState.Implementation
 import qualified Concordium.KonsensusV1.TreeState.LowLevel as LowLevel
 import Concordium.KonsensusV1.TreeState.Types
 import Concordium.KonsensusV1.Types
 import Concordium.Logger
-import Concordium.Types.SeedState (triggerBlockTime)
+import Concordium.Types.SeedState (currentLeadershipElectionNonce, triggerBlockTime)
 import Concordium.Types.UpdateQueues
 import Concordium.Types.Updates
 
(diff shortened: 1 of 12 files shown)


concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -1207,6 +1207,136 @@ getDelegatorsRewardPeriod bhi maybeBakerId = do
...
+                            | otherwise -> Left EQEInvalidEpoch
+                        Left EmptyEpoch -> Left EQEBlockNotFound
+                        Right block -> Right $! getHash @BlockHash block
+                )
+                ( do
+                    -- Consensus version 1#
+                    (SkovV1.getFirstFinalizedBlockOfEpoch (Left erEpoch) =<< get) <&> \case
+                        Nothing
+                            | isCurrentVersion -> Left EQEFutureEpoch
+                            | otherwise -> Left EQEInvalidEpoch
```
EmilLa1, line 1259: The functions looking up the block for each tree state looks very similar, but yet there is a discrepancy with the result codes, i.e. for consensus version 0 there is a `EQEBlockNotFound`, but not for the latter? By looking at the code for consensus version 1 it looks like the two cases will be treated the same (`EQEFutureEpoch`) ?

EmilLa1: Also, couldnt the two consensus versions actually share the underlying implementation by passing in the handles to the respective treestate stores? 

I guess my question is, why are two separate implementations required for something that is essentially the same except for the treestate store handles (and the bp type).

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -325,3 +328,62 @@ getProtocolUpdateState = do
                     { puQueuedTime = ts,
                       puProtocolUpdate = pu
                     }
+
+-- |Get the list of rounds that fall within a finalized epoch, together with which baker was
+-- eligible to bake in that round and whether a block in that round was included in the finalized
+-- chain.
+-- This returns 'Nothing' if the epoch is not completely finalized (i.e. there is no finalized
+-- block in a higher round).
+getWinningBakersForEpoch ::
```
EmilLa1, line 337: Why is this function being put here as opposed to `Implementation.hs`. The function isn't part of the consensus protocol, it's a query on the provided `SkovData`.

concordium-consensus/src/Concordium/KonsensusV1/Consensus.hs
```diff
@@ -325,3 +328,62 @@ getProtocolUpdateState = do
...
+    -- result.
+    mFirstBlockNextEpoch <- getFirstFinalizedBlockOfEpoch (Left $ targetEpoch + 1) sd
+    forM mFirstBlockNextEpoch $ \startBlock -> do
+        -- The start block will be a finalized block and not the genesis block (its epoch is > 1).
+        -- Its parent will be in the target epoch. (There are no empty epochs.)
+        lastOfEpoch <- parentOfFinalized startBlock
+        bakers <- BS.getCurrentEpochBakers (bpState lastOfEpoch)
+        seedState <- BS.getSeedState (bpState lastOfEpoch)
+        let leNonce = seedState ^. currentLeadershipElectionNonce
+        let roundWB rnd present =
```
EmilLa1, line 359:
```suggestion
        let getWinner = getLeaderFullBakers bakers leNonce
        let roundWB rnd present =
                WinningBaker
                    { wbWinner = getWinner rnd ^. Accounts.bakerIdentity,
                      wbRound = rnd,
                      wbPresent = present
                    }
```

EmilLa1 (approved):
Looking good! 
A few comments and questions.

concordium-consensus/src/Concordium/Skov/Query.hs
```diff
@@ -76,6 +77,46 @@ doGetBlocksAtHeight h = do
             mb <- getFinalizedAtHeight h
             return (toList mb)
 
+-- |Result of querying for the first finalized block in an epoch.
+data EpochResult
```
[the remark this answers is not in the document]
td202: I prefer using `Either`, but I will rename to `EpochFailureResult`.

concordium-consensus/src/Concordium/Queries.hs
td202: The `EQEBlockNotFound` case cannot happen in consensus version 1 because it requires an epoch to be empty, which is (theoretically) possible in consensus version 0.

The implementations do share `interpolationSearchFirstM`, which is the bulk of the look-up. I don't think unifying the implementations beyond this is useful, as there is no common interface over the treestate for the different consensus versions.

abizjak (approved):
As discussed in the office, it's a bit annoying that these two queries take a BlockHashInput (potentially) as argument but do not respond with the block hash that was used for the query.

But since they can also take an epoch number it is not always possible to return a block like it is for other queries that take just the BlockHashInput.

So it's probably the best if it stays like this.

And in the SDKs, we probably want to take only specific block hash, block height, or epoch number as arguments for the winning bakers  to avoid mistakes.

concordium-consensus/src/Concordium/Skov/Query.hs
```diff
@@ -76,6 +77,46 @@ doGetBlocksAtHeight h = do
...
+    TreeStateMonad m =>
+    Either Epoch (BlockPointerType m) ->
+    m (Either EpochFailureResult (BlockPointerType m))
+doGetFirstFinalizedOfEpoch epochOrBlock = do
+    epochLength <- gdEpochLength <$> getGenesisData
+    let blockEpoch b = fromIntegral (blockSlot b `div` epochLength)
+    let targetEpoch = case epochOrBlock of
+            Left epoch -> epoch
+            Right block -> blockEpoch block
+    -- This should be safe to call for any height up to the height of the last finalized block.
```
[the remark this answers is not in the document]
td202: Added some documentation to clarify.

Issue #1000, opened 2023-08-15
Remove concept of pending blocks


eb-concordium:
**Task description**

Remove the concept of pending blocks. This involves removing the pending block table and the pending block queue.

PR #1005, opened 2023-08-18
Remove pending blocks

eb-concordium:
## Purpose

Closes https://github.com/Concordium/concordium-node/issues/1000.

## Changes

- The behaviour of the function `receiveBlockUnknownParent` has changed, so that it now returns `BlockResultPending` without marking or adding the block as pending. It still returns `BlockResultEarly` if if the timestamp is no less than the receive time of the block + the early block threshold.
- The type `PendingBlocks` (consisting of the pending block table and the pending block queue) is removed. As a consequence `SkovData` has no field `_skovPendingBlocks` anymore.
- The type `InMemoryBlockStatus` is removed, and the `_liveMap` in `BlockTable` is now a `HM.HashMap BlockHash (BlockPointer pv)` instead of the former `HM.HashMap BlockHash (InMemoryBlockStatus pv)`.
- The functions `addPendingBlock`, `markPending`, `blockArriveDead`, `purgePending`, `takePendingChildren`, `takeNextPendingUntil`, `processPendingChild` and `processPendingChildren` are removed.
- The `BlockPending` variant of `BlockStatus` is removed.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Fix a bug in how the last timeout certificate is recovered at start-up.
 
+- Remove the concept of pending blocks.
+
 ## 6.0.4
 
 - Fix a bug in how timeout certificates across epoch boundaries are handled in catch-up.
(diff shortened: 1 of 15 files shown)


td202 (approved):
Looks good!

abizjak (approved):
Please reflect the changes in the changelog.

Otherwise looks good.

PR #1020, opened 2023-09-01
Fix #147

### LINKED ISSUE #147: Block state hash reported as "000...0" for genesis blocks
**Bug Description**

The reported block state hash for genesis blocks is `"0000000000000000000000000000000000000000000000000000000000000000"`.
Instead, the correct block state hash should be reported.

**Steps to Reproduce**
Call `GetBlockInfo` for the genesis block.

**Expected Result**
The actual hash of the genesis block state.

**Actual Result**
`"0000000000000000000000000000000000000000000000000000000000000000"`.

**Versions**
- Software Version 1.x

EmilLa1:
## Purpose

Fix a bug described here: https://github.com/Concordium/concordium-node/issues/147

which caused the wrong genesis block state hash to be returned.

## Changes

- For genesis blocks "recalculate" the state hash (as it is not stored together with the genesis configuration) 
  Note that the actual state hash is cached (as a `HashedPersistentBlockState` instance) as the genesis block pointer is cached and this is the pointer we use the state from.
- Remove unused code (basic block pointer and friends)
- Remove `blockStateHash` from `BlockData` type class as it was not required and returned a wrong result for a genesis block.

## Checklist

      hard-to-understand areas.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased changes
+- Fix for https://github.com/Concordium/concordium-node/issues/147 so the correct state hash is now returned.
```
abizjak, line 4: We don't generally link issues like this. It's better to describe the bug directly.

concordium-consensus/src/Concordium/GlobalState/Block.hs
```diff
@@ -343,7 +343,8 @@ instance forall pv. (IsProtocolVersion pv) => BlockData (Block pv) where
             STOV1 -> emptyTransactionOutcomesHashV1
     blockTransactionOutcomesHash (NormalBlock bb) = blockTransactionOutcomesHash bb
 
-    -- FIXME: replace stub, and move into gendata
+    -- Note. Do not rely on this for type class for retrieving the block state hash of a
+    -- genesis block. Instead use @bpBlockStateHash@ of 'BlockPointerData'.
     blockStateHash GenesisBlock{} = StateHashV0 minBound
     blockStateHash (NormalBlock bb) = blockStateHash bb
```
abizjak, line 350: Do we need this method then at all?

abizjak: We don't usually solve problems like this, by having a function we write to not use.

limemloh (approved):
Nice work :fire: No comments from here

concordium-consensus/src/Concordium/GlobalState/Block.hs
EmilLa1: Actually, I think we can remove it without causing too much disturbance.

td202 (approved):

abizjak (approved):

Issue #989, opened 2023-08-08
Node collector does not work when TLS is enabled on the node


abizjak:
**Bug Description**

If the node is configured with TLS for its GRPC interface then the node collector is not able to collect data from it.

**Versions**
- Software Version: all

EmilLa1:
Options I see for this.
1) Add localhost to SANs - however this comes with its own fair share of consequences and should be avoided. Moreover no CA's issue certificates with localhost so the certificate in that case must be a self-signed certificate which in turn would be very troublesome and annoying for "real" consumers of the grpc api. 
2) Add an extra http listener for the grpc api (with only the required endpoints exposed for the collector in order to function), and then the collector should always use that endpoint. 
3) Remove TLS configuration all together on the node and let node operators handle secure connections.
4) Collector does simply not work if tls is enabled on the node. I.e. stay as is. 
The collecting architecture is already flawed as is and there has been some talk about phasing the collector and regular dashboard out in favor of just ccdscan. 


I propose option 2 as that would integrate well into the **current** setup, but if we are to phase out the collector and dashboard anyhow anytime soon, I would vote for option 4.
Option 3 I see as the ideal solution, but it's ramifications are unclear as we do not know how many rely on tls in the node (or if operators are having a tls termination proxy in front of the node anyhow).
Option 1 is not really an option, just a thought.

Thoughts @td202 @abizjak ?

abizjak:
1. That is not a good option.
2. This is hacky. Either we should just have http enabled for all endpoints or not, we should not hardcode some arbitrary set that happens to be used.
3. Is a no-go. TLS on the node is used and simplifies the setup quite a bit.
4. CCDScan uses the collector, so they are not exclusive.


We can do the following. Make the collector understand TLS (based on the schema) (currently it does not, it only supports http). Then we can document that if you enable TLS on the node you need to point the collector to the public URL of the node, and have that accessible.

Long term we will migrate away from the separate collector if we want to make it robust anyhow.

EmilLa1:
Ok. We'll go with option 5)
If node is configured with TLS, then one has to configure the collector to use the url configured for the node, and hence it would have to go through proper dns resolving. 

I'll add this possibility to the collector and document it. Is there any external documentation that must be updated?

PR #1021, opened 2023-09-04
Allow the collector to consume data when node is configured with tls.

EmilLa1:
## Purpose
Fixes https://github.com/Concordium/concordium-node/issues/989

## Changes

If the `grpc_host` (the node) is configured for using TLS, then the collector now sets up a tls connection based on the system trust store.

## Checklist

      hard-to-understand areas.


collector/src/bin/collector.rs
```diff
@@ -196,7 +196,18 @@ async fn collect_data<'a>(
...
+    // Set up a tls client if required using the locally installed trust store.
+    let endpoint = Channel::from_shared(grpc_host)?.timeout(Duration::from_secs(grpc_timeout));
+    let channel = if endpoint.uri().scheme().map_or(false, |x| x == &http::uri::Scheme::HTTPS) {
+        endpoint.tls_config(ClientTlsConfig::new())?
+    } else {
+        endpoint
+    };
+
+    //    let channel =
+    // Channel::from_shared(grpc_host)?.timeout(Duration::from_secs(grpc_timeout));
```
abizjak, line 209: What is this?

collector/src/bin/collector.rs
```diff
@@ -196,7 +196,18 @@ async fn collect_data<'a>(
 
     // Setting up Client
     let grpc_timeout = conf.grpc_timeout;
-    let channel = Channel::from_shared(grpc_host)?.timeout(Duration::from_secs(grpc_timeout));
+
+    // Set up a tls client if required using the locally installed trust store.
+    let endpoint = Channel::from_shared(grpc_host)?.timeout(Duration::from_secs(grpc_timeout));
+    let channel = if endpoint.uri().scheme().map_or(false, |x| x == &http::uri::Scheme::HTTPS) {
+        endpoint.tls_config(ClientTlsConfig::new())?
+    } else {
```
abizjak, line 204: I am not sure why this works. By default no TLS roots are enabled. tonic has a feature https://docs.rs/tonic/latest/tonic/#feature-flags `tls-roots` which is not enabled by default.

Perhaps it's enabled by another dependency? But we should be explicit about this regardless.

collector/src/bin/collector.rs
EmilLa1: No idea. It's gone now :)

collector/src/bin/collector.rs
EmilLa1: Explicit now.

collector/src/bin/collector.rs
```diff
@@ -196,7 +196,15 @@ async fn collect_data<'a>(
 
     // Setting up Client
     let grpc_timeout = conf.grpc_timeout;
-    let channel = Channel::from_shared(grpc_host)?.timeout(Duration::from_secs(grpc_timeout));
+
+    // Set up a tls client if required using the locally installed trust store.
+    let endpoint = Channel::from_shared(grpc_host)?.timeout(Duration::from_secs(grpc_timeout));
+    let channel = if endpoint.uri().scheme().map_or(false, |x| x == &http::uri::Scheme::HTTPS) {
```
td202, line 202:
```suggestion
    let channel = if endpoint.uri().scheme() == Some(&http::uri::Scheme::HTTPS) {
```

collector/README.md
```diff
@@ -9,3 +9,6 @@ Assuming you have a node running locally with GRPC V2 available at `127.0.0.1:20
 ```console
 $> cargo run -- --collector-url http://127.0.0.1:8080/post/nodes --grpc-host http://127.0.0.1:20000  --node-name "my-node-name"
 ```
+
+### TLS
+If the node is configured to be using TLS connections then one has to configure the collector (the `--grpc-host` flag) to be using the URL of the domain of which the certificate was issued for.
```
td202, line 14:
```suggestion
If the node is configured to use TLS connections then the collector must be configured (with the `--grpc-host` flag) to use the URL with the domain for which the certificate was issued.
```

collector/src/bin/collector.rs
abizjak: So this is resolved, but the code is unchanged. What is the meaning of this?

abizjak (approved):
The code looks OK assuming Thomas' comment is addressed.

I assume you tested this right? Just to make sure. You can easily test it on the service node on stagenet which is configured with TLS.

collector/src/bin/collector.rs
EmilLa1: Probably clicked the wrong button by accident

EmilLa1:
> The code looks OK assuming Thomas' comment is addressed.
> 
> I assume you tested this right? Just to make sure. You can easily test it on the service node on stagenet which is configured with TLS.

Yes, I checked that it works with and without TLS.

limemloh (approved):

td202 (approved):

PR #1041, opened 2023-09-28
New account map

EmilLa1:
## Purpose

This feature introduces a new persistent store for the account map, i.e. a helper structure used for looking up accounts via their addresses. 
The introduced store is simply a mapping `AccountAddress -> AccountIndex` and it supersedes the `PersistentAccountMap` trie structure.

The new account map functions across protocol versions and hence resides in its own lmdb database. 
This also means that the node no longer stores a `PersistentAccountMap` for each block but instead simply uses the new account map thus reducing the rate at which the database grows.

## Changes

- A new `AccountMapStoreMonad` structure is introduced for interacting with the new lmdb database.
- A `DifferenceMap` is introduced to handle accounts created in blocks that are not yet persisted.
- For an introduction to control flow please refer to the documentation [here](https://github.com/Concordium/concordium-node/blob/15ed9b30e33b618c72b85e3f9ba408d0d03dea20/concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs#L15)

- The feature is protocol independent and hence it is implemented for both consensus versions.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +122,111 @@ import Concordium.Types.HashableTo
...
-            accts
-                { accountMap = acctMap,
-                  accountTable = acctTable
-                }
+            accts{accountTable = acctTable}
 
--- | An 'Accounts' with no accounts.
-emptyAccounts :: Accounts pv
-emptyAccounts = Accounts AccountMap.empty L.empty Trie.empty
+emptyAccounts :: Maybe DiffMap.DifferenceMap -> Accounts pv
```
limemloh, line 203: **Suggestion (non-blocking):** Consider removing the parameter.

I think this is only ever called with `Nothing` as the first argument.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +122,111 @@ import Concordium.Types.HashableTo
...
+emptyAccounts Nothing = Accounts L.empty Trie.empty Nothing
```
limemloh, line 204: **Suggestion (non-blocking):** Documentation would be nice here

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,64 @@
...
+--  a block identified by a 'BlockHash' and its associated 'BlockHeight'.
+--  The difference map only contains accounds that was added since the '_dmParentMap'.
+data DifferenceMap = DifferenceMap
+    { -- | Accounts added in a block.
+      dmAccounts :: !(Map.Map AccountAddress AccountIndex),
+      -- | Parent map of non-finalized blocks.
+      --  In other words, if the parent block is finalized,
+      --  then the parent map is @Notnhing@ as the LMDB account map
+      --  should be consulted instead.
+      dmParentMap :: !(Maybe DifferenceMap)
```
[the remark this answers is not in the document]
EmilLa1: The `storeUpdate` clears out the difference map when a block is being persisted (certified or finalized for ConcordiumBFT).

But you lead me on to track down a bug! Thanks!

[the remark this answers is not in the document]
limemloh: Right, so I guess the same issue leads to something always referencing these resulting in a memory leak.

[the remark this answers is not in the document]
abizjak: It didn't do that though. It only reset the value that was for the block that was being written, but descendant blocks have already copied that value.

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,74 @@
...
+--  a block identified by a 'BlockHash' and its associated 'BlockHeight'.
+--  The difference map only contains accounds that was added since the '_dmParentMap'.
+data DifferenceMap = DifferenceMap
+    { -- | Accounts added in a block.
+      dmAccounts :: !(Map.Map AccountAddress AccountIndex),
+      -- | Parent map of non-finalized blocks.
+      --  In other words, if the parent block is finalized,
+      --  then the parent map is @Notnhing@ as the LMDB account map
+      --  should be consulted instead.
+      dmParentMap :: !(IORef (Maybe DifferenceMap))
```
abizjak, line 25: This should explain why IORef is used.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +123,123 @@ import Concordium.Types.HashableTo
...
     storeUpdate Accounts{..} = do
-        (pMap, accountMap') <- storeUpdate accountMap
         (pTable, accountTable') <- storeUpdate accountTable
         (pRegIdHistory, regIdHistory') <- storeUpdate accountRegIdHistory
+        mAccDiffMap <- liftIO $ readIORef accountDiffMap
+        case mAccDiffMap of
+            Nothing -> return ()
+            Just accDiffMap -> do
+                flattenedAccounts <- liftIO $ DiffMap.flatten accDiffMap
+                LMDBAccountMap.insert flattenedAccounts
```
limemloh, line 171: **Suggestion (non-blocking):** This side effect adds the requirement that `storeUpdate` must only be called when finalizing a block, which is breaking (my understanding of) the `BlobStorable` abstraction. I would prefer to separate this from `storeUpdate`.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +123,123 @@ import Concordium.Types.HashableTo
...
+                -- The difference map is set to @Nothing@ as any potential new accounts have been written to the
+                -- lmdb account map.
+                liftIO $ modifyIORef' accountDiffMap (const Nothing)
```
abizjak, line 174: Don't you need `atomicWriteIORef` instead? This variable can potentially be accessed by concurrent readers, can it not?

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
abizjak: I agree with that. This is quite a hacky way of storing these accounts.

Would it not make more sense to do this as part of either `saveBlockState`  or I think probably one level higher is cleaner, in `markFinalized` for V0 consensus and `makeStoredBlock` in V1 consensus.

abizjak: You could return a list of accounts to store from `saveBlockState` to avoid traversing things twice.

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,68 @@
...
+lookup :: AccountAddress -> DifferenceMap -> Maybe AccountIndex
+lookup addr = check
+  where
+    check DifferenceMap{..} = case Map.lookupGE k dmAccounts of
+        Nothing -> check =<< dmParentMap
+        Just (foundAccAddr, accIdx) ->
+            if checkEquivalence foundAccAddr
+                then Just accIdx
+                else Nothing
+    k = createAlias addr 0
```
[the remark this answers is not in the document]
EmilLa1: Yes I believe it's needed. Since we're doing a prefix lookup we need to be sure that we don't get accidentally look past the account. The account address we're looking up via might just be largest allowed alias for the account, but we've stored the address in canonical form (hash of the credential registration id).

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
EmilLa1: Agree. I chose the former, i.e. it's being written to the lmdb store as part of `saveBlockState`.

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,68 @@
...
+    checkEquivalence found = accountAddressEmbed k == accountAddressEmbed found
+
+-- | Insert an account into the difference map.
+--  Note that it is up to the caller to ensure only the canonical 'AccountAddress' is added.
```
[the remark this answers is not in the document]
EmilLa1: No, I mean the former. I cleaned up the lookup as per suggestions above and it should now be clear.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -169,47 +237,53 @@ getAccountByCredId cid accs@Accounts{..} =
         Just ai -> fmap (ai,) <$> indexedAccount ai accs
 
 -- | Get the account at a given index (if any).
-getAccountIndex :: (IsProtocolVersion pv, MonadBlobStore m) => AccountAddress -> Accounts pv -> m (Maybe AccountIndex)
-getAccountIndex addr Accounts{..} = AccountMap.lookup addr accountMap
+getAccountIndex :: (SupportsPersistentAccount pv m) => AccountAddress -> Accounts pv -> m (Maybe AccountIndex)
```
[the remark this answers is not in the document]
EmilLa1: Added a few words.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -287,14 +401,13 @@ migrateAccounts ::
...
     -- that is purely in-memory and just copied over.
     newAccountRegIds <- Trie.migrateUnbufferedTrieN return accountRegIdHistory
     return $!
         Accounts
-            { accountMap = newAccountMap,
-              accountTable = newAccountTable,
-              accountRegIdHistory = newAccountRegIds
+            { accountTable = newAccountTable,
+              accountRegIdHistory = newAccountRegIds,
+              accountDiffMap = Nothing
```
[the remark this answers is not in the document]
EmilLa1: Yes it'll be empty so it's actually fine to pass it in. Reusing the reference.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
```diff
@@ -309,24 +309,30 @@ destroyBlobStore bs@BlobStore{..} = do
...
+            (bscLoadCallback, bscStoreCallback) <- mkCallbacksFromBlobStore bscBlobStore
+            return BlobStore{..}
+        res <- runBlobStoreT a bs
+        liftIO $ do
+            let BlobStore{..} = bs
+                BlobStoreAccess mv _ mmap = bscBlobStore
+            _ <- takeMVar mv
+            writeIORef mmap BS.empty
+            freeCallbacks bscLoadCallback bscStoreCallback
+            return res
```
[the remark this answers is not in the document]
EmilLa1: It's because the type of `runBlobStoreTemp` changed a bit, so this makes the code cleaner as opposed to even more `liftIO`s.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3311,24 +3322,33 @@ instance Cache.HasCache Modules.ModuleCache (PersistentBlockStateContext pv) whe
...
 
--- | Create a new account cache of the specified size for running the given monadic operation by
+-- | Create a new account cache of the specified size and a temporary 'LMDBAccountMap' for running the given monadic operation by
 --  extending the 'BlobStore' context to a 'PersistentBlockStateContext'.
-withNewAccountCache :: (MonadIO m) => Int -> BlobStoreT (PersistentBlockStateContext pv) m a -> BlobStoreT BlobStore m a
-withNewAccountCache size bsm = do
-    ac <- liftIO $ newAccountCache size
-    mc <- liftIO $ Modules.newModuleCache 100
-    alterBlobStoreT (\bs -> PersistentBlockStateContext bs ac mc) bsm
+-- Note. this function should only be used for tests.
```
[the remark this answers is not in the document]
EmilLa1: Because it adds a warning :D 

Thing is, it's used from the `KonsensusV1/TestMonad` so it'll produce a warning on each compilation. And I don't want to move stuff around too much in this PR.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -2231,10 +2237,10 @@ doAccountList pbs = do
     bsp <- loadPBS pbs
     Accounts.accountAddresses (bspAccounts bsp)
 
-doAddressWouldClash :: (SupportsPersistentState pv m) => PersistentBlockState pv -> AccountAddress -> m Bool
-doAddressWouldClash pbs addr = do
+doGetAccountListHistorical :: (SupportsPersistentState pv m) => PersistentBlockState pv -> m [AccountAddress]
```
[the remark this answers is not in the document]
EmilLa1: Added some context.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -424,7 +439,8 @@ loadSkovPersistentData rp _treeStateDirectory pbsc = do
             genStoredBlock <-
                 maybe (logExceptionAndThrowTS GenesisBlockNotInDataBaseError) return
                     =<< liftIO (getFirstBlock _db)
-            _genesisBlockPointer <- liftIO $ makeBlockPointer genStoredBlock
+
+            _genesisBlockPointer <- makeBlockPointer genStoredBlock
```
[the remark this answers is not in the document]
EmilLa1: Types changed because of a `MonadLogger` constraint.

concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -352,17 +366,17 @@ liftSkovQueryBHIAndVersion av0 av1 bhi = do
                 liftSkovQueryLatest
                     ( do
                         -- consensus version 0
-                        bp <- case other of
-                            Best -> bestBlock
-                            LastFinal -> lastFinalizedBlock
-                        (bpHash bp,) . Just <$> av0 evc bp
+                        (bp, blockContext) <- case other of
+                            Best -> (,BCBest) <$> bestBlock
```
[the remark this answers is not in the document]
EmilLa1: ... And it will return all the accounts that existed at the end of that (now former) block. Which was the best block when the query started and so the `BHIQueryResponse` will yield the hash of that now former best block. 

Or do you mean that this logic is faulty as it should do something different?

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
EmilLa1: Yes I need that.

concordium-consensus/src/Concordium/GlobalState/AccountMap/LMDB.hs
```diff
@@ -0,0 +1,294 @@
...
+        _accountMapStore <-
+            AccountMapStore
+                <$> mdb_dbi_open'
+                    txn
+                    (Just accountMapStoreName)
+                    [MDB_CREATE | not readOnly]
+        return DatabaseHandlers{..}
+
+-- | Initialize database handlers in ReadWrite mode.
+--  This simply loads the references and does not initialize the databases.
```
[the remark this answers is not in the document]
EmilLa1: elaborated

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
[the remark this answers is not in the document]
abizjak: For things that are only for tests we often have warnings on the actual function, and disable those warnings in the test file only.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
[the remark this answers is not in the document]
abizjak: Ok.

concordium-consensus/src/Concordium/Queries.hs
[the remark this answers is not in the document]
abizjak: This
```haskell
-- | Get a list of all account addresses and their associated account indices.
--  There are no guarantees of the order of the list. This is because the resulting list is potentially
--  a concatenation of two lists of account addresses.
allAccounts :: (SupportsPersistentAccount pv m) => Accounts pv -> m [(AccountAddress, AccountIndex)]
allAccounts accounts = do
    persistedAccs <- LMDBAccountMap.all
    mDiffMap <- liftIO $ readIORef (accountDiffMap accounts)
    case mDiffMap of
        Nothing -> return persistedAccs
        Just accDiffMap -> do
            flattenedDiffMapAccounts <- DiffMap.flatten accDiffMap
            return $! persistedAccs <> flattenedDiffMapAccounts
```

will always return at least all the accounts that are in LMDB. Regardless of what block you give it.

And this is ultimately called. So you're claiming to return from the best block, and this is the hash you are returning in the header, but the list of accounts is from some other block.

concordium-consensus/src/Concordium/GlobalState/AccountMap/LMDB.hs
```diff
@@ -0,0 +1,294 @@
...
+        -- The key to use for looking up an account.
+        -- We do a prefix lookup on the first 29 bytes of the account address as
+        -- the last 3 bytes are reserved for aliases.
+        accLookupKey = BS.take prefixAccountAddressSize $ FBS.toByteString accAddr
+        checkEquivalence x y = accountAddressEmbed x == accountAddressEmbed y
+
+    all = asReadTransaction $ \dbh txn -> loadAll txn (dbh ^. accountMapStore)
+
+    isInitialized = do
+        size <- asReadTransaction $ \dbh txn -> databaseSize txn (dbh ^. accountMapStore)
```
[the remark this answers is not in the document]
EmilLa1: 0, as it has no entries.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +122,138 @@ import Concordium.Types.HashableTo
...
+                      ..
                     }
-        return (pMap >> pTable >> pRegIdHistory, newAccounts)
+
+        -- put an empty 'OldMap.PersistentAccountMap'.
+        -- In earlier versions of the node the above mentioned account map was used,
+        -- but this is now superseded by the 'LMDBAccountMap.MonadAccountMapStore'.
+        -- We put this (0 :: Int) here to remain backwards compatible as this simply indicates an empty map.
+        -- This should be revised as part of a future protocol update when the database layout can be changed.
+        return (put (0 :: Word64) >> pTable >> pRegIdHistory, newAccounts)
```
td202, line 192: Why not just `put OldMap.empty`?

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3673,6 +3696,26 @@ migrateBlockPointers migration BlockStatePointers{..} = do
...
+--
+--  Further the 'DiffMap.DifferenceMap' of the accounts structure in the provided block state is
+--  "bumped" in the sense that a new one is created for the new thawed block with a pointer to the parent difference map.
+--  The parent difference map is empty if the parent is persisted otherwise it may contain new accounts created in that block.
+doThawBlockState ::
+    (SupportsPersistentState pv m) =>
+    HashedPersistentBlockState pv ->
+    m (PersistentBlockState pv)
+doThawBlockState HashedPersistentBlockState{..} = do
+    -- This load is cheap as the underlying block state is retained in memory as we're building from it, so it must be the "best" block.
```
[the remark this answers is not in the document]
abizjak: The comment about invokeInstance is irrelevant. The whole state is not thawed there, my bad, only the contract state is.

However the other point about this being an odd place remains. In particular this is changing the preconditions of this method. This is a method on the BlockStateStorage class, which as of now does not say anything about best block.

It's fine to refine that, but please state these additional conditions in the typeclass definition, rather than inside the implementaion of the instance.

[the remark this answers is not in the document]
EmilLa1: I've added it as a precondition on the type class definition.

concordium-consensus/src/Concordium/GlobalState/AccountMap/LMDB.hs
```diff
@@ -0,0 +1,269 @@
...
+--      * All accounts in the store are in persisted blocks (finalized or certified).
+class (Monad m) => MonadAccountMapStore m where
+    -- | Inserts the accounts to the underlying store.
+    insertAccount :: [(AccountAddress, AccountIndex)] -> m ()
+
+    -- | Looks up the ‘AccountIndex’ for the provided ‘AccountAddress’ by using the
+    --  equivalence class 'AccountAddressEq'.
+    --  Returns @Just AccountIndex@ if the account is present in the ‘AccountMap’
+    --  and returns @Nothing@ if the account was not present.
+    lookupAccountIndex :: AccountAddress -> m (Maybe AccountIndex)
```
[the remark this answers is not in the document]
EmilLa1: Clarified the documentation.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -169,34 +264,43 @@ getAccountByCredId cid accs@Accounts{..} =
         Just ai -> fmap (ai,) <$> indexedAccount ai accs
 
 -- | Get the account at a given index (if any).
-getAccountIndex :: (IsProtocolVersion pv, MonadBlobStore m) => AccountAddress -> Accounts pv -> m (Maybe AccountIndex)
-getAccountIndex addr Accounts{..} = AccountMap.lookup addr accountMap
+--  Note that this is looking up via the account alias mechanism introduced in protocol version 3 for all protocol versions.
+--  This is fine as there are no clashes and this approach simplifies the implementation.
```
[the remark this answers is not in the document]
EmilLa1: `lookupAccountIndex` will based on the protocol version lookup via an exact match (i.e. when aliases are not supported) and when aliases are supported, then it will lookup via the equivalence class.

I updated documentation for `lookupAccountIndex` to make this clear to any implementer.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
EmilLa1: It didn't have a `Serialize` instance and I didn't want to add anything to it now it's "deprecated". 

However, I'm `storeUpdate`ing on the empty `OldMap` now, so there's no magic constants.

concordium-consensus/src/Concordium/Queries.hs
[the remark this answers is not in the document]
EmilLa1: I changed this.

concordium-consensus/src/Concordium/GlobalState/AccountMap/LMDB.hs
```diff
@@ -0,0 +1,269 @@
...
+newtype AccountMapStoreMonad (m :: Type -> Type) (a :: Type) = AccountMapStoreMonad {runAccountMapStoreMonad :: m a}
+    deriving (Functor, Applicative, Monad, MonadIO, MonadThrow, MonadCatch, MonadLogger, MonadReader r, MonadState s, TimeMonad) via m
+    deriving (MonadTrans) via IdentityT
+
+deriving instance (MonadProtocolVersion m) => MonadProtocolVersion (AccountMapStoreMonad m)
+
+-- | Delete the provided accounts from the LMDB store.
+--
+--  This function should only be used when rolling back certified blocks. When rolling back finalized blocks,
+--  no accounts should be deleted as they are already confirmed to be finalized.
```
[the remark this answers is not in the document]
EmilLa1: No longer applicable, revised this. We no longer roll back the account map. 
Instead, when looking up accounts via the map, we perform a bounds check with respect to the size of the account table.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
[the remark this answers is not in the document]
EmilLa1: I changed this, so it preserves semantics for older protocol versions.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -256,24 +373,63 @@ updateAccountsAtIndex' fupd ai = fmap snd . updateAccountsAtIndex fupd' ai
...
+            -- We fold in ascending order of the @AccountIndex@
+            -- so we @zip@ it correctly when returning @[(AccountAddress, AccountIndex)]@
+            foldAccounts
+                ( \(!accum) pacc -> do
+                    !addr <- accountCanonicalAddress pacc
+                    return $ addr : accum
+                )
+                []
+                accts
+        return $ zip addresses [0 ..]
```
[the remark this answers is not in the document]
td202: My testing on stagenet confirms the current behaviour is wrong.

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,139 @@
...
+import Prelude hiding (lookup)
+
+import Concordium.Types
+import Concordium.Types.Option (Option (..))
+
+-- | A mutable reference to a 'DiffMap.DifferenceMap'.
+type DifferenceMapReference = IORef (Option DifferenceMap)
+
+-- | The empty reference
+emptyReference :: (MonadIO m) => m (IORef (Option DifferenceMap))
```
td202, line 50:
```suggestion
emptyReference :: (MonadIO m) => m DifferenceMapReference
```

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,139 @@
...
+emptyReference = liftIO $ newIORef Absent
```
td202, line 51: I think it might be worth calling this `newEmptyReference`. There could also be a `finalizedEmptyReference`, which is `unsafePerformIO emptyReference` (with a no-inline pragma). This would serve as a common empty reference that can be used when it is guaranteed that the contents of the reference will never be changed.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +114,139 @@ import Concordium.Types.HashableTo
...
+            accts{accountTable = acctTable}
 
--- | An 'Accounts' with no accounts.
-emptyAccounts :: Accounts pv
-emptyAccounts = Accounts AccountMap.empty L.empty Trie.empty
+-- | Create a new empty 'Accounts' structure with a pointer
+--  to an empty 'DiffMap.DifferenceMap'.
+emptyAccounts :: (MonadIO m) => m (Accounts pv)
+emptyAccounts = do
+    accountDiffMapRef <- liftIO DiffMap.emptyReference
```
abizjak, line 214:
```suggestion
    accountDiffMapRef <- DiffMap.emptyReference
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +114,139 @@ import Concordium.Types.HashableTo
...
-                { accountMap = acctMap,
-                  accountTable = acctTable
-                }
+            accts{accountTable = acctTable}
 
--- | An 'Accounts' with no accounts.
-emptyAccounts :: Accounts pv
-emptyAccounts = Accounts AccountMap.empty L.empty Trie.empty
+-- | Create a new empty 'Accounts' structure with a pointer
+--  to an empty 'DiffMap.DifferenceMap'.
```
abizjak, line 211:
```suggestion
-- | Create a new empty 'Accounts' structure.
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -168,35 +256,60 @@ getAccountByCredId cid accs@Accounts{..} =
...
+                else
+                    DiffMap.lookupExact addr accDiffMap >>= \case
+                        Just accIdx -> return $ Just accIdx
+                        Nothing -> lookupDisk
+  where
+    -- Lookup the 'AccountIndex' in the lmdb backed account map,
+    -- and make sure it's within the bounds of the account table.
+    -- We do the bounds check as it could be that the lmdb backed account map
+    -- yields accounts which are not yet present in the @accountTable@.
+    -- In particular this can be the case if finalized blocks has been rolled
```
abizjak, line 285:
```suggestion
    -- In particular this can be the case if finalized blocks have been rolled
```

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3553,6 +3571,19 @@ instance (IsProtocolVersion pv, PersistentState av pv r m) => BlockStateStorage
         flushStore
         return ref
 
+    saveAccounts HashedPersistentBlockState{..} = do
+        -- this load should be cheap as the blockstate is in memory.
+        accs <- bspAccounts <$> loadPBS hpbsPointers
+        -- write the accounts that was created in the block and
```
abizjak, line 3577:
```suggestion
        -- write the accounts that were created in the block and
```

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3673,6 +3705,31 @@ migrateBlockPointers migration BlockStatePointers{..} = do
...
+    (SupportsPersistentState pv m) =>
+    HashedPersistentBlockState pv ->
+    m (PersistentBlockState pv)
+doThawBlockState HashedPersistentBlockState{..} = do
+    -- This load is cheap as the underlying block state is retained in memory as we're building from it, so it must be the "best" block.
+    bsp@BlockStatePointers{bspAccounts = a0@Accounts.Accounts{..}} <- loadPBS hpbsPointers
+    mDiffMap <- liftIO $ readIORef accountDiffMapRef
+    newDiffMapRef <- case mDiffMap of
+        -- reuse the reference pointing to @Nothing@.
+        Absent -> return accountDiffMapRef
```
abizjak, line 3725: This looks incorrect to me if we derive more than one block from the same parent. Won't this cause issues if those blocks have new accounts?

concordium-consensus/src/Concordium/KonsensusV1/SkovMonad.hs
```diff
@@ -661,20 +708,26 @@ migrateSkovV1 ::
...
         initGS :: InitMonad pv (SkovData pv)
         initGS = do
-            stateRef <- saveBlockState newInitialBlockState
-            chainParams <- getChainParameters newInitialBlockState
-            genEpochBakers <- genesisEpochBakers newInitialBlockState
-            let genMeta = regenesisMetadata (getHash newInitialBlockState) regenesis
+            newState <- newInitialBlockState
+            stateRef <- do
+                ref <- saveBlockState newState
+                saveAccounts newState
```
abizjak, line 722: Why is this necessary? If this is migration from the old consensus aren't all accounts already stored?

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/TransactionProcessingTest.hs
```diff
@@ -128,6 +131,15 @@ instance (MonadReader r m) => MonadReader r (FixedTimeT m) where
...
     local f (FixedTime k) = FixedTime $ local f . k
 
+newtype NoLoggerT m a = NoLoggerT {runNoLoggerT :: m a}
+    deriving (Functor, Applicative, Monad, MonadIO, MonadReader r, MonadFail)
+
+instance (Monad m) => MonadLogger (NoLoggerT m) where
+    logEvent _ _ _ = return ()
+
+deriving instance (TimeMonad m) => TimeMonad (NoLoggerT m)
+deriving instance (MonadState s m) => MonadState s (NoLoggerT m)
```
abizjak, line 141: Why can't you use newtype deriving for this? Why does it need to be this more verbose standalone?

abizjak (commented):
I would like to see simple unit tests at least for the issues that have been pointed out in review.

There are a lot of moving pieces.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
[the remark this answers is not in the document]
EmilLa1: Hm yea. Originally I was using `foldAccountsDesc`, but can see some refactoring has gone wrong. I'll add a test for this as well.

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
EmilLa1: I've renamed to `newEmptyReference`. Not sure if it's worth it to add the `finalizedEmptyReference`, there will only ever be 1 alive right? So in practice there will be two empty references (as opposed to 1 qua your suggestion) will be from a (child to a block A) is finalized and until A is GC'ed. @td202 @abizjak  This seems like a premature optimization. But it can be that I'm wrong and in that case I am happy to revise this. 

Resolving for now.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3673,6 +3705,31 @@ migrateBlockPointers migration BlockStatePointers{..} = do
...
+        Present _ -> do
+            -- create a new reference pointing to
+            -- a new difference map which inherits the parent difference map.
+            liftIO $ newIORef $ Present (DiffMap.empty accountDiffMapRef)
```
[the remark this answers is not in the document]
EmilLa1: Yes I believe you are correct. I have fixed the code as suggested and I will add some tests.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased changes
+- The account map is now kept solely on disk in a separate lmdb database and it is no longer part of the internal block state database.
```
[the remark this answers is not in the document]
EmilLa1: merged

concordium-consensus/src/Concordium/GlobalState/AccountMap/LMDB.hs
```diff
@@ -0,0 +1,275 @@
...
+-- | Close the database. The database should not be used after it is closed.
+closeDatabase :: DatabaseHandlers -> IO ()
+closeDatabase dbHandlers = runInBoundThread $ mdb_env_close $ dbHandlers ^. dbhStoreEnv . seEnv
+
+-- ** Monad implementation
+
+-- | The 'AccountMapStoreMonad' for interacting with the LMDB database.
+newtype AccountMapStoreMonad (m :: Type -> Type) (a :: Type) = AccountMapStoreMonad {runAccountMapStoreMonad :: m a}
+    deriving (Functor, Applicative, Monad, MonadIO, MonadThrow, MonadCatch, MonadLogger, MonadReader r, MonadState s, TimeMonad) via m
+    deriving (MonadTrans) via IdentityT
```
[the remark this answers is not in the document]
EmilLa1: We do not need it. Fixed.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1352,6 +1353,13 @@ class (BlockStateOperations m, FixedSizeSerialization (BlockStateRef m)) => Bloc
     --  changes to it must not affect 'BlockState', but an efficient
     --  implementation should expect that only a small subset of the state will
     --  change, and thus a variant of copy-on-write should be used.
+    --
+    --  The caller of this function should adhere to the following precondition:
+    --  * This function must only be called on the best block or a block that is already
+    --    retained in memory.
+    --    This function loads the provided blockstate into memory (which is fine for the
+    --    best block, as it is already in memory) but it should be avoided for blocks
+    --    that are not already in memory.
```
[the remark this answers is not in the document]
EmilLa1: It's not .

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
EmilLa1: Yes, this is incorrect. It's fixed. ( I still have to add a few tests.)

concordium-consensus/src/Concordium/KonsensusV1/SkovMonad.hs
EmilLa1: Yes, this shouldn't be necessary as you correctly point out the account map functions across protocol (and consensus versions).

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +114,139 @@ import Concordium.Types.HashableTo
...
+        Present accountsCreated -> do
+            listOfAccountsCreated <- liftIO $ DiffMap.flatten accountsCreated
+            liftIO $ atomicWriteIORef accountDiffMapRef Absent
+            LMDBAccountMap.insertAccounts listOfAccountsCreated
+
+-- | Create and set the 'DiffMap.DifferenceMap' for the provided @Accounts pv@.
+--  Precondition: The provided @IORef (Option (DiffMap.DifferenceMap))@ MUST correspond to the parent map.
+reconstructDifferenceMap :: (SupportsPersistentAccount pv m) => DiffMap.DifferenceMapReference -> [AccountAddress] -> Accounts pv -> m DiffMap.DifferenceMapReference
+reconstructDifferenceMap parentRef listOfAccounts Accounts{..} = do
+    let diffMap' = DiffMap.fromList parentRef $ zip listOfAccounts $ map AccountIndex [L.size accountTable + 1 ..]
```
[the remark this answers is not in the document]
EmilLa1: Yes, this is bogus. 

I have corrected this and will add some tests.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +114,139 @@ import Concordium.Types.HashableTo
...
+      --  added in live blocks.
+      --  This is @Nothing@ if either the block is persisted or no accounts have been
+      --  added in the block.
+      --  Otherwise if the block is not persisted and accounts have been added, then
+      --  the 'DiffMap.DifferenceMap' yields the @AccountAddress -> AccountIndex@ mapping for
+      --  accounts created in the block, where the account addresses are in canonical form.
+      --  The 'DiffMap.DifferenceMap' is wrapped in an 'IORef' because it is inherited
+      --  by child blocks, and so when this block state is persisted then we need to clear it
+      --  for any children block states.
+      accountDiffMapRef :: !DiffMap.DifferenceMapReference
```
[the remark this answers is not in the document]
EmilLa1: Not sure if I understand, but that is exactly how we currently perform lookups.

However, I have cleaned up the documentation so it aligns more with what you wrote - and I agree it's simpler to think of it in that way.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -10,39 +10,84 @@
 -- for pattern matching. (See: https://gitlab.haskell.org/ghc/ghc/-/issues/20896)
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
 
+-- |
+--  * Adding accounts
+--  When an account is added (via ‘putNewAccount’) then it is first added to the ‘DifferenceMap’,
+--  it is kept in memory for the block until it either gets finalized or pruned.
+--  If a block is pruned then the retaining pointers are dropped and thus the block and associated ‘DifferenceMap’ is evicted from memory.
+--
+--  This in return invokes ‘storeUpdate’ for all underlying references for the block state, for the particular block.
```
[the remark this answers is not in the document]
EmilLa1: That line is now squashed - it was a remnant of some old documentation.

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -0,0 +1,139 @@
...
+--  Note that this implementation is very inefficient for large difference maps and thus should be revised
+--  if the credential deployments limit gets revised significantly.
+lookupExact :: (MonadIO m) => AccountAddress -> DifferenceMap -> m (Maybe AccountIndex)
+lookupExact addr diffMap = do
+    listOfAccs <- flatten diffMap
+    case find isEq listOfAccs of
+        Nothing -> return Nothing
+        Just (_, accIdx) -> return $ Just accIdx
+  where
+    isEq (accAddr, _) = addr == accAddr
```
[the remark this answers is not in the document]
EmilLa1: fixed as suggested - even though it hurts my eyes a bit.

concordium-consensus/src/Concordium/GlobalState/LMDB/Helpers.hs
```diff
@@ -271,18 +283,39 @@ data StoreEnv = StoreEnv
...
+defaultStepSize :: Int
+defaultStepSize = 2 ^ (26 :: Int) -- 64MB
+
+-- | Maximum step to increment the database size.
+defaultMaxStepSize :: Int
+defaultMaxStepSize = 2 ^ (30 :: Int) -- 1GB
+
+-- | Default start environment size.
+defaultEnvSize :: Int
+defaultEnvSize = 2 ^ (27 :: Int) -- 128MB
```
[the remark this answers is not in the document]
EmilLa1: added a bit of explanation. 

defaultEnvSize is moved out the the call sites instead. Reason is that the tree state databases are very much different in nature from the accountmap database.

concordium-consensus/src/Concordium/GlobalState/AccountMap/LMDB.hs
```diff
@@ -0,0 +1,275 @@
...
+            AccountMapStore
+                <$> mdb_dbi_open'
+                    txn
+                    (Just accountMapStoreName)
+                    [MDB_CREATE | not readOnly]
+        return DatabaseHandlers{..}
+
+-- | Create the lmdb stores and return back database handlers for interacting with it.
+--  This simply loads the references and does not initialize the databases.
+--  The initial environment size is set to 128MB.
```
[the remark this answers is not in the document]
EmilLa1: changed it to 4mb, seems as it should be enough for common use. There'll be a resize when doing the initial migration though. (currently mainnet is using 6.9mb for 90k accounts).

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -317,39 +316,30 @@ metadataStoreName = "metadata"
...
-dbInitSize :: Int
-dbInitSize = dbStepSize
-
 -- | Initialize database handlers in ReadWrite mode.
 --  This simply loads the references and does not initialize the databases.
---  The initial size is set to 64MB.
+--  The initial size is set to 128MB.
 databaseHandlers :: FilePath -> IO (DatabaseHandlers pv st)
-databaseHandlers treeStateDir = makeDatabaseHandlers treeStateDir False dbInitSize
+databaseHandlers treeStateDir = makeDatabaseHandlers treeStateDir False defaultEnvSize
```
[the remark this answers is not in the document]
EmilLa1: Changed the env size back to 64MB and revised doc.

concordium-consensus/src/Concordium/KonsensusV1/SkovMonad.hs
```diff
@@ -460,38 +498,39 @@ initialiseExistingSkovV1 ::
...
-                        runReaderT
-                            (runDiskLLDBM $ rollBackBlocksUntil checkBlockState)
-                            lldb
-                    when (rollCount > 0) $ do
+            pbsc <- liftIO $ do
+                pbscAccountCache <- newAccountCache (rpAccountsCacheSize gscRuntimeParameters)
+                pbscModuleCache <- Modules.newModuleCache (rpModulesCacheSize gscRuntimeParameters)
+                pbscBlobStore <- loadBlobStore gscBlockStateFile
+                pbscAccountMap <- LMDBAccountMap.openDatabase gscAccountMapDirectory
+                return PersistentBlockStateContext{..}
```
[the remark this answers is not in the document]
EmilLa1: uses `newPersistentBlockStateContext` now

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -317,39 +316,30 @@ metadataStoreName = "metadata"
...
+    -- | Initital database size
     Int ->
     IO (DatabaseHandlers pv st)
 makeDatabaseHandlers treeStateDir readOnly initSize = do
     _storeEnv <- makeStoreEnv
     -- here nobody else has access to the environment, so we need not lock
     let env = _storeEnv ^. seEnv
-    mdb_env_set_mapsize env (initSize + dbStepSize - initSize `mod` dbStepSize)
+        stepSize = _storeEnv ^. seStepSize
+    mdb_env_set_mapsize env (initSize + stepSize - initSize `mod` stepSize)
```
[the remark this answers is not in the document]
EmilLa1: I reverted this, so the default environment size is 64MB again, back from 128 MB.

concordium-consensus/src/Concordium/GlobalState/Persistent/Trie.hs
```diff
@@ -566,7 +566,7 @@ instance (Show v, FixShowable fix) => Show (TrieN fix k v) where
     show (TrieN _ t) = showFix showTrieFString t
 
 instance (BlobStorable m (fix (TrieF k v)), BlobStorable m v, Base (fix (TrieF k v)) ~ TrieF k v) => BlobStorable m (TrieN fix k v) where
-    storeUpdate v@EmptyTrieN = return (put (0 :: Int), v)
+    storeUpdate v@EmptyTrieN = return (put (0 :: Word64), v)
```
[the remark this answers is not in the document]
abizjak: Isn't it more sensible to fix the mistake of using an Int for size, when it's not technically specified what it's size is?

concordium-consensus/src/Concordium/KonsensusV1/SkovMonad.hs
```diff
@@ -446,6 +467,23 @@ data ExistingSkov pv m = ExistingSkov
       esProtocolUpdate :: !(Maybe ProtocolUpdate)
     }
 
+-- | Internal type used for deriving 'HasDatabaseHandlers' and 'LMDBAccountMap.HasDatabaseHandlers'
+--  used for computations where both lmdb databases are required.
+data LMDBDatabases pv = LMDBDatabases
+    { -- | the skov lmdb database
+      _lmdbSkovLmdb :: !(DatabaseHandlers pv),
+      -- | the account map lmdb database
+      _lmdbDSAccMap :: !LMDBAccountMap.DatabaseHandlers
```
td202, line 476: I don't love these names. Maybe `_lmdbSkov` and `_lmdbAccountMap`?

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -722,7 +685,9 @@ rollBackBlocksUntil ::
     ) =>
     -- | Callback for checking if the state at a given reference is valid.
     (BlockStateRef pv -> DiskLLDBM pv m Bool) ->
-    DiskLLDBM pv m (Int, BlockStateRef pv)
+    -- | Returns the number of blocks rolled back, the best state after the roll back and a list of
+    --  accounts created in certified blocks that was rolled back.
```
td202, line 689: This is not accurate.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/StartUp.hs
```diff
@@ -316,7 +322,12 @@ loadCertifiedBlocks ::
     m ()
 loadCertifiedBlocks = do
     certBlocks <- LowLevel.lookupCertifiedBlocks
-    mapM_ loadCertBlock certBlocks
+    -- The first certified block will have the empty parent difference map reference.
+    emptyParent <- DiffMap.newEmptyReference
+    -- Load all certified blocks
+    -- This sets the skov state, puts transactions in the transaction table,
+    -- and reconstructs the account map difference maps for the certified blocks.
+    foldM_ (flip loadCertBlock) emptyParent certBlocks
```
td202, line 330: This is not correct, because the certified blocks do not in general form a single chain - there may be branches. I think when you fold the accumulator should take a map from block hashes to the DiffMap for the block. Then `loadCertBlock` will look up the DiffMap for the parent, and if it's not found use the `emptyParent` (as the parent is presumably finalized). It will also add the DiffMap constructed for the block to the map.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/TransactionProcessingTest.hs
```diff
@@ -128,6 +131,15 @@ instance (MonadReader r m) => MonadReader r (FixedTimeT m) where
...
 
+newtype NoLoggerT m a = NoLoggerT {runNoLoggerT :: m a}
+    deriving (Functor, Applicative, Monad, MonadIO, MonadReader r, MonadFail, TimeMonad, MonadState s)
+
+instance (Monad m) => MonadLogger (NoLoggerT m) where
+    logEvent _ _ _ = return ()
+
+-- deriving instance (TimeMonad m) => TimeMonad (NoLoggerT m)
+-- deriving instance (MonadState s m) => MonadState s (NoLoggerT m)
+
```
td202, line 142:
```suggestion
```

concordium-consensus/tests/globalstate/GlobalStateTests/Accounts.hs
```diff
@@ -1,34 +1,37 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
td202, line 1: This is generally a red flag. Why is it needed, and can we tweak to avoid it?

td202 (commented):
Looks pretty good now. Just a few smaller things.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Finality.hs
```diff
@@ -189,14 +190,25 @@ processFinalizationEntry newFinalizedPtr newFinalizationEntry =
...
-makeStoredBlock blockPtr = do
-    statePointer <- saveBlockState (bpState blockPtr)
+makeStoredBlock finalized blockPtr = do
+    statePointer <-
+        if finalized
+            then do
+                ref <- saveBlockState (bpState blockPtr)
+                saveAccounts (bpState blockPtr)
+                return ref
+            else saveBlockState (bpState blockPtr)
```
td202, line 211:
```suggestion
    statePointer <- saveBlockState (bpState blockPtr)
    when finalized $ saveAccounts (bpState blockPtr)
```

concordium-consensus/src/Concordium/KonsensusV1/TreeState/StartUp.hs
EmilLa1: great catch

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -168,35 +285,71 @@ getAccountByCredId cid accs@Accounts{..} =
...
+    -- We do the bounds check as it could be that the lmdb backed account map
+    -- yields accounts which are not yet present in the @accountTable@.
+    -- In particular this can be the case if finalized blocks have been rolled
+    -- back as part of database recovery.
+    --
+    -- The extra @diffMapSize@ constraint is necessary in the case where a parent block to the last finalized block
+    -- is used for extending. This is normally not the case as blocks are typically built from the "best" block always.
+    --
+    -- However for scenarios like dry runs, then we need this extra constraint, as otherwise an account created in the
+    -- in the dry run would point to an account created in a successor of the block that the dry run is extending.
```
abizjak, line 325: As far as I recall no accounts can be created as part of a dry run. So this comment is odd.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -168,35 +285,71 @@ getAccountByCredId cid accs@Accounts{..} =
...
+    lookupDisk diffMapSize =
```
abizjak, line 326: It would be simpler to argue for the need for this diffMapSize on the basis of hygiene rather than this case by case basis.

It makes sense that if the account is not in the diffmap then it must have an index < #num of account - size of diffMap.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -168,35 +285,71 @@ getAccountByCredId cid accs@Accounts{..} =
...
+                    DiffMap.lookupViaEquivalenceClass (accountAddressEmbed addr) accDiffMap >>= \case
+                        Just accIdx -> return $ Just accIdx
+                        Nothing -> do
+                            diffMapSize <- length <$> DiffMap.flatten accDiffMap
+                            lookupDisk $ fromIntegral diffMapSize
+                else
+                    DiffMap.lookupExact addr accDiffMap >>= \case
+                        Just accIdx -> return $ Just accIdx
+                        Nothing -> do
+                            diffMapSize <- length <$> DiffMap.flatten accDiffMap
```
abizjak, line 311: Is there no better way of keeping track of the size?

In normal operation this should be quick enough, since few accounts are created per block, and there are few non-finalized ones. But it's a bit unsatisfactory.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -168,35 +285,71 @@ getAccountByCredId cid accs@Accounts{..} =
...
+  where
+    -- Lookup the 'AccountIndex' in the lmdb backed account map,
+    -- and make sure it's within the bounds of the account table.
+    -- We do the bounds check as it could be that the lmdb backed account map
+    -- yields accounts which are not yet present in the @accountTable@.
+    -- In particular this can be the case if finalized blocks have been rolled
+    -- back as part of database recovery.
+    --
+    -- The extra @diffMapSize@ constraint is necessary in the case where a parent block to the last finalized block
+    -- is used for extending. This is normally not the case as blocks are typically built from the "best" block always.
```
abizjak, line 322: I don't understand which block is "a parent block to the last finalized block is used for extending". Could you please state things more directly.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -256,24 +413,67 @@ updateAccountsAtIndex' fupd ai = fmap snd . updateAccountsAtIndex fupd' ai
   where
     fupd' = fmap ((),) . fupd
 
+-- | Get a list of all account addresses and their associated account indices.
+--  There are no guarantees of the order of the list. This is because the resulting list is potentially
+--  a concatenation of two lists of account addresses.
```
abizjak, line 418: The last sentence isn't really useful. The diffMap flatten on its own says nothing about the order.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -256,24 +413,67 @@ updateAccountsAtIndex' fupd ai = fmap snd . updateAccountsAtIndex fupd' ai
...
+-- | Fold over the account table in descending order of account index.
 foldAccountsDesc :: (SupportsPersistentAccount pv m) => (a -> PersistentAccount (AccountVersionFor pv) -> m a) -> a -> Accounts pv -> m a
 foldAccountsDesc f a accts = L.mfoldDesc f a (accountTable accts)
 
+-- | Initialize the LMDB account map if it is not already.
+--  This puts in all accounts from the account table of the provided block state into the account map.
+--  If there already are accounts present in the account map, then we check that the size of the account map
+--  corresponds with number of accounts in the account table.
+--  If the number of accounts in the account table and account map matches, then this function does nothing.
+--  If they do not match, wipe the account map and recreate it from the account table.
```
abizjak, line 455: This seems like an overreaction and not in line with what we discussed.

It will be a common occurrence in crashes that the account map in LMDB will have more accounts than for the last finalized block stored in the tree state database. But this should be fine since they are finalized.

Why is it necessary to spend a lot of time reconstructing in that case? LMDB has complete ASID guarantees if I recall, so we can rely on the data being there fully for every account written. 

An if we expect to have a lot more accounts then we should not be cavalier about doing all this work. And that was meant to be the goal of this feature.

abizjak: Note that it is a common case that even in normal operation you have more accounts in the LMDB account map than what is in the block state for any particular block.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -357,13 +308,19 @@ makeDatabaseHandlers treeStateDir readOnly initSize = do
                     [MDB_CREATE | not readOnly]
         return DatabaseHandlers{..}
 
+-- | Default start environment size.
+--  Tree state database sizes for historical protocol versions have been between 7-60 times
+--  the 'defaultEnvSize'.
+defaultEnvSize :: Int
+defaultEnvSize = 2 ^ (27 :: Int) -- 128MB
```
abizjak, line 315: We have multiple of these constants, some 64, some 128MB. Why?

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
```diff
@@ -879,6 +849,7 @@ rollBackBlocksUntil checkState = do
                         let finHash = getHash fin
                         _ <- deleteRecord txn (dbh ^. blockStore) finHash
                         _ <- deleteRecord txn (dbh ^. finalizedBlockIndex) (blockHeight fin)
+
```
abizjak, line 852: Good change there :P

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/TransactionProcessingTest.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
```
abizjak, line 5: No need for this I think.

concordium-consensus/tests/e2e/EndToEndTests/CredentialDeploymentTests.hs
```diff
@@ -0,0 +1,328 @@
...
+          bbEpoch = 0,
+          bbTimestamp = 1_000,
+          bbBaker = bakerId,
+          bbQuorumCertificate = genesisQuorumCertificate genesisHash,
+          bbTimeoutCertificate = Absent,
+          bbEpochFinalizationEntry = Absent,
+          bbNonce = computeBlockNonce genesisLEN 1 (bakerVRFKey bakerId),
+          bbTransactions = Vec.fromList [credBi1],
+          bbTransactionOutcomesHash = read "b9444648bf759471276fdba1930af0c543847d22de89c27939791898d757516d",
+          bbStateHash = read "b8bc96ec5f162db36784ea96ec29e3e8ad92abff341a6847e3bf524fdada28ff"
```
abizjak, line 87: Please explain where these come from. How do we maintain this?

concordium-consensus/tests/e2e/EndToEndTests/CredentialDeploymentTests.hs
```diff
@@ -0,0 +1,328 @@
...
+          bbEpoch = 0,
+          bbTimestamp = 3_000,
+          bbBaker = bakerId,
+          bbQuorumCertificate = validQCFor testBB1,
+          bbTimeoutCertificate = Absent,
+          bbEpochFinalizationEntry = Absent,
+          bbNonce = computeBlockNonce genesisLEN 2 (bakerVRFKey bakerId),
+          bbTransactions = Vec.empty,
+          bbTransactionOutcomesHash = read "375fef64a251f353d608171d283d00fe00aa0bd77596ba7703c810f48056ef89",
+          bbStateHash = read "798d5089818bcc7b8873e2585fb4fbf3d4dceffca32531259f466e7c435c8817"
```
abizjak, line 107: Also here.

concordium-consensus/tests/e2e/EndToEndTests/CredentialDeploymentTests.hs
```diff
@@ -0,0 +1,328 @@
...
+    bakerId = 3
+
+-- | Compute the 'AccountCreation' from the provided 'AccountCreation'.
+getAccAddress :: AccountCreation -> AccountAddress
+getAccAddress accCreation = case credential accCreation of
+    InitialACWP x -> initialCredentialAccountAddress $ icdiValues x
+    NormalACWP x -> credentialAccountAddress $ cdiValues x
+
+-- | Test that two credential deployments (each on their own branch and with same block height) does not:
+--  * Alter the state of the parent block (a new child difference map and assoicated reference is created).
```
abizjak, line 233:
```suggestion
--  * Alter the state of the parent block (a new child difference map and associated reference is created).
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +116,166 @@ import Concordium.Types.HashableTo
...
+--  * The function assumes that the account table already yields every account added for the block state.
+--  * The provided 'DiffMap.DifferenceMapReference@ MUST correspond to the parent map.
+--  * The provided list of accounts MUST be in ascending order of account index, hence the list of accounts
+--   MUST be provided in the order of which the corresponding credential deployment transactions were executed.
+reconstructDifferenceMap ::
+    (SupportsPersistentAccount pv m) =>
+    -- | Reference to the parent difference map.
+    DiffMap.DifferenceMapReference ->
+    -- | Account addresses to add to the difference map.
+    [AccountAddress] ->
```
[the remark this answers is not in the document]
EmilLa1: Is it not clear from the documentation above? 
```
--  * The provided list of accounts MUST be in ascending order of account index, hence the list of accounts
--   MUST be provided in the order of which the corresponding credential deployment transactions were executed.
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -69,96 +116,166 @@ import Concordium.Types.HashableTo
...
+mkNewChildDifferenceMap :: (SupportsPersistentAccount pv m) => Accounts pv -> m (Accounts pv)
+mkNewChildDifferenceMap accts@Accounts{..} = do
+    newDiffMapRef <- liftIO $ newIORef $ Present $ DiffMap.empty accountDiffMapRef
+    return accts{accountDiffMapRef = newDiffMapRef}
+
+-- | Create and set the 'DiffMap.DifferenceMap' for the provided @Accounts pv@.
+--  The function is highly unsafe and can cause state invariant failures if not all of the
+--  below preconditions are respected.
+--  Precondition:
+--  * The function assumes that the account table already yields every account added for the block state.
```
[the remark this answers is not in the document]
EmilLa1: Sounds like you understood it :-).. I've phrased it differently.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/LowLevel/LMDB.hs
EmilLa1: mistake, there should only be the 128mb one. (that is how it's been before)

concordium-consensus/tests/e2e/EndToEndTests/CredentialDeploymentTests.hs
EmilLa1: The `TestMonad` has a logger instance which will log errors. If any of the hashes does not match then the test will fail and the execution environment will log it. So one can grab the hashes from there. 

These will need to change when e.g. the hashing scheme is changed.

concordium-consensus/tests/e2e/EndToEndTests/CredentialDeploymentTests.hs
EmilLa1: See above comment.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
EmilLa1: We can cache it per layer.

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
abizjak: The most common case of a crash and database corruption we see is exactly that the node is killed between writing to the different databases. That's what I'm talking about. Yes the account map is written first, that is kind of my point.

That the accounts there will always be for finalized blocks, but maybe for more finalized blocks than we have recorded in the tree state database. And that is fine as I see it.

EmilLa1: fixed

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -393,17 +416,16 @@ updateAccountsAtIndex' fupd ai = fmap snd . updateAccountsAtIndex fupd' ai
...
-    -- Get all persisted accounts from the account map up to and including the last account of the account table.
-    persistedAccs <- LMDBAccountMap.getAllAccounts $ (AccountIndex . L.size) (accountTable accounts) - 1
     mDiffMap <- liftIO $ readIORef (accountDiffMapRef accounts)
     case mDiffMap of
-        Absent -> return persistedAccs
+        Absent -> LMDBAccountMap.getAllAccounts $ (AccountIndex . L.size) (accountTable accounts) - 1
         Present accDiffMap -> do
+            -- Get all persisted accounts from the account map up to and including the last account of the account table minus what we found the in the difference map.
             flattenedDiffMapAccounts <- DiffMap.flatten accDiffMap
+            persistedAccs <- LMDBAccountMap.getAllAccounts $ (AccountIndex . L.size) (accountTable accounts) - (1 + fromIntegral (length flattenedDiffMapAccounts))
```
td202, line 428:
```suggestion
            persistedAccs <- LMDBAccountMap.getAllAccounts . AccountIndex $ L.size (accountTable accounts) - (1 + fromIntegral (length flattenedDiffMapAccounts))
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Accounts.hs
```diff
@@ -256,24 +415,70 @@ updateAccountsAtIndex' fupd ai = fmap snd . updateAccountsAtIndex fupd' ai
...
+-- | Fold over the account table in descending order of account index.
 foldAccountsDesc :: (SupportsPersistentAccount pv m) => (a -> PersistentAccount (AccountVersionFor pv) -> m a) -> a -> Accounts pv -> m a
 foldAccountsDesc f a accts = L.mfoldDesc f a (accountTable accts)
 
+-- | Initialize the LMDB account map if it is not already.
+--  This puts in all accounts from the account table of the provided block state into the account map.
+--  If there already are accounts present in the account map, then we check that the size of the account map
+--  corresponds with number of accounts in the account table.
+--  If the number of accounts in the account table and account map matches, then this function does nothing.
+--  If they do not match, wipe the account map and recreate it from the account table.
```
td202, line 456:
```suggestion
--  If the account map has fewer accounts than the provided account table, the account map is
--  wiped and repopulated from the account table. Otherwise, the account map is unchanged.
```

concordium-consensus/tests/e2e/Spec.hs
```diff
@@ -0,0 +1,21 @@
+module Main where
```
td202, line 1: I am not sure why this should be a separate test suite - doesn't it make sense for it to just combine with the consensus tests?

td202 (approved):

concordium-consensus/src/Concordium/GlobalState/AccountMap/DifferenceMap.hs
```diff
@@ -31,6 +31,8 @@ module Concordium.GlobalState.AccountMap.DifferenceMap (
     -- Lookup in a difference map (and potential parent maps) whether
     -- it yields the 'AccountIndex' for the provided 'AccountAddress'.
     lookupExact,
+    -- Clear up the references of difference map(s).
```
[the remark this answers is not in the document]
EmilLa1: oki, removing

concordium-consensus/tests/e2e/Spec.hs
abizjak: I would also prefer that. It is in the style of other consensus tests.

concordium-consensus/tests/e2e/EndToEndTests/CredentialDeploymentTests.hs
abizjak: Ok, but my point is that if somebody is to maintain this in the future, what are they supposed to do?

It's not clear, so please document this. Nobody will find this comment in a year when this needs to be updated. Explain in the code what the rationale is.

abizjak (approved):

concordium-consensus/tests/e2e/Spec.hs
EmilLa1: It's moved now.

Issue #1045, opened 2023-10-02
Start-up time regression


td202:
**Bug Description**

Start-up time on mainnet for node 6.1.5 (~2m 36s) is slower than for node 6.0.4 (~1m 26s) before catch-up.

This regression is due to the fix #1020 for issue #147. Prior to the fix, a dummy value was used for the block state hash of genesis blocks. This avoided having to compute the block state hash, but meant that an incorrect value was returned by queries that requested it. The fix causes the hash to always be computed at start-up. This applies to all (re)genesis blocks on consensus version 0.

**Steps to Reproduce**

Compare the start-up times on mainnet from an existing recent database for 6.1.5 and 6.0.4 until "Consensus layer started".

**Expected Result**

The times should be very close.

**Actual Result**

The 6.1.5 node is slower (~2m 36s compared to ~1m 26s for @mh-concordium).

**Versions**
- Software Version: 6.1.2 - 6.1.5
- OS: all

PR #1046, opened 2023-10-02
Fix startup time regression

td202:
## Purpose

Addresses #1045

This fixes a regression in start-up time that is caused by always having to recompute the block state hashes of genesis blocks. Instead, the block state hashes are computed once and stored in the tree state database, so that future restarts will not need to compute them.

## Changes

These changes apply only to consensus version 0. Consenus version 1 (ConcordiumBFT) already stores the genesis state hash in the tree state database.

- Add a "genesisStateHash" entry in the metadata table of the tree state.
- When looking up blocks in the tree state database, also return the state hash, if it is available. (This is either present in the block, for baked blocks, or is looked up in the metadata table, for the genesis block.)
- Set the "genesisStateHash" when starting a new genesis, or when loading an existing state where it has not already been written.
- When loading blocks, use the state hash provided by the tree state database where available. The state hash only needs to be computed for genesis blocks when the "genesisStateHash" is not present.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
@@ -38,6 +38,7 @@ module Concordium.GlobalState.Persistent.LMDB (
     getFinalizedBlockAtHeight,
     getLastBlock,
     getFirstBlock,
+    writeGenesisBlockStateHash,
     writeBlock,
     writeFinalizationRecord,
     writeTransactionStatus,
@@ -216,6 +217,13 @@ instance MDBDatabase MetadataStore where
 --  This key should map to a serialized 'VersionMetadata' structure.
 versionMetadata :: DBKey MetadataStore
 versionMetadata = "version"
+
+-- | Key used to store the state hash for the genesis block in the metadata.
+--  If a value is present at this key, it must be the block state hash for the genesis block.
+--  Since this entry was not present in older databases, its absence is permissible and must be
+--  handled gracefully.
+genesisStateHashMetadata :: DBKey MetadataStore
+genesisStateHashMetadata = "genesisStateHash"
 
 data VersionMetadata = VersionMetadata
     { -- | Version signifier for the database itself.
@@ -396,10 +404,12 @@ initializeDatabase ::
     PersistentBlockPointer pv bs ->
     -- | Genesis block state
     st ->
+    -- | Genesis block state hash
+    StateHash ->
     -- | Tree state directory
     FilePath ->
     IO (DatabaseHandlers pv st)
-initializeDatabase gb stRef treeStateDir = do
+initializeDatabase gb stRef gbStateHash treeStateDir = do
     createDirectoryIfMissing False treeStateDir
     let storedGenesis =
             StoredBlock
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -608,22 +638,25 @@ loadBlocksFinalizationIndexes dbh = transaction (dbh ^. storeEnv) True $ \txn ->
...
-            loadRecord txn (dbh ^. blockStore) (finalizationBlockPointer finRec) >>= \case
+            loadBlock txn dbh (finalizationBlockPointer finRec) >>= \case
                 Just block -> return $ Right (finRec, block)
                 Nothing -> return . Left $ "Could not read last finalized block by hash " <> show (finalizationBlockPointer finRec)
         Just (Left s) -> return $ Left $ "Could not read last finalization record: " ++ s
         Nothing -> return $ Left "No last finalized block found"
 
 -- | Get the first block
-getFirstBlock :: (IsProtocolVersion pv, FixedSizeSerialization st) => DatabaseHandlers pv st -> IO (Maybe (StoredBlock pv st))
+getFirstBlock :: (IsProtocolVersion pv, FixedSizeSerialization st) => DatabaseHandlers pv st -> IO (Maybe (StoredBlock pv st, Maybe StateHash))
```
abizjak, line 656: Please improve documentation or introduce a type or alias for the pair
```
(StoredBlock pv st, Maybe StateHash)
```

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -416,12 +416,17 @@ loadSkovPersistentData rp _treeStateDirectory pbsc = do
...
         Right (_lastFinalizationRecord, lfStoredBlock) -> do
             -- Truncate the blobstore beyond the last finalized blockstate.
-            liftIO $ truncateBlobStore (bscBlobStore . PBS.pbscBlobStore $ pbsc) (sbState lfStoredBlock)
+            liftIO $ truncateBlobStore (bscBlobStore . PBS.pbscBlobStore $ pbsc) (sbState . fst $ lfStoredBlock)
             -- Get the genesis block.
             genStoredBlock <-
                 maybe (logExceptionAndThrowTS GenesisBlockNotInDataBaseError) return
                     =<< liftIO (getFirstBlock _db)
             _genesisBlockPointer <- liftIO $ makeBlockPointer genStoredBlock
+            when (isNothing (snd genStoredBlock)) $ do
```
abizjak, line 425: It'd be useful to clarify here that this is only expected to run when upgrading from node version < 6.1

abizjak (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/LMDB.hs
```diff
@@ -513,16 +524,35 @@ resizeDatabaseHandlers dbh size = do
     logEvent LMDB LLDebug $ "Resizing database from " ++ show oldMapSize ++ " to " ++ show newMapSize
     liftIO . withWriteStoreEnv (dbh ^. storeEnv) $ flip mdb_env_set_mapsize newMapSize
 
+-- | Load a block and its state hash (if available).
+-- Normal blocks already contain their state hash. For genesis blocks, the state hash is loaded
+-- from the metadata table if it is present there.
+loadBlock ::
```
EmilLa1, line 530: Shouldn't it always be present given that you are setting it when calling `loadSkovPersistentData` ? Maybe that `Maybe` is not actually required?

EmilLa1 (approved):

limemloh (approved):

PR #1051, opened 2023-10-05
Dry run

td202:
## Purpose

Support for dry-running transactions and contract operations.

## Changes

- Scheduler refactoring to support dry-running transactions and observing return values for smart contract invocations.
- Implement a dry-run FFI that maintains state in Haskell, and provides calls for each dry run operation.
- Implement the dry-run endpoint in the Rust GRPC, taking a stream input and making the corresponding FFI calls.

## Checklist

      hard-to-understand areas.


EmilLa1 (commented):
Please add it to the changelog.

EmilLa1 (approved):
Looks great

concordium-node/src/grpc2.rs
```diff
@@ -2470,6 +2492,93 @@ pub mod server {
...
+                        request = input.next() => { Ok(request) }
+                        _ = &mut sleep => {
+                            Err(tonic::Status::deadline_exceeded("dry run deadline elapsed"))
+                        }
+                    }?;
+
+                    if let Some(dry_run_request) = next {
+                        let request = dry_run_request.map_err(|_| {
+                            tonic::Status::invalid_argument("invalid dry run request")
+                        })?;
```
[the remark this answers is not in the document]
td202: I have changed to use `try_stream`. It does shut down the stream.

concordium-node/src/grpc2.rs
```diff
@@ -2470,6 +2492,87 @@ pub mod server {
...
+                            use crate::grpc2::types::dry_run_state_operation::Operation::*;
+                            match operation.operation.require()? {
+                                SetTimestamp(timestamp) => yield dry_run.set_timestamp(timestamp)?,
+                                MintToAccount(mint) => yield dry_run.mint_to_account(mint)?,
+                                RunTransaction(run_transaction_input) => {
+                                    yield dry_run.transaction(run_transaction_input)?
+                                }
+                            }
+                        }
+                    }
```
abizjak, line 2563: Would it not make more sense to make a separate stream object which owns the necessary data instead of bringing in this dependency and using rather obscure methods.

Something like

```rust
struct Response {
    pub dry_run: DryRun,
    pub input:   tonic::Streaming<self::types::DryRunRequest>,
    pub timeout: std::pin::Pin<Box<tokio::time::Sleep>>,
    pub permit: Option<OwnedSemaphorePermit>,
    pub done: bool,
}

impl futures::Stream for Response {
    type Item = tonic::Result<Vec<u8>>;

    fn poll_next(
        mut self: std::pin::Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
    ) -> std::task::Poll<Option<Self::Item>> {
        use std::task::Poll::Ready;
        if self.done {
            return Ready(None);
        };
        if let std::task::Poll::Ready(_) = std::pin::Pin::new(&mut self.timeout).poll(cx) {
            self.done = true;
            return Ready(Some(Err(tonic::Status::deadline_exceeded("..."))));
        };
        let Some(dry_run_request) = futures::ready!(std::pin::Pin::new(&mut self.input).poll_next(cx)) else {
            return Ready(None);
        };
        let request = dry_run_request
            .map_err(|_| tonic::Status::invalid_argument("invalid dry run request"))?;
        use crate::grpc2::types::dry_run_request::Request::*;
        let result = match request.request.require()? {
            LoadBlockState(block_hash_input) => self.dry_run.load_block_state(&block_hash_input),
            StateQuery(query) => {
                use crate::grpc2::types::dry_run_state_query::Query::*;
                match query.query.require()? {
                    GetAccountInfo(account) => self.dry_run.get_account_info(&account),
                    GetInstanceInfo(instance) => self.dry_run.get_instance_info(&instance),
                    InvokeInstance(invoke_instance_input) => {
                        self.dry_run.invoke_instance(&invoke_instance_input)
                    }
                }
            }
            StateOperation(operation) => {
                use crate::grpc2::types::dry_run_state_operation::Operation::*;
                match operation.operation.require()? {
                    SetTimestamp(timestamp) => self.dry_run.set_timestamp(timestamp),
                    MintToAccount(mint) => self.dry_run.mint_to_account(mint),
                    RunTransaction(run_transaction_input) => {
                        self.dry_run.transaction(run_transaction_input)
                    }
                }
            }
        };
        Ready(Some(result))
    }
}
```

concordium-node/src/grpc2.rs
```diff
@@ -880,6 +883,10 @@ pub mod server {
         net::SocketAddr,
         sync::{Arc, Mutex},
     };
+    use tokio::{
+        time,
+        time::{Duration, Instant},
```
abizjak, line 888: I found the use of this confusing since there is also std::time. Can you use qualified names instead of importing `time` here?

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1665,6 +1796,233 @@ pub fn get_consensus_ptr(
...
+        // Optional Address to ffi
+        let (
+            invoker_address_type,
+            invoker_account_address_ptr,
+            invoker_contract_index,
+            invoker_contract_subindex,
+        ) = if let Some(address) = &request.invoker {
+            match address.r#type.as_ref().require()? {
+                crate::grpc2::types::address::Type::Account(account) => {
+                    (1, crate::grpc2::types::account_address_to_ffi(account).require()?, 0, 0)
```
abizjak, line 1877: It's not entirely clear this is ok since you're making temporary references here, then taking a pointer to those, and dropping the references.

It probably is, but usually it'd be best if the reference to `account` is retained after the pointer is needed..

concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -222,73 +223,110 @@ checkTransactionVerificationResult (TVer.NotOk TVer.InvalidPayloadSize) = Left I
...
+            -- During processing of transactions the amount on the sender account is decreased by the
+            -- amount corresponding to the deposited energy, i.e., the maximum amount that can be charged
+            -- for execution. The amount corresponding to the unused energy is refunded at the end of
+            -- processing; see `withDeposit`.
+            -- Note, for transactions that require specific constraints on the protocol version,
+            -- those constraints are asserted.  'decodePayload' ensures that those assertions
+            -- will not fail.
+            case payload of
+                -- Update is the only operation that can produce a return value, so
+                -- 'handleUpdateContract' is polymorphic in the result type.
```
abizjak, line 317: This is not true, initialization can also produce return values.

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+-- | The base cost for dry-running a transaction.
+--  In addition, the energy cost of the execution will be charged.
+costTransactionBase :: Energy
+costTransactionBase = 400
+
+-- | An opaque rust vector.
+data ForeignVec
+
+-- | A function that copies bytes to a rust vector.
+type CopyToForeignVec = Ptr ForeignVec -> Ptr Word8 -> Int64 -> IO ()
```
abizjak, line 102: Don't we have this already in External? We certainly do this for all streaming calls.

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+                        ( DryRunSuccessBlockStateLoaded
+                            { drsBlockHash = blkHash,
+                              drsCurrentTimestamp = ts,
+                              drsProtocolVersion = bscProtocolVersion newBSC
+                            }
+                        )
+            return OK
+
+-- | A dummy 'StateHash' value that can be used to create a 'HashedPersistentBlockState' where the
+-- actual hash does not matter.
```
abizjak, line 353: There should be some explanation about the context and the need for this. What problem does it solve.

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+                            (ContractSubindex contractSubindex),
+                      ccAmount = Amount amount,
+                      ccMethod = method,
+                      ccParameter = parameter,
+                      ccEnergy = energyLimit
+                    }
+        res <- runWithEBlockStateContext shiMVR shiBSC $ \drsRef -> do
+            -- We "freeze" the block state using a dummy hash so that we do not recompute the
+            -- state hash, since it is not required by invokeContract.
+            DryRunState{..} <- liftIO $ readIORef drsRef
```
abizjak, line 572: Could we change the InvokeContractMonad to not need the frozen  state instead?

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+                        }
+            let schedulerState =
+                    Scheduler.SchedulerState
+                        { _ssNextIndex = 0,
+                          _ssExecutionCosts = 0,
+                          _ssEnergyUsed = 0,
+                          _ssBlockState = drsBlockState
+                        }
+            let exec = flip runContT return $ do
+                    let exit = ContT . const . return
```
abizjak, line 711:
```suggestion
                    let exit = ContT . const . return . Right . toProto
```

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+                                        }
+                            let transaction = makeAccountTransaction dummySignature header payload
+                            let cost =
+                                    Cost.baseCost
+                                        (getTransactionHeaderPayloadSize header)
+                                        signatureCount
+                            -- Check that the energy amount covers the base cost of checking
+                            -- the transaction header.
+                            when (thEnergyAmount header < cost) $
+                                exit . Right . toProto $
```
abizjak, line 737:
```suggestion
                                exit $
```

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+                                    DryRunResponse
+                                        (DryRunErrorEnergyInsufficient cost)
+                                        shiQuotaRem
+                            -- Check that the sender account has sufficient funds to cover the
+                            -- deposit amount.
+                            accBalance <- lift $ TVer.getAccountAvailableAmount acc
+                            depositAmount <- lift $ TVer.energyToCcd (thEnergyAmount header)
+                            when (accBalance < depositAmount) $
+                                exit . Right . toProto $
```
abizjak, line 746:
```suggestion
                                exit $
```

concordium-consensus/src/Concordium/External/DryRun.hs
```diff
@@ -0,0 +1,881 @@
...
+    -- | The return code (see 'DryRunReturnCode').
+    IO Int64
+dryRunTransaction dryRunPtr senderPtr energyLimit payloadPtr payloadLen sigPairs sigCount outVec =
+    dryRunStateHelper dryRunPtr outVec costTransactionBase $ \StateHelperInfo{..} -> do
+        sender <- decodeAccountAddress senderPtr
+        payloadBytes <- BS.packCStringLen (castPtr payloadPtr, fromIntegral payloadLen)
+        let payload = EncodedPayload (SBS.toShort payloadBytes)
+        sigMap <-
+            if sigCount == 0
+                then return (Map.singleton 0 (Map.singleton 0 Sig.dummySignatureEd25519))
```
abizjak, line 680: It's unclear to me where we need this dummy value. For getting the cost we just pass in the number of signatures.

What is preventing us from not having this dummy signature? It's rather ugly.

abizjak (changes requested):
It's good work, but there are some issues the way I see it.

1. Initializing contracts also returns values (for V1 at least).
2. There were two main use-cases for this feature, one for estimating costs for complex protocols from a dapp is well-covered by the work done. But the second one which is to re-execute a transaction, in particular a contract update or init in the middle of the block is not. In particular to do that reliably there must be a way to create new accounts, otherwise the behaviour of the contract call might be different.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1665,6 +1796,233 @@ pub fn get_consensus_ptr(
...
+                sig.credential.try_into().map_err(|_| {
+                    tonic::Status::invalid_argument("Credential index out of bounds")
+                })?,
+            );
+            signature_vec.push(
+                sig.key
+                    .try_into()
+                    .map_err(|_| tonic::Status::invalid_argument("Key index out of bounds"))?,
+            );
+        }
```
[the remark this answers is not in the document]
td202: As I mentioned there, there's no real benefit to using a map. It would have to be a map of maps, which also means introducing a new type for the inner map, and then the target would be empty values. It's just more convenient to work with a list of pairs and not worry too much about duplicates.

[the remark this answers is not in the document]
abizjak: I disagree but not enough to persist at it.

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -1665,6 +1805,234 @@ pub fn get_consensus_ptr(
...
+    }
+
+    /// Convert a result code returned by a dry-run FFI into an appropriate
+    /// [`tonic::Status`].
+    fn check_result(result: i64, origin: &str) -> Result<(), tonic::Status> {
+        match result {
+            0 => Ok(()),
+            1 => Err(tonic::Status::internal(format!(
+                "Internal error: {} could not be completed",
+                origin
```
abizjak, line 2026:
```suggestion
                "Internal error: {origin} could not be completed"
```

abizjak (approved):

concordium-node/src/grpc2.rs
```diff
@@ -2470,6 +2486,120 @@ pub mod server {
...
+            cx: &mut std::task::Context<'_>,
+        ) -> std::task::Poll<Option<Self::Item>> {
+            use std::task::Poll::Ready;
+            if self.done {
+                return Ready(None);
+            }
+            let timeout = std::pin::pin!(&mut self.timeout);
+            if timeout.poll(cx).is_ready() {
+                self.done = true;
+                return Ready(Some(Err(tonic::Status::deadline_exceeded(
```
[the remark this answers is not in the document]
td202: The permit should be dropped when we drop the dry-run handle, which is when the stream is dropped. Releasing the permit early like this seems to defat this.

concordium-consensus/src/Concordium/External/DryRun.hs
td202: This should be part of a larger refactor to change how operations work on updatable/non-updatable block states.

concordium-consensus/src/Concordium/External/DryRun.hs
td202: Which signatures are present affects the behaviour of some transactions.

PR #1061, opened 2023-10-31
Trim the transaction table

EmilLa1:
## Purpose

The `TransactionTable` and in particular non finalized transactions now only retains entries for accounts which have non-finalized transactions.
Accounts without non-finalized transactions are expunged from the transaction table and in that case the last finalized block is used for keeping track of the next available account nonce.  

## Changes

- When purging, then remove any entries (from the non-finalized transactions table) that has no non-finalized transactions.
- Don't populate the transaction table at startup. 

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -774,16 +785,20 @@ finalizeTransactions = mapM_ removeTrans
     removeTrans WithMetadata{wmdData = NormalTransaction tr, ..} = do
         let nonce = transactionNonce tr
             sender = accountAddressEmbed (transactionSender tr)
-        anft <- use (transactionTable . TT.ttNonFinalizedTransactions . at' sender . non TT.emptyANFT)
-        unless (anft ^. TT.anftNextNonce == nonce) $
+        senderMinNonce <- do
+            macct <- flip getAccount (aaeAddress sender) =<< use (lastFinalized . to bpState)
+            fromMaybe minNonce <$> mapM (getAccountNonce . snd) macct
+        nextNonce <- use (transactionTable . TT.ttNonFinalizedTransactions . at' sender . non (TT.emptyANFTWithNonce senderMinNonce) . TT.anftNextNonce)
+        unless (nextNonce == nonce) $
```
abizjak, line 792: I think it's a bad idea to look up the account for each transaction we finalize here. The `unless` check is a debug assertion, it is not meant to be triggered.

Also, since you are finalizing transactions, should they not be in the `ttNonFinalizedTransactions` already?

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -774,16 +785,20 @@ finalizeTransactions = mapM_ removeTrans
...
             throwM . TreeStateInvariantViolation $
                 "The recorded next nonce for the account "
                     ++ show sender
                     ++ " ("
-                    ++ show (anft ^. TT.anftNextNonce)
+                    ++ show nextNonce
                     ++ ") doesn't match the one that is going to be finalized ("
                     ++ show nonce
                     ++ ")"
+        anft <- use (transactionTable . TT.ttNonFinalizedTransactions . at' sender . non TT.emptyANFT)
```
abizjak, line 801: This might be OK but it is a bit wonky.  In principle this anft is wrong. The next nonce is not the right one.

But as I wrote above, should this ever really be empty? Since we are finalizing the transactions in this function, they should be in the ttNonFianzliedTransactions, should they not?

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -800,11 +815,12 @@ finalizeTransactions = mapM_ removeTrans
         transactionTable
             . TT.ttNonFinalizedTransactions
             . at' sender
-            ?=! ( anft
+            . non TT.emptyANFT
+            .=! ( anft
```
abizjak, line 819: The `non` here will not remove the empty value, since the `anftNextNonce` is not `minNonce`, which is what `emptyANFT` has.

I suggest being a bit more verbose and explicit in writing out the logic.

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -260,16 +259,31 @@ emptyTransactionTableWithSequenceNumbers accs upds =
...
+addTransaction ::
+    -- | Transaction to add.
+    BlockItem ->
+    -- | Commit point of the transaction.
+    CommitPoint ->
+    -- | The associated verification result.
+    TVer.VerificationResult ->
+    -- | An optional "next account nonce" to initialize the @anftNextNonce@ with.
+    --  This MUST be provided for normal transactions.
+    Maybe Nonce ->
```
abizjak, line 271: There seems to be a lot of annoyance caused by this, but is this actually needed?

We seem to use it to
- double check our own work when finalizing transactions. If this fails the node crashes, so we should not need this in production.
- when adding transactions. But there we always do transaction preverification, which ensures that at least the nonce is >= nextNonce for the last finalized block. So the check we have here 

```
tt0 ^. senderANFT . anftNextNonce <= nonce 
```

should never be needed.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -901,9 +933,14 @@ instance
         finTrans WithMetadata{wmdData = NormalTransaction tr, ..} = do
             let nonce = transactionNonce tr
                 sender = accountAddressEmbed (transactionSender tr)
-            anft <- use (skovPersistentData . transactionTable . ttNonFinalizedTransactions . at' sender . non emptyANFT)
-            if anft ^. anftNextNonce == nonce
+            senderMinNonce <- do
+                st <- blockState =<< use (skovPersistentData . lastFinalized)
+                macct <- getAccount st (aaeAddress sender)
+                fromMaybe minNonce <$> mapM (getAccountNonce . snd) macct
+            nextNonce <- use (skovPersistentData . transactionTable . ttNonFinalizedTransactions . at' sender . non (emptyANFTWithNonce senderMinNonce) . anftNextNonce)
```
abizjak, line 940: I think this is a very bad idea, to look up each account during transaction finalization. That's not cheap.

abizjak (changes requested):

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -207,14 +207,13 @@ emptyNFCUWithSequenceNumber = NonFinalizedChainUpdates Map.empty
 --  may also have a non-zero highest commit point if it is received in a block, but that block
 --  is not yet considered arrived (e.g. it is pending its parent).
 --
---  Generally, '_ttNonFinalizedTransactions' should have an entry for every account,
---  with the exception of where the entry would be 'emptyANFT'. Similarly with
---  '_ttNonFinalizedChainUpdates' and 'emptyNFCU'.  In particular, there should be
---  an entry if the next nonce/sequence number is not the minimum value.
+--  The '_ttNonFinalizedTransactions' should have an entry for every account which have non-finalized transactions.
```
td202, line 210:
```suggestion
--  The '_ttNonFinalizedTransactions' should have an entry for every account which has non-finalized transactions.
```

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -207,14 +207,13 @@ emptyNFCUWithSequenceNumber = NonFinalizedChainUpdates Map.empty
...
+--  The '_ttNonFinalizedChainUpdates' should have an entry for all kinds of 'UpdateType's with the exception of where it would be 'emptyNFCU'.
+--  In particular, there should be an entry if the next nonce/sequence number is not the minimum value.
```
td202, line 212: This only applies for chain updates now.

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -260,16 +259,26 @@ emptyTransactionTableWithSequenceNumbers accs upds =
 -- | Add a transaction to a transaction table if its nonce/sequence number is at least the next
 --  non-finalized nonce/sequence number.  A return value of 'True' indicates that the transaction
 --  was added.  The caller should check that the transaction is not already present.
```
td202, line 261: This is no longer accurate, as the nonce is not checked.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -742,20 +741,28 @@ getNonFinalizedCredential txhash sd = do
...
-        case Map.lookupMax (anfts ^. TT.anftMap) of
-            Nothing -> (anfts ^. TT.anftNextNonce, True)
-            Just (nonce, _) -> (nonce + 1, False)
+    m (Nonce, Bool)
+getNextAccountNonce addr sd =
+    fetchFromTransactionTable >>= \case
+        Just ttResult -> return ttResult
+        Nothing -> fetchFromLastFinalizedBlock
+  where
+    fetchFromTransactionTable = return $! TT.nextAccountNonce addr (sd ^. transactionTable)
```
td202, line 760:
```suggestion
    case fetchFromTransactionTable of
        Just ttResult -> return ttResult
        Nothing -> fetchFromLastFinalizedBlock
  where
    fetchFromTransactionTable = TT.nextAccountNonce addr (sd ^. transactionTable)
```

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -774,37 +783,32 @@ finalizeTransactions = mapM_ removeTrans
...
-            ?=! ( anft
-                    & (TT.anftMap . at' nonce .~ Nothing)
-                    & (TT.anftNextNonce .~ nonce + 1)
-                )
-    removeTrans WithMetadata{wmdData = CredentialDeployment{}, ..} =
+        tt <- use transactionTable
+        let mAnft = tt ^? TT.ttNonFinalizedTransactions . ix sender
+        case mAnft of
+            Nothing -> throwM . TreeStateInvariantViolation $ "When finalizing transactions there was no recorded next nonce for sender " <> show sender
+            Just anft -> do
```
td202, line 790:
```suggestion
        anft <- use (transactionTable . TT.ttNonFinalizedTransactions . at' sender . non TT.emptyANFT)
```

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3789,22 +3789,7 @@ cacheStateAndGetTransactionTable ::
...
-            -- cached. (Indeed, 'cache' is defined to be 'pure'.)
-            nonce <- accountNonce acct
-            unless (nonce == minNonce) $ do
-                addr <- accountCanonicalAddress acct
-                MTL.modify
-                    ( TransactionTable.ttNonFinalizedTransactions . at' (accountAddressEmbed addr)
-                        ?~ TransactionTable.emptyANFTWithNonce nonce
-                    )
-            return acct
-    (accts, tt0) <- MTL.runStateT (liftCache perAcct bspAccounts) TransactionTable.emptyTransactionTable
```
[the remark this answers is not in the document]
EmilLa1: Yes

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
```diff
@@ -107,20 +107,35 @@ purgeTables lastFinCommitPoint oldestArrivalTime currentTime TransactionTable{..
...
-        let updptt (Just (Max newHigh)) (Just (low, _))
+            -- Update the pending transaction table.
+            updptt (Just (Max newHigh)) (Just (low, _))
                 | newHigh < low = Nothing
                 | otherwise = Just (low, newHigh)
             updptt _ _ = Nothing
             !ptt1 = ptt0 & pttWithSender . at' addr %~ updptt mmax
-        put (ptt1, trs1)
-        return AccountNonFinalizedTransactions{_anftMap = newANFTMap, ..}
+            anfts' = if null newANFTMap then anfts else HM.insert addr AccountNonFinalizedTransactions{_anftMap = newANFTMap, ..} anfts
```
limemloh, line 137: **Question:** Is this right?

I might be off here, but doesn't `newANFTMap` represent the remaining non-finalized transactions for an account? If so, then when this is empty, we don't update the map, which seems wrong to me.

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
```diff
@@ -175,8 +190,11 @@ purgeTables lastFinCommitPoint oldestArrivalTime currentTime TransactionTable{..
             !ptt1 = ptt0 & pttUpdates . at' uty %~ updptt mmax
         in  (nfcu{_nfcuMap = newNFCUMap}, (ptt1, uis1))
     purge = do
-        -- Purge each account
-        nnft <- HM.traverseWithKey purgeAccount _ttNonFinalizedTransactions
+        -- Purge each account, and possibly remove the @AccountNonFinalizedTransactions@
+        -- if an account does not have any pending transactions left.
+        s <- get
+        let (nnft, s') = HM.foldlWithKey' purgeAccount (HM.empty, s) _ttNonFinalizedTransactions
+        put s'
```
limemloh, line 197: **Minor thing:** Please introduce a variable name instead of just `s` and `s'`.

I don't think it is obvious what this is, and it would save the reader some time instead of having to build up enough context.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -902,33 +920,30 @@ instance
...
-                            logErrorAndThrowTS $ "Tried to finalize transaction which is not known to be in the set of non-finalized transactions for the sender " ++ show sender
+                    -- Remove any other transactions with this nonce from the transaction table.
+                    -- They can never be part of any other block after this point.
+                    forM_ (Map.keys (Map.delete wmdtr nfn)) $
+                        \deadTransaction -> skovPersistentData . transactionTable . ttHashMap . at' (getHash deadTransaction) .= Nothing
+                    -- Mark the status of the transaction as finalized, and remove the data from the in-memory table.
+                    ss <- deleteAndFinalizeStatus wmdHash
+                    -- Remove the transaction from the non finalized transactions.
+                    -- If there are no non-finalized transactions left then remove the entry
+                    -- for the sender in @ttNonFinalizedTransactions@.
```
limemloh, line 935: **Nitpick:** Introduce a function for removing non-finalized transactions from the transaction table.

The code below seems to deal with the internal structure of the transaction table and only behaves correct, since the `_anftNextNonce` is no longer part of the type. I think it would be better to factor this out and keep it close to the type for better 'separation of concerns'.

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
```diff
@@ -450,19 +445,20 @@ reversePTT trs ptt0 = foldr reverse1 ptt0 trs
...
 --  provided account are finalized.
+--  Returns @Nothing@ if no non-finalized transactions were recorded for the provided account.
 nextAccountNonce ::
     -- | The account to look up the next account nonce for.
     AccountAddressEq ->
     -- | The transaction table to look up in.
     TransactionTable ->
     -- | ("the next available account nonce", "whether all transactions from the account are finalized").
-    (Nonce, Bool)
+    Maybe (Nonce, Bool)
```
limemloh, line 455: **Minor thing:** Remove `Bool` from the return type.

It seems the boolean will always be `False` and is therefore not conveying any new information anymore.

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
```diff
@@ -796,15 +794,17 @@ finalizeTransactions = mapM_ removeTrans
...
-        -- Update the non-finalized transactions for the sender
+        -- Remove the transaction from the non finalized transactions.
+        -- If there are no non-finalized transactions left then remove the entry
+        -- for the sender in @ttNonFinalizedTransactions@.
         transactionTable
             . TT.ttNonFinalizedTransactions
             . at' sender
-            ?=! ( anft
+            . non TT.emptyANFT
+            .=! ( anft
```
limemloh, line 804: **Nitpick:** As already mentioned in my other comment related to the transaction table internals, this should probably be factored out.

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
EmilLa1: What do you mean by updating the map? If it's empty then we do not add it to the accummulated `ttNonFinalizedTransactions` Map. If there are still non-finalized transactions, then we add it to the accummulated `ttNonFinalizedTransactions`.

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
EmilLa1: I don't agree, but I added some more words. Hope it's clear now. It's a "complicated"
 state in the sense that it's a tuple where the members each have long names. 
 
But I do think it's already clear when looking at the type signature for `purgeAccount` (which you probably look at anyway) - also the lsp server also tells it.

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
limemloh: ahh, my bad then.
I got confused, thought `anfts` was the current entry value (`AccountNonFinalizedTransactions`) but it was actually the accumulator.

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
EmilLa1: good idea.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
EmilLa1: good idea

concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs
EmilLa1: good idea

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
EmilLa1: ah. Yeah.

limemloh (approved):

td202 (commented):
Mostly minor things. Otherwise, ensure that the account nonce is re-checked before adding incoming transactions (not from blocks), per our discussion.

concordium-consensus/src/Concordium/GlobalState/TransactionTable.hs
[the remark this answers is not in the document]
EmilLa1: It's gone and doc updated.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -880,12 +898,26 @@ instance
...
+                -- because it could be that a block was finalized thus the next account nonce being incremented
+                -- after this transaction was received and pre-verified.
+                isNonceOk <- case wmdData of
+                    NormalTransaction tr -> do
+                        lfbState <- use (skovPersistentData . lastFinalized . to _bpState)
+                        mAcc <- getAccount lfbState $ transactionSender tr
+                        nonce <- maybe (pure minNonce) getAccountNonce (snd <$> mAcc)
+                        return $! nonce <= transactionNonce tr
+                    -- the sequence number will be checked by @Impl.addTransaction@.
+                    _ -> return False
```
td202, line 911:
```suggestion
                    _ -> return True
```

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
```diff
@@ -880,12 +898,26 @@ instance
...
+                if isNonceOk
```
td202, line 912: I would suggest combining the `oldCredential` and `isNonceOk` checks into one check, since they are both casing on the `wmdData`. Actually, it seems very suspicious that we would add a credential deployment to the transaction table where it's already in the on-disk transaction table!!

td202 (approved):
Good job!

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
EmilLa1: good catch.

concordium-consensus/src/Concordium/GlobalState/Persistent/TreeState.hs
EmilLa1: good catch.

concordium-consensus/tests/consensus/ConcordiumTests/EndToEnd/TransactionTableIntegrationTest.hs
```diff
@@ -0,0 +1,178 @@
...
+          bbEpochFinalizationEntry = Absent,
+          bbNonce = computeBlockNonce genesisLEN 4 (bakerVRFKey bakerId),
+          bbTransactions = Vec.fromList [transfer2],
+          bbTransactionOutcomesHash = read "d46c011009b5315c7cd32bb1345bd2e73a3cd6111a7e4d06c33e863f16c8c8bd",
+          bbStateHash = read "a47ca3a8412ad577df94ae8ebc288f8972a499ce5315033bfc2f2c18ce00bfb8"
+        }
+  where
+    bakerId = 3
+
+-- | Test that the anftNextNonce is correctly set when adding a new transaction after
```
abizjak, line 125: This does not appear accurate anymore, seeing there is no `anftNextNonce` anymore.

abizjak (approved):

concordium-consensus/src/Concordium/GlobalState/PurgeTransactions.hs
```diff
@@ -175,8 +190,11 @@ purgeTables lastFinCommitPoint oldestArrivalTime currentTime TransactionTable{..
             !ptt1 = ptt0 & pttUpdates . at' uty %~ updptt mmax
         in  (nfcu{_nfcuMap = newNFCUMap}, (ptt1, uis1))
     purge = do
-        -- Purge each account
-        nnft <- HM.traverseWithKey purgeAccount _ttNonFinalizedTransactions
+        -- Purge each account, and possibly remove the @AccountNonFinalizedTransactions@
+        -- if an account does not have any pending transactions left.
+        pttAndTxTable <- get
+        let (nnft, pttAndTxTable') = HM.foldlWithKey' purgeAccount (HM.empty, pttAndTxTable) _ttNonFinalizedTransactions
+        put pttAndTxTable'
```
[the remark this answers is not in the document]
EmilLa1: Sure. Added  strictness.

PR #1081, opened 2023-11-14
Support "validator" terminology.

### LINKED ISSUE #1080: Support validator terminology in user options.
**Task description**

In order to streamline documentation the node should support the new "validator" terminology for the options that users are expected to use.

In particular this applies to
- baker-credentials
- bakerId inside baker credentials.

The existing options should still work unmodified.

abizjak:
## Purpose

Closes #1080 

Goes together with https://github.com/Concordium/concordium-client/pull/286

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/configuration.rs
+++ b/concordium-node/src/configuration.rs
@@ -360,8 +360,8 @@ impl BakerConfig {
             (Some(bcred), Some(vcred)) => {
                 anyhow::ensure!(
                     bcred == vcred,
-                    "Both --baker-credentials-file and --validator-credentials-file are supplied, \
-                     and different."
+                    "Different --baker-credentials-file and --validator-credentials-file are \
+                     supplied."
                 );
                 bcred
             }
@@ -375,7 +375,7 @@ impl BakerConfig {
         if self.decrypt_credentials() {
             let et = serde_json::from_slice(&read_data)?;
             let pass = rpassword::read_password_from_tty(Some(
-                "Enter password to decrypt baker credentials: ",
+                "Enter password to decrypt validator credentials: ",
             ))?;
             match concordium_base::common::encryption::decrypt(&pass.into(), &et) {
                 Ok(d) => Ok(Some(d)),
--- a/concordium-node/src/configuration.rs
+++ b/concordium-node/src/configuration.rs
@@ -163,7 +163,14 @@ impl PrometheusConfig {
 }
 
 #[derive(StructOpt, Debug)]
-// Parameters related to Baking (only used in cli).
+/// Parameters related to Baking (only used in cli).
+//
+// This structure has a bit of redundancy. In particular it has duplicate
+// options for specifying credentials. The reason for this is that Concordium
+// underwent a renaming of "baker" to "validator". In order to streamline
+// documentation the additional option using "validator" was added, but in order
+// to keep backwards compatibility the existing options that use "baker" needed
+// to remain.
 pub struct BakerConfig {
     #[cfg(feature = "profiling")]
     #[structopt(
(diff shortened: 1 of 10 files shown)


limemloh (approved):
Maybe out of scope for this, but we also mention 'baked blocks' in the prometheus exporter
https://github.com/Concordium/concordium-node/blob/f940006d33c5643525d83936cd1e6f540190872d/concordium-node/src/stats_export_service.rs#L371

concordium-node/src/configuration.rs
```diff
@@ -317,6 +340,56 @@ pub struct BakerConfig {
...
+        let path = match (
+            self.baker_credentials_file.as_ref(),
+            self.validator_credentials_file.as_ref(),
+        ) {
+            (None, None) => return Ok(None),
+            (Some(bcred), Some(vcred)) => {
+                anyhow::ensure!(
+                    bcred == vcred,
+                    "Both --baker-credentials-file and --validator-credentials-file are supplied, \
+                     and different."
```
EmilLa1, line 364:
```suggestion
                    "Different --baker-credentials-file and --validator-credentials-file are supplied"
```

concordium-node/src/configuration.rs
```diff
@@ -317,6 +340,56 @@ pub struct BakerConfig {
...
+            (None, Some(cred)) => cred,
+        };
+        let read_data = match std::fs::read(path) {
+            Ok(read_data) => read_data,
+            Err(e) => anyhow::bail!("Cannot open the validator credentials file ({})!", e),
+        };
+        if self.decrypt_credentials() {
+            let et = serde_json::from_slice(&read_data)?;
+            let pass = rpassword::read_password_from_tty(Some(
+                "Enter password to decrypt baker credentials: ",
```
EmilLa1, line 378:
```suggestion
                "Enter password to decrypt validator credentials: ",
```

EmilLa1 (approved):

service/windows/src/config.rs
```diff
@@ -154,6 +154,25 @@ fn load_config_file(conf_str: &str, conf_root: &Path) -> anyhow::Result<Config>
...
+                        bcred == vcred,
+                        "Both `baker_credentials` and `validator_credentials` are specified, and \
+                         different."
+                    );
+                    Some(bcred)
+                }
+                (baker_credentials, validator_credentials) => {
+                    baker_credentials.or(validator_credentials)
+                }
+            };
```
td202, line 175: I think it would be more intuitive to ignore the `baker_credentials` if `validator_credentials` is supplied.

concordium-node/src/configuration.rs
```diff
@@ -295,18 +302,34 @@ pub struct BakerConfig {
...
     )]
-    pub baker_credentials_file: Option<PathBuf>,
+    baker_credentials_file: Option<PathBuf>,
+    #[structopt(
+        long = "validator-credentials-file",
+        help = "Path to the validator credentials file. If the path is relative it is interpreted \
+                relative to the node process' working directory.",
+        env = "CONCORDIUM_NODE_VALIDATOR_CREDENTIALS_FILE"
+    )]
+    validator_credentials_file: Option<PathBuf>,
```
td202, line 317: I think it would be better to use the aliasing functionality i.e.
```suggestion
    #[structopt(
        long = "validator-credentials-file",
        alias = "baker-credentials-file",
        help = "Path to the validator credentials file. If the path is relative it is interpreted \
                relative to the node process' working directory.",
        env = "CONCORDIUM_NODE_VALIDATOR_CREDENTIALS_FILE"
    )]
    validator_credentials_file: Option<PathBuf>,
```
This doesn't support aliasing the environment variable, but we can just check `CONCORDIUM_NODE_BAKER_CREDENTIALS_FILE` if this option is not set in `read_baker_credentials`. (Similarly for `decrypt-validator-credentials`.)

abizjak: I first thought that, but I think it's clearer that all environment variables are set in one place rather than randomly read somewhere in the code.

service/windows/src/config.rs
abizjak: I think this prevents more stupid errors. If the user supplied both and different they are doing something wrong and I think they should be told this. At least this is what I would want to be told.

concordium-node/src/configuration.rs
abizjak: Good catch.

PR #1089, opened 2023-11-22
Share Account Map Database

### LINKED ISSUE #1088: Node crashes on protocol update (Windows)
**Bug Description**

The node crashes when executing a protocol update. The exact circumstances that cause the crash are unclear, as it does not crash on every protocol update, and different node versions (6.2.1 and 6.2.2) seem to behave differently. However, it does seem to be specific to Windows.

**Steps to Reproduce**

The following crashes have been observed:

*6.2.1:* Starting from a (P6) database containing ~4m accounts constructed by node 6.1.7, execute a P6.Reboot protocol update.

*6.2.2:* Catch up from genesis with mainnet. The crash occurs at the P2->P3 protocol update.

**Expected Result**

The node should successfully execute the protocol update and continue as usual.

**Actual Result**

The node crashes with an access violation (exit code 0xc0000005). This can be observed in the Windows event viewer.

**Versions**
- Software Version: 6.2.1, 6.2.2
- OS: Windows

td202:
## Purpose

Closes #1088

This revises the handling of the account map database so that it is only opened once, and then shared across all protocol versions. This should fix #1088, which appears to be caused by re-opening the account map database, in particular after it has already been resized.

## Changes

The account map `DatabaseHandlers` object is created earlier, and passed in when each consensus instance is started.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix an bug that caused the node to crash on Windows when processing a protocol update.
+
 ## 6.2.2
 
 - The transaction table is only used for tracking next available account nonce
--- a/concordium-consensus/src/Concordium/External.hs
+++ b/concordium-consensus/src/Concordium/External.hs
@@ -39,7 +39,6 @@ import Concordium.GlobalState.Persistent.TreeState (InitException (..))
 import Concordium.MultiVersion (
     Callbacks (..),
     CatchUpConfiguration (..),
-    DiskStateConfig (..),
     MVR (..),
     MultiVersionConfiguration (..),
     MultiVersionRunner (..),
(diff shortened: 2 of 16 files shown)


EmilLa1 (commented):
Looks good. 

Shouldn't the database connection be closed as part of shutting down the consensus? I.e. `shutdownMultiVersionRunner`?

abizjak (approved):
I think it's not critical, but it would still be nice properly shut down the account map database
```
shutdownMultiVersionRunner :: MultiVersionRunner finconf -> IO ()
shutdownMultiVersionRunner MultiVersionRunner{mvConfiguration=MultiVersionConfiguration{..},..} = mask_ $ do
    -- Kill the baker thread, if any.
    forM_ mvBaker $ \Baker{..} -> tryTakeMVar bakerThread >>= mapM_ killThread
    -- Kill the transaction purging thread, if any.
    tryTakeMVar mvTransactionPurgingThread >>= mapM_ killThread
    -- Acquire the write lock. This prevents further updates, as they will block.
    takeMVar mvWriteLock
    versions <- readIORef mvVersions
    -- Close databases for each protocol
    runLoggerT (forM_ versions evcShutdown) mvLog
    -- And then close the global account map database.
    LMDBAccountMap.closeDatabase (globalAccountMap mvcStateConfig)
```
or similar.

This change is good, but I'm a bit annoyed that I cannot reproduce it in any of the ways that are described, nor with much more aggressive database resizing and protocol updates.

PR #1101, opened 2023-11-30
Run queries in dedicated threads

abizjak:
## Purpose

Use a dedicated threadpool for running queries.

Since the queries call into Haskell it typically takes much more than 100us between await points. As a result there are problems with availability where queries pause for some time before resuming. The main reason for this seems to be that all the tokio workers are stuck in the "haskell code" so the scheduler cannot schedule other tasks, such as accepting connections, responding to requests, etc.

The guideline from tokio is that there should not be more than 10-100us in between await points.

> To give a sense of scale of how much time is too much, a good rule of thumb is no more than 10 to 100 microseconds between each .await. That said, this depends on the kind of application you are writing.

(from one of the tokio developers https://ryhl.io/blog/async-what-is-blocking/ )

This is much less than the queries take typically (and queries can also block for a time due to locking).

## Changes

- Run node queries in a dedicated threadpool
- Introduce a configuration option to limit the number of threads used for executing queries in Haskell code.
  This also allows limiting the resources the node is using for non-consensus related operations.
- Default max-concurrent-streams to 200. This is also the default in new releases of the `hyper` server. Without limiting this there seem to occassionally be some blocking issues as well where the queries end up doing nothing. The default value makes the node more robust in the default configuration.
- Update dependencies, mainly tonic & prost to get fixes for some other blocking behaviour.

## Checklist

      hard-to-understand areas.


concordium-node/src/grpc2.rs
```diff
@@ -1223,6 +1252,30 @@ pub mod server {
         }
     }
 
+    impl RpcServerImpl {
+        async fn run_blocking<R: Send + Sync + 'static>(
```
td202, line 1256: Add some documentation.

td202 (approved):
Looks good.

PR #1104, opened 2023-12-05
Include finalization committee and block height in block hash

### LINKED ISSUE #1067: Block hash: Include missing information needed for light client
To reduce the work needed for light client implementations, starting from Protocol version 7 the block hash should include more information:

- Protocol version.
- Block height (Both relative and absolute height).
- Current epoch set of finalizers (Public key and weight).
- Next epoch set of finalizers (Public key and weight).

limemloh:
## Purpose

Closes #1067

The purpose of this PR is to make the block hash to be computed from more information, specifically:

- Weight and BLS verify key of the current and next finalization committee.
- Absolute block height
- Relative block height
- Genesis index.

This PR should be merged after https://github.com/Concordium/concordium-base/pull/490.

## Changes

Starting from protocol version 7:

- The baked block no longer contains the hash of the transaction outcomes and the hash of the block state, instead these are combined into one hash (`BlockResultHash`) which is also computed from the new information listed above.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -182,6 +186,27 @@ finalizerByIndex :: FinalizationCommittee -> FinalizerIndex -> Maybe FinalizerIn
...
+    FinalizationCommitteeHash $
+        LFMBT.hashAsLFMBT emptyCommitteeHash $
+            computeFinalizerInfoHash <$> Vector.toList committeeFinalizers
+  where
+    emptyCommitteeHash = Hash.hash "EmptyFinalizationCommittee"
+    computeFinalizerInfoHash :: FinalizerInfo -> Hash.Hash
+    computeFinalizerInfoHash info = Hash.hashOfHashes weightHash blsKeyHash
+      where
+        weightHash = Hash.hash $ runPut $ put (finalizerWeight info)
+        blsKeyHash = Hash.hash $ runPut $ put (finalizerBlsKey info)
```
abizjak, line 208: The serialization of the BLS key is quite expensive typically.

How often will this be serialized/computed? If it is often we might consider storing a hash of the key since it changes very rarely, and not more than once a day for the committee.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -1229,6 +1272,101 @@ bakeBlock BakeBlockInputs{..} = do
...
+
+-- | Extract information from SkovData and the block state to compute the result block hash.
+computeBlockResultHash ::
+    forall m.
+    ( MonadState (SkovData (MPV m)) m,
+      BlockStateStorage m,
+      BlockState m ~ HashedPersistentBlockState (MPV m),
+      IsConsensusV1 (MPV m)
+    ) =>
+    -- | The new block state for the block.
```
EmilLa1, line 1329: What does `new` here mean? A (frozen) block can only have 1 state.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -1229,6 +1272,101 @@ bakeBlock BakeBlockInputs{..} = do
...
+            )
+            ( Hash.hashOfHashes
+                (blockHeightInfoHash shiBlockHeightInfo)
+                ( Hash.hashOfHashes
+                    (theFinalizationCommitteeHash shiCurrentFinalizationCommitteeHash)
+                    (theFinalizationCommitteeHash shiNextFinalizationCommitteeHash)
+                )
+            )
+  where
+    blockHeightInfoHash BlockHeightInfo{..} = Hash.hash $ Serialize.runPut $ do
```
EmilLa1, line 1316: Should we put a tag on this in the case we ever want to change this?

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -182,6 +186,27 @@ finalizerByIndex :: FinalizationCommittee -> FinalizerIndex -> Maybe FinalizerIn
 finalizerByIndex finCom finInd =
     committeeFinalizers finCom Vector.!? fromIntegral (theFinalizerIndex finInd)
 
+-- | Hash of the finalization committee, only the weight and BLS verify key of each finalizer are used
+-- for computing this.
+newtype FinalizationCommitteeHash = FinalizationCommitteeHash
+    { theFinalizationCommitteeHash :: Hash.Hash
+    }
```
EmilLa1, line 193:
```suggestion
    } deriving newtype (Show, Eq)
```

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -182,6 +186,27 @@ finalizerByIndex :: FinalizationCommittee -> FinalizerIndex -> Maybe FinalizerIn
...
+
+-- | Compute the hash of the finalization committee. Only the weight and BLS verify key of each
+-- finalizer are used for computing this.
+computeFinalizationCommitteeHash :: FinalizationCommittee -> FinalizationCommitteeHash
```
EmilLa1, line 197: I think a unit test of this would be nice. Maybe even one that shows that some validator with a certain key and weight is in the committee?

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
EmilLa1: Why not just cache a hash of the whole finalization committee? The finalization committee itself does not change in between paydays.

limemloh: > How often will this be serialized/computed?

I did not expect this to be expensive, but to answer the question, this will be computed every time a block is baked or processed by another node.

> Why not just cache the hash of the whole finalization committee?

I did consider this as well during development, but did not consider this computation expensive enough to be worth it.
I have now changed the behavior to be caching it, but put the cache as part of the type `BakersAndFinalizers` instead, since the `FinalizationCommittee` type seems to be used as part of timeout network messages.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/TreeStateTest.hs
```diff
@@ -354,9 +385,8 @@ toStoredBlock BlockPointer{..} =
 --
 --  * 'testB' (alive)
 --  * 'focusB' (parent is 'testB')
---  * 'pendingB' (pending)
 --  * the block indicated by 'deadH' has added to the dead cache.
-skovDataWithTestBlocks :: SkovData pv
+skovDataWithTestBlocks :: SkovData 'P6
```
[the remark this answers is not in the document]
limemloh: I have isolated the changes needed for P7 and onwards in this PR https://github.com/Concordium/concordium-node/pull/1111

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -1229,6 +1272,99 @@ bakeBlock BakeBlockInputs{..} = do
...
+    return $
+        makeBlockResultHash
+            BlockResultHashInput
+                { shiBlockStateHash = theBlockStateHash,
+                  shiTransationOutcomesHash = transactionOutcomesHash,
+                  shiCurrentFinalizationCommitteeHash = currentFinalizationCommitteeHash,
+                  shiNextFinalizationCommitteeHash = nextFinalizationCommitteeHash,
+                  shiBlockHeightInfo = blockHeightInfo
+                }
+
```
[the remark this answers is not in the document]
limemloh: I moved the types and the `makeBlockResultHash` function, but not the `computeBlockResultHash` since it relies on context, which is only relevant for this file.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -1229,6 +1272,101 @@ bakeBlock BakeBlockInputs{..} = do
...
+    { -- | The absolute height of the block.
+      bhiAbsoluteBlockHeight :: AbsoluteBlockHeight,
+      -- | The genesis index of the block.
+      bhiGenesisIndex :: !GenesisIndex,
+      -- | The relative block height from the genesis prior to this block.
+      bhiRelativeBlockHeight :: !BlockHeight
+    }
+
+-- | Compute the block result hash given the result hash input.
+makeBlockResultHash :: BlockResultHashInput -> BlockResultHash
```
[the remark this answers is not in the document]
limemloh: Converting the tests to run on P7 will test this (at least implicitly), do you think we also need some unit test?
Not sure whether this would catch anything interesting

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
limemloh: I imagine this would fit better as a test using the merkle proofs introduce in another PR by Thomas.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/LMDB.hs
```diff
@@ -86,7 +87,7 @@ dummyBlockSig = Block.sign dummyKP "someMessage"
 -- | A helper function for creating a 'BakedBlock' given a round. Used by 'dummyBlock' to create blocks.
 --  The block is well-formed and contains the supplied transactions and is for the specified round.
 --  Beyond that, there should be no expectation on the data in the block.
-dummyBakedBlock :: Round -> Vector.Vector BlockItem -> BakedBlock
+dummyBakedBlock :: Round -> Vector.Vector BlockItem -> BakedBlock 'P6
```
abizjak, line 90: Is there a reason why we don't have one such for P7?

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -1,4 +1,6 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MonoLocalBinds #-}
```
abizjak, line 3: THis is implied by GADTs.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -49,6 +51,8 @@ import Concordium.Startup
 import Concordium.TimerMonad
 import Concordium.Types.Option
 
+type PV = 'P6
```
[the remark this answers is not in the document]
abizjak: I assume this is also resolved in the related PR.

abizjak (approved):

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -1301,3 +1389,49 @@ newtype QuorumCertificateCheckedWitness = QuorumCertificateCheckedWitness Epoch
...
+      -- | The finalization committee hash for the next epoch.
+      shiNextFinalizationCommitteeHash :: FinalizationCommitteeHash,
+      -- | The block height information of this block.
+      shiBlockHeightInfo :: BlockHeightInfo
+    }
+
+-- | The block height information of a block.
+data BlockHeightInfo = BlockHeightInfo
+    { -- | The absolute height of the block.
+      bhiAbsoluteBlockHeight :: AbsoluteBlockHeight,
```
EmilLa1, line 1410:
```suggestion
      bhiAbsoluteBlockHeight :: !AbsoluteBlockHeight,
```

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -1301,3 +1389,49 @@ newtype QuorumCertificateCheckedWitness = QuorumCertificateCheckedWitness Epoch
...
+    { -- | Hash of the block state.
+      shiBlockStateHash :: StateHash,
+      -- | Hash of the transaction outcomes.
+      shiTransationOutcomesHash :: TransactionOutcomesHash,
+      -- | The finalization committee hash for the current epoch.
+      shiCurrentFinalizationCommitteeHash :: FinalizationCommitteeHash,
+      -- | The finalization committee hash for the next epoch.
+      shiNextFinalizationCommitteeHash :: FinalizationCommitteeHash,
+      -- | The block height information of this block.
+      shiBlockHeightInfo :: BlockHeightInfo
```
EmilLa1, line 1404: Any specific reason that these are not strict?

EmilLa1 (approved):
Great job! 

I presume that the things mentioned will be taken care of by https://github.com/Concordium/concordium-node/pull/1111 and future PRs.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
limemloh: No, I'll change them to be strict :+1:

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/LMDB.hs
limemloh: This is now generalized (as part of #1111)

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Consensus/Blocks.hs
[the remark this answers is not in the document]
limemloh: After the PR #1111 this is only used in one place, so this is now inlined instead.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
```diff
@@ -1229,6 +1267,53 @@ bakeBlock BakeBlockInputs{..} = do
...
+        -- information in the block state.
+        currentEpoch <- use (roundStatus . rsCurrentEpoch)
+        nextSkovBakersAndFinalizers <- gets (getBakersForEpoch (currentEpoch + 1))
+        case nextSkovBakersAndFinalizers of
+            Just bakersAndFinalizers -> return $ bakersAndFinalizers ^. bfFinalizerHash
+            Nothing -> do
+                nextFullBakers <- getNextEpochBakers newState
+                nextFinalizationParameters <- getNextEpochFinalizationCommitteeParameters newState
+                let nextFinalizationCommittee = computeFinalizationCommittee nextFullBakers nextFinalizationParameters
+                return $ computeFinalizationCommitteeHash nextFinalizationCommittee
```
td202, line 1298: This relies on the epoch of the block matching the current epoch (as in `roundStatus . rsCurrentEpoch`). This should be documented as a precondition of the function, and we should ensure that it holds true everywhere it's used.

td202: Actually, I'm not entirely sure that this is guaranteed to be the case in `processBlock`. Instead, I would suggest passing in the epoch of the block and then using `getBakersForEpoch` on that, falling back to computing them from the block state.

td202 (approved):
Looks good. Just `computeBlockResultHash` is a concern for me.

concordium-consensus/src/Concordium/KonsensusV1/Consensus/Blocks.hs
limemloh: I have changed this and also remove the use of `bakersForCurrentEpoch` above this.

PR #1105, opened 2023-12-06
Merkelize BakerInfos

### LINKED ISSUE #1069: Block hash: Merkelize BakerInfos
Starting from protocol version 7, the hash of baker infos for current epoch and next epoch, should be done as a merkelized binary tree as this will allow for smaller merkle proofs.

limemloh:
## Purpose

Closes #1069.

## Changes

Change type parameter for `BakerInfo` from account version to protocol version, and change the hashing of this type based on the protocol version.


concordium-consensus/src/Concordium/GlobalState/Persistent/Bakers.hs
```diff
@@ -72,10 +74,14 @@ instance (IsAccountVersion av, MonadBlobStore m) => BlobStorable m (BakerInfos a
...
-        v' <- mapM loadPersistentBakerInfoRef v
-        return $ H.hashLazy $ runPutLazy $ mapM_ put v'
+instance forall pv m. (IsProtocolVersion pv, MonadBlobStore m) => MHashableTo m H.Hash (BakerInfos pv) where
+    getHashM (BakerInfos infos) = do
+        loadedInfos <- mapM loadPersistentBakerInfoRef infos
+        case sBlockHashVersionFor (protocolVersion @pv) of
+            SBlockHashVersion0 -> do
+                return $ H.hashLazy $ runPutLazy $ mapM_ put loadedInfos
+            SBlockHashVersion1 -> do
+                return $ hashAsLFMBTV1 (H.hash "NoBakerInfos") $ H.hashLazy . runPutLazy . put <$> Vec.toList loadedInfos
```
abizjak, line 84: I'm not sure if this hashLazy here is beneficial. Isn't a bakerinfo quite a small structure? So it's probably better to just serialize and then do one hash FFI call.

abizjak (approved):

td202 (approved):

EmilLa1 (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/Bakers.hs
limemloh: Not sure when to consider it a small structure.
It includes the baker ID, all the baker public keys and the pool info which is open status, metadata url and commission rates.
I can change it if you want

limemloh: In order to know what is actually best, we need benchmarks, for now this will most likely perform similar to before, so I won't do anything.

PR #1110, opened 2023-12-14
Revise hashing to support Merkle proofs

td202:
## Purpose

This involves several changes to support hashing blocks to support Merkle proofs. Primarily, these involves updating how `LFMBTree` hashes are computed for the new P7 protocol version.

## Changes

- New hashing scheme for LFMBTrees, which includes the size of the tree at the root.
- `CapitalDistribution` is hashed differently depending on `BlockHashVersion`, which determines the LFMBTree hashing scheme to use.
- `PoolRewards` parametrised by `BlockHashVersion`. The next and current capital distribution hashes are dependent on the `BlockHashVersion`, as well as the hashing of the `bakerPoolRewardDetails` LFMBTree.
- As the majority of `Concordium.GlobalState.Basic.BlockState.PoolRewards` was not used, the remaining useful definitions are lifted to `Concordium.GlobalState.PoolRewards`.
- Introduce new `AccountsHash`, `ModulesHash` and `InstancesHash` types for the accounts, modules and instances tables, which are parametrised by the protocol version.
- A lot of legacy code for the in-memory instances/instance table is removed. The tests on the instances table are revised to use the persistent version.
- Accounts hashing revised to use the new LFMBTree hashing scheme in P7.
- `BlockRewardDetails` is revised to parametrise by the `BlockHashVersion`.
- `TransactionOutcomes` are moved from base into a new `Concordium.Types.TransactionOutcomes`.
- `PersistentTransactionOutcomes` supports a new `TransactionOutcomesVersion` that uses the new LFMBTree hashing scheme. A `TransactionOutcomesHashV` that is parametrised by the `TransactionOutcomesVersion` is introduced for hashing the `PersistentTransactionOutcomes`. `TransactionOutcomesHash` is still used in places where the version parameter is inconvenient or unnecessary.
- Module table hashing revised to use the new LFMBTree hashing scheme in P7.
- Changes to the instances table:
  - Fix a bug in `newContractInstanceIT` when inserting in a table with vacancies.
  - Change the instances table to track the total number of leaves including vacancies, instead of just the non-vacant leaves. This is a non-breaking change since it has never yet been possible to delete an instance, and so these numbers have historically been the same. This change affects the behaviour of `newContractInstance` and `deleteContractInstance`.
  - Change the hashing scheme at P7 to incorporate the number of leaves in the root hash (as for the LFMBTree).
- ConsensusV1 block types are now parametrised by the protocol version.
- Partial implementation of `MerkleProvable` for `BakedBlock`, supporting deriving Merkle proofs about a range of block attributes.

## Checklist

      hard-to-understand areas.


limemloh (approved):
Only very minor comments

EmilLa1 (commented):
Looks good :+1: 

I think we should add some tests for the provable properties of the block.

concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/LFMBTree.hs
```diff
@@ -100,6 +119,27 @@ data T v
     | Leaf !v
     deriving (Eq, Show)
 
+-- | Hash representation type for an LFMBTree. This is parametrised by the block hash version.
```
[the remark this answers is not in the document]
td202: 'Parametrised' is a generally accepted spelling, and is the one I prefer and is used quite a lot in the code base. 'Parameterized' is also used in the code, but is less frequent. I don't think we should introduce a third spelling.

concordium-consensus/src/Concordium/GlobalState/Persistent/Instances.hs
```diff
@@ -665,15 +684,25 @@ emptyInstances :: Instances pv
...
+    (res, newInst) <- createInstanceFn ca
     (res,) . InstancesTree 1 <$> membed (Leaf newInst)
-newContractInstance fnew (InstancesTree s it) = (\(res, it') -> (res, InstancesTree (s + 1) it')) <$> newContractInstanceIT fnew it
+newContractInstance createInstanceFn (InstancesTree size tree) = do
+    ((isFreshIndex, !result), !nextTree) <- newContractInstanceIT createFnWithFreshness tree
+    let !nextSize = if isFreshIndex then size + 1 else size
+        nextInstancesTree = InstancesTree nextSize nextTree
+    return (result, nextInstancesTree)
+  where
+    createFnWithFreshness newContractAddress = do
```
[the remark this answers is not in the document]
td202: Yes. The function returns whether the contract index is "fresh" (i.e. not previously used by any contract).

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -1226,24 +1250,73 @@ instance HashableTo BlockQuasiHash (BakedBlock pv) where
...
+                put bbTimestamp
+                put bbBaker
+        let nonce = optProof (biPath ++ ["nonce"]) . rawMerkle . encode $ bbNonce
+        let bakerInfo = optProof biPath [timestampBaker, nonce]
+        let chPath = ["quasi", "meta", "certificatesHash"]
+        let qcPath = chPath ++ ["quorumCertificate"]
+        qcMerkleProof <- Merkle.buildMerkleProof (open . (qcPath ++)) bbQuorumCertificate
+        let qcHash = optProof qcPath qcMerkleProof
+        let tfPath = chPath ++ ["timeoutFinalization"]
+        let tcPath = tfPath ++ ["timeoutCertificate"]
```
abizjak, line 1297: Is it intended that the timeout certificate is under timeoutFinalization?

concordium-consensus/src/Concordium/Types/TransactionOutcomes.hs
```diff
@@ -0,0 +1,146 @@
...
+{-# NOINLINE emptyTransactionOutcomesHashV1 #-}
+emptyTransactionOutcomesHashV1 =
+    TransactionOutcomesHashV $
+        H.hashShort
+            ( "TransactionOutcomesHashV1"
+                <> H.hashToShortByteString (H.hash "EmptyLFMBTree")
+                <> H.hashToShortByteString (H.hash "EmptyLFMBTree")
+            )
+
+-- | Hash of the empty V1 transaction outcomes structure. This transaction outcomes
```
abizjak, line 115:
```suggestion
-- | Hash of the empty V2 transaction outcomes structure. This transaction outcomes
```

concordium-consensus/tests/globalstate/GlobalStateTests/Instances.hs
```diff
@@ -59,33 +68,157 @@ validContractArtifactsV1 = mapMaybe packModule contractSourcesV1
...
-invariantIT offset (Branch h f v hsh l r) = do
-    (hl, fl, vl, offset', hshl, cl) <- invariantIT offset l
+-- | Assert that a binary predicate holds.
+checkBinary :: (Show a, MonadFail m) => (a -> a -> Bool) -> a -> a -> String -> String -> String -> m ()
+checkBinary bop x y sbop sx sy =
+    unless (bop x y) $
+        fail $
+            "Not satisfied: " ++ sx ++ " (" ++ show x ++ ") " ++ sbop ++ " " ++ sy ++ " (" ++ show y ++ ")"
+
+-- | Check an invariant on 'Instances.IT'. The return value is a tuple consisting of:
```
abizjak, line 78: What invariant is it checking?

concordium-consensus/tests/globalstate/GlobalStateTests/Instances.hs
```diff
@@ -59,33 +68,157 @@ validContractArtifactsV1 = mapMaybe packModule contractSourcesV1
...
-    return (succ h, f, v, offset'', hsh, cl + cr)
-
-invariantInstanceTable :: InstanceTable -> Either String ()
-invariantInstanceTable Empty = Right ()
-invariantInstanceTable (Tree c0 t) = do
-    (_, _, _, _, _, c) <- invariantIT 0 t
-    checkBinary (==) c0 c "==" "reported number of instances" "actual number"
+    return (succ h, f, v, offset'', hsh)
+
+-- | Check an invariant on 'Instances.Instances'.
```
abizjak, line 106: What is the invariant?

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -1226,24 +1250,73 @@ instance HashableTo BlockQuasiHash (BakedBlock pv) where
...
-instance HashableTo BlockHash (BakedBlock pv) where
-    getHash bb = computeBlockHash (getHash bb) (getHash bb)
+-- | Compute the block hash from the header hash and quasi-hash.
+computeBlockHash :: BlockHeaderHash -> BlockQuasiHash' bhv -> BlockHash
+computeBlockHash bhh bqh = computeBlockHash' bhh (theBlockQuasiHash bqh)
+
+instance (IsProtocolVersion pv) => HashableTo BlockHash (BakedBlock pv) where
+    getHash bb = computeBlockHash @(BlockHashVersionFor pv) (getHash bb) (getHash bb)
+
+instance (Monad m, BlockHashVersionFor pv ~ 'BlockHashVersion1) => Merkle.MerkleProvable m (BakedBlock pv) where
```
abizjak, line 1280: I might have missed them, but I can't seem to find any unit tests for this implementation.

I would expect a worked example or two coded as a unit test so we have something stable to compare things to when making changes.

concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/LFMBTree.hs
```diff
@@ -284,28 +341,85 @@ fromAscListMaybes l = fromList $ go l 0
...
+--  the number of leaves in the tree and the V0 hash.
+hashAsLFMBTV1 ::
+    -- | Hash to use for empty list
+    H.Hash ->
+    -- | List of hashes to construct into Merkle tree
+    [H.Hash] ->
+    H.Hash
+hashAsLFMBTV1 e l = H.hashLazy $! S.runPutLazy $ do
+    S.putWord64be (fromIntegral $ length l)
+    S.put $ hashAsLFMBTV0 e l
```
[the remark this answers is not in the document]
td202: It is indeed redundant to include the size in each subtree. The only reason I can think of that it might be useful would be in decomposing a Merkle proof for an LFMBTree, but even then I am not sure it would add any benefit. The downside would be that Merkle proofs would become longer, and probably more complex to check and construct.

I have added a comment elsewhere explaining the benefit of having the size at the root.

concordium-consensus/src/Concordium/GlobalState/Persistent/PoolRewards.hs
```diff
@@ -68,12 +68,13 @@ data PoolRewards = PoolRewards
 
 -- | Migrate pool rewards from @m@ to the new backing store @t m@.
 --  This takes the new next payday epoch as a parameter, since this should always be updated on
---  a protocol update.
+--  a protocol update. This does not allow the hashing scheme to change in the migration, and
+--  thus can reuse hashes.
```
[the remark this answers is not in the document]
td202: The idea is that the capital distribution is potentially a large structure, and so it's wasteful to recompute the hash on migration if it's not changed. However, it's probably a relatively small effort in the overall migration, so I'll just merge these.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
td202: Yes. timeoutFinalization combines the timeout certificate and epoch finalization entry.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
td202: Added tests.

abizjak (approved):
I think this suffers from some of the same as the related PR in base.

Ideally this would be 2 pull requests, one changing the hashing scheme, and the other one doing the Merkle proof production.

EmilLa1 (approved):

PR #1111, opened 2023-12-18
Convert P6 tests to also run for P7 and onwards

limemloh:
## Purpose

Prior to this PR, numerous of our tests are only targeting protocol version 6, this PR changes this such that these tests are generalized to run against every protocol version starting from P6 and onwards.

## Changes

- No changes are intended by this PR besides generalizing the tests, the only exception is a test related to account nonce https://github.com/Concordium/concordium-node/pull/1111/files#diff-176f6e3e32412c792f8e0873ce966e99076d1be8da6a228d9226d6b87301fbefR760

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Common.hs
+++ b/concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Common.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
 
 -- | A common module for various helper definitions used for testing purposes.
 module ConcordiumTests.KonsensusV1.Common where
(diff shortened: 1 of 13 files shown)


abizjak (approved):
It's a bit annoying to review this since a lot of the changes are just whitespace changes.

But looks good.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Common.hs
```diff
@@ -77,3 +135,34 @@ sigKeyPair = fst $ Dummy.randomBlockKeyPair $ mkStdGen 42
...
 sigPublicKey = Sig.verifyKey sigKeyPair
+
+-- | Run tests for each protocol version.
+-- The tests are provided with the protocol version and a string representation of the protocol version.
+forEveryProtocolVersion ::
+    (forall pv. (IsProtocolVersion pv) => SProtocolVersion pv -> String -> Spec) ->
+    Spec
+forEveryProtocolVersion check =
+    sequence_
+        [ check SP1 "P1",
```
EmilLa1, line 146: Why don't we just use the `Show` instance for the `ProtocolVersion` ?

EmilLa1 (approved):
Looks good, great job! :+1:

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Common.hs
limemloh: If found this simpler, it becomes something like this then:
```suggestion
        [ check SP1 (show $ demoteProtocolVersion SP1),
```
I don't mind changing it, if you prefer this.

PR #1113, opened 2024-01-09
Include the protocol version in the block hash from P7 onwards.

### LINKED ISSUE #1112: Incorporate protocol version in block hash (P7 onwards)
The block hash scheme should be modified to incorporate the protocol version at the root of the hashing scheme in protocol version 7.

td202:
## Purpose

Closes #1112

This adds the protocol version to the root of the block hash in protocol version 7 onwards.

## Changes

- `computeBlockHash` and `computeBlockHash'` are revised to take the protocol version, and use it in the hashing scheme from P7 onwards.
- Since the block hashing is changed, the deserialization of a `FinalizationEntry` becomes dependent on the protocol version. This is since the block hash of the successor block is computed from other information in the finalization entry. To deal with this, `FinalizationEntry` is now parametrised by a protocol version.
- A new `ProtoFinalizationEntry` type is introduced that is not parametrised by the protocol version. This can be converted to a `FinalizationEntry` given the specific protocol version. This type is used instead of `FinalizationEntry` in catch-up messages, since these messages are handled in a way that is independent of the protocol version.

## Checklist

      hard-to-understand areas.
- [ ] (If necessary) I have updated the CHANGELOG.s


limemloh (approved):
LGTM, but I don't think the tests compile currently.

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
```diff
@@ -412,45 +444,73 @@ data FinalizationEntry = FinalizationEntry
...
+
+-- | Serialize a 'FinalizationEntry'. This is provided separately from the 'Serialize' instance
+--  since it does not depend on the @IsProtocolVersion pv@ constraint.
+putFinalizationEntry :: FinalizationEntry pv -> Put
+putFinalizationEntry = put . toProtoFinalizationEntry
+
+-- | Convert a 'ProtoFinalizationEntry' to a 'FinalizationEntry'. This depends on the protocol
+--  version for deriving the hash of the successor block from the 'ProtoFinalizationEntry' data.
+toFinalizationEntry' :: forall pv. SProtocolVersion pv -> ProtoFinalizationEntry -> FinalizationEntry pv
+toFinalizationEntry' spv ProtoFinalizationEntry{..} =
```
abizjak, line 466: Why do we need this (where the `'` in the name does not mean anything) as well as the `toFinalizationEntry` one?

Coudl we at least name them in a more informative way.

abizjak (approved):

EmilLa1 (approved):
looks good

concordium-consensus/src/Concordium/KonsensusV1/Types.hs
td202: Usually, having the `IsProtocolVersion` constraint it preferable, but in some cases we have an `SProtocolVersion` instead (e.g. with `getFinalizationEntry`). It isn't convenient to go from the latter to the former, so we provide both versions.

PR #1114, opened 2024-01-11
Update Haskell stackage snapshot to LTS-22.9

### LINKED ISSUE #1093: Upgrade GHC
**Task description**

Upgrade GHC version.

Currently we use lts-20.16 (GHC 9.2.7) which is fairly old. We should upgrade to a newer lts version. As of this writing then lts-21.22 (GHC 9.4.8) is the most recent one.

Components which should be upgraded:

- [x] Node
- [x] Base
- [ ] Client
- [ ] Wallet proxy
- [ ] Tokenomics reporter

td202:
## Purpose

Closes #1093

This upgrades the stack resolver to LTS-22.9, and the corresponding GHC version to 9.6.4.

## Changes

- Resolver is update to lts-22.9 in stack.yaml files.
- The package `proto-lens-setup` and its dependencies `proto-lens-protoc` and `ghc-source-gen` are not included in the LTS distribution, so specific versions from Hackage are pinned in the stack.yaml files.
- The `-fPIC` compiler option is disabled on Windows, as it results in the compiled code crashing due to access violations. Position independent code is not used on Windows in any case. There is an open GHC issue that seems to be related: https://gitlab.haskell.org/ghc/ghc/-/issues/24016
- `TypeOperators` is added to the default extensions. In future, this will be needed for using `~`, and currently will generate a warning if not enabled.
- Some imports are changed due to changes in library structure (notably, `Control.Monad` is explicitly imported, whereas it previously was implicitly imported through other monad modules).
- Some code changes make types more explicit, e.g. in (type-changing record updates)[https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.6#type-changing-record-updates-involving-type-families].
- In some cases where redundant constraint warnings were suppressed, this has been lifted. In other cases, the reason why they are suppressed is updated.

## Checklist

      hard-to-understand areas.


concordium-consensus/package.yaml
```diff
@@ -91,19 +91,74 @@ library:
...
+        ghc-options: -fPIC
+      else
+        -- The foreign library should not be built when doing a static build of the node.
+        buildable: False
+    
+    other-modules:
+      Concordium.External
+      Concordium.External.DryRun
+      Concordium.External.Helpers
+      Concordium.External.GRPC2
```
abizjak, line 133: Can you please add an explanation to the README about where these modules are located and why.

abizjak (approved):
Had some problems with installing ghc 9.6.3 which were "fixed" by restarting the build a few times.

Not very reassuring, but seems to work.

EmilLa1 (approved):

.github/workflows/build-test.yaml
```diff
@@ -128,6 +128,10 @@ jobs:
         - rust: 1.68
 
     steps:
+    - name: Remove unnecessary files
+      run: |
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```
abizjak, line 134: Wouldn't it better to just invalidate the cache?

That's probably the reason for these problems since both the old and new toolchains are now cached.

concordium-consensus/package.yaml
td202: Done.

.github/workflows/build-test.yaml
abizjak: I had the same issue in https://github.com/Concordium/concordium-node/pull/1118 which was fixed by clearing the cache of old junk.

PR #1137, opened 2024-04-02
Support Rust 1.73 on Windows.

td202:
## Purpose

Support Rust version 1.73 on Windows. This changes how the Haskell code is linked against Rust libraries to use dynamic (rather than static) linking on Windows.

## Changes

- The `smart_contract_execution_engine` static library is no longer copied to `lib`, and is removed if already present.
- The `smart_contract_execution_engine.dll` is copied to the install directory in a post-copy hook, to ensure that it is available when binaries and tests are run.
- The DLLs included in the installer are changed to include the DLLs `sha_2.dll`, `concordium_base.dll` and `smart_contract_execution_engine.dll`. `libunwind.dll` is no longer required.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/utils.rs
+++ b/concordium-node/src/utils.rs
@@ -9,14 +9,6 @@ use std::{
     path::Path,
 };
 
-pub fn to_hex_string(bytes: &[u8]) -> String {
-    use std::fmt::Write;
-    bytes.iter().fold(String::new(), |mut output, b| {
-        let _ = write!(output, "{:02x}", b);
-        output
-    })
-}
-
 /// Setup a log4rs logger based on the given configuration file.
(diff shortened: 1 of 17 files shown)


concordium-node/src/utils.rs
```diff
@@ -10,7 +10,11 @@ use std::{
 };
 
 pub fn to_hex_string(bytes: &[u8]) -> String {
```
abizjak, line 12: Just remove this function. It is not used anywhere. It's some leftover from times where we did not use types from base for hashes.

abizjak (approved):

limemloh (approved):

PR #1138, opened 2024-04-07
Integrate new cost assignment version.

abizjak:
## Purpose

Integrate reduce costs for P7.

This is a very minor change compared to #1135 . The main annoyance is that tests needed to be updated because of the both cost changes in P7 and parametrization of module lookup by protocol version.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/Scheduler.hs
@@ -1103,6 +1103,7 @@ checkAndGetBalanceInstanceV0 ownerAccount istance transferAmount = do
 handleContractUpdateV1 ::
     forall r m.
     (StaticInformation m, AccountOperations m, ContractStateOperations m, ModuleQuery m, MonadProtocolVersion m) =>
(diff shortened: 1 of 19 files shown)


concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -1102,6 +1103,7 @@ checkAndGetBalanceInstanceV0 ownerAccount istance transferAmount = do
 handleContractUpdateV1 ::
     forall r m.
     (StaticInformation m, AccountOperations m, ContractStateOperations m, ModuleQuery m, MonadProtocolVersion m) =>
+    Word ->
```
limemloh, line 1106: Please add doc comment for this

limemloh (approved):

concordium-consensus/src/Concordium/Scheduler.hs
abizjak: Added

Issue #1057, opened 2023-10-25
Change log level for grpc not_found queries


abizjak:
**Description**

Currently the node logs

```
[ERROR tower_http::trace::on_failure] response failed classification=Code: 5 latency=0 ms
```
when for example the GetAccountInfo is queried where either the block or account do not exist.

It is not ideal to use `ERROR` level for this, since it is a user/request problem. 

We should change log levels for request failures that are the client's fault to not be ERROR, and just be normal info logs.

I think this is logged by the layer
```
                let log_layer = tower_http::trace::TraceLayer::new_for_grpc();
```

which should be updated to change the `.on_failure` classifier.

**Versions**
- Software Version: 4.5.0+

PR #1169, opened 2024-05-14
Change severity of logs for certain gRPC status codes

limemloh:
## Purpose

Fixes https://github.com/Concordium/concordium-node/issues/1057

## Changes

- Using DEBUG severity for logging of failed gRPC API requests.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/src/grpc2.rs
+++ b/concordium-node/src/grpc2.rs
@@ -1087,20 +1087,19 @@ pub mod server {
                 });
                 let service = service::queries_server::QueriesServer::new(server);
                 let log_layer = {
-                    use tower_http::classify::GrpcCode as Code;
-                    // Change the trace level for gRPC status codes
-                    let classifier = tower_http::classify::GrpcErrorsAsFailures::new()
-                        .with_success(Code::Cancelled)
-                        .with_success(Code::InvalidArgument)
-                        .with_success(Code::NotFound)
-                        .with_success(Code::AlreadyExists)
-                        .with_success(Code::PermissionDenied)
-                        .with_success(Code::Unimplemented)
-                        .with_success(Code::Unavailable)
-                        .with_success(Code::Unauthenticated);
-                    tower_http::trace::TraceLayer::new(tower_http::classify::SharedClassifier::new(
-                        classifier,
-                    ))
+                    #[derive(Default, Copy, Clone)]
+                    struct OnFailure;
+                    impl tower_http::trace::OnFailure<tower_http::classify::GrpcFailureClass> for OnFailure {
+                        fn on_failure(
+                            &mut self,
+                            failure: tower_http::classify::GrpcFailureClass,
+                            _latency: std::time::Duration,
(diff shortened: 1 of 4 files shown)


concordium-node/src/grpc2.rs
```diff
@@ -1086,7 +1086,22 @@ pub mod server {
                     }
                 });
                 let service = service::queries_server::QueriesServer::new(server);
-                let log_layer = tower_http::trace::TraceLayer::new_for_grpc();
+                let log_layer = {
+                    use tower_http::classify::GrpcCode as Code;
+                    // Change the trace level for gRPC status codes
+                    let classifier = tower_http::classify::GrpcErrorsAsFailures::new()
+                        .with_success(Code::Cancelled)
```
lassemand, line 1093: It is not clear to me That Code::Ok is per default added to this list. Can you just add it?

concordium-node/src/grpc2.rs
```diff
@@ -1086,7 +1086,22 @@ pub mod server {
...
+                        .with_success(Code::InvalidArgument)
+                        .with_success(Code::NotFound)
+                        .with_success(Code::AlreadyExists)
```
lassemand, line 1096: RESOURCE_EXHAUSTED can be argued both places. We actually have user limits right? So I would say that maybe also add that. Could that perhaps be made something like a warning? (Could also indicate a try on a DoS attack)

lassemand (approved):
Looks good to me.

I believe it would make more sense to also include RESOURCE_EXHAUSTED but I can also argue for why it should not be there. I do believe this one is usually cast when the node is gracefully exhausted. (Else it would be down?)

If a person tries to DoS us what should we do? On the one hand it would be nice to provide the error logs, but on the other hand, the server is not in a bad state, only the client is unable to request the server.

I can accept both.

concordium-node/src/grpc2.rs
```diff
@@ -1086,7 +1086,22 @@ pub mod server {
...
+                        .with_success(Code::PermissionDenied)
+                        .with_success(Code::Unimplemented)
+                        .with_success(Code::Unavailable)
+                        .with_success(Code::Unauthenticated);
+                    tower_http::trace::TraceLayer::new(tower_http::classify::SharedClassifier::new(
+                        classifier,
+                    ))
```
td202, line 1103: Possibly we should set the logging for failures lower than ERROR. I would suggest even DEBUG is high enough. Reasons:

- Generally, failure codes don't indicate serious errors.
- If a serious error in processing a request occurs, usually there's an ERROR log message generated anyway.
- The log messages generated on failures look like `[ERROR tower_http::trace::on_failure] response failed classification=Code: 12 latency=0 ms` - which is not really very useful.

limemloh: I have now changed it to make every error status code be logged at debug level.

td202 (approved):
Looks good!

PR #1198, opened 2024-06-17
Hotfix 6.3.1

### LINKED ISSUE #1197: Nodes fail to produce a Timeout Certificate
**Bug Description**

Under certain circumstances, when a timeout occurs at the start of a new epoch, nodes may fail to produce a Timeout Certificate, despite having received sufficient Timeout Messages to do so. This is caused by incorrectly calculating the total voting weight of the finalizers that have signed, when the finalizers signed timeout messages across two different epochs.

This bug can cause the consensus to stall (in uncommon circumstances) if all nodes are affected, since none of them will produce a Timeout Certificate and so the validators will not progress to the next round.

**Versions**
- Software Version: 6.* =< 6.3.0
- OS: All

td202:
## Purpose

Fixes #1197

This fixes a bug in the calculation of the weight of finalizers that have signed timeout messages when deciding whether a timeout certificate can be constructed.

## Changes

- Factor out and revise the finalizer weight calculation.
- Unit tests.
- Bump node to 6.3.1 for a hotfix release.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -125,7 +125,7 @@ jobs:
     strategy:
       matrix:
         plan:
-        - rust: 1.68
+        - rust: 1.73
           ghc: 9.6.4
 
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## Unreleased changes
 
+- Change the severity of logs for failed gRPC API requests to DEBUG level.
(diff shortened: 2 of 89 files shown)


limemloh (approved):

chportma (approved):

tschudid (approved):
Looks good to me

eb-concordium (approved):
Looks good!

PR #1229, opened 2024-09-06
Account migration fix

td202:
## Purpose

Fixes a bug in state migration that results in migration failure in protocol update P6->P7.

## Changes

- Extra TRACE logging during state migration.
- Use a monad transformer transformer to manage the staked balance state during the migration. Using this instead of (simply) a the `StateT` monad transformer ensures that the additional state and the transformed monad transformer are removed when using `lift`, rather than just the additional state. This is required for references to be migrated correctly from one state to another.
- Tweak the account migration tests for P6->P7 to ensure it triggers the bug.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
@@ -1,9 +1,11 @@
 {-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE BinaryLiterals #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 -- We suppress redundant constraint warnings since GHC does not detect when a constraint is used
@@ -16,11 +18,13 @@
 module Concordium.GlobalState.Persistent.Account.StructureV1 where
 
 import Control.Monad
+import qualified Control.Monad.State.Class as State
 import Control.Monad.Trans
-import qualified Control.Monad.Trans.State.Strict as State
+import qualified Control.Monad.Trans.State.Strict as State (StateT (..))
 import Data.Bits
 import Data.Bool.Singletons
 import Data.Foldable
+import Data.Kind
 import qualified Data.Map.Strict as Map
 import Data.Serialize
 import Data.Word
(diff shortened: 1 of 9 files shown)


limemloh (approved):
Well done! :+1:

concordium-consensus/tests/globalstate/GlobalStateTests/AccountsMigrationP6ToP7.hs
```diff
@@ -28,6 +28,7 @@ import Concordium.Types.Accounts
```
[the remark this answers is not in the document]
td202: Yes. I changed the test before I implemented the fix to make sure that was the case.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -206,8 +211,42 @@ migratePersistentAccountStakeEnduring PersistentAccountStakeEnduringBaker{..} =
...
     return $! PersistentAccountStakeEnduringDelegator{..}
 
+-- | A monad transformer transformer that left-composes @StateT Amount@
+--  with a given monad transformer @t@. The purpose of this is to add state functionality for
+--  tracking the current active stake on an account, which is used when migrating an account
+--  between certain protocol versions.
+--
+--  A monad transformer transformer is used so that the 'lift' operation removes both the
+--  @StateT Amount@ and the underlying monad transformer @t@. This is important as the reference
+--  migration functions depend on using 'lift' to access the source block state.
```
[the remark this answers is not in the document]
td202: I agree that it's not ideal. I think these days, the right way to do things would be to use an effect system rather than monad transformers.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -206,8 +211,42 @@ migratePersistentAccountStakeEnduring PersistentAccountStakeEnduringBaker{..} =
...
+newtype StakedBalanceStateTT (t :: (Type -> Type) -> (Type -> Type)) (m :: Type -> Type) (a :: Type) = StakedBalanceStateTT
```
drsk0, line 222: what's the reason we need a transformer over a transformer, instead of just a normal monad transformer?

td202: Consider, for instance `migrateReference`:
```haskell
migrateReference ::
    forall t m ref1 ref2 a b.
    (MonadTrans t, Reference m ref1 a, Reference (t m) ref2 b) =>
    (a -> t m b) ->
    ref1 a ->
    t m (ref2 b)
migrateReference f hb = do
    a <- f =<< lift (refLoad hb)
    newRef <- refMake a
    (!newFlushedRef, _) <- refFlush newRef
    !_ <- lift (refUncache hb)
    return $! newFlushedRef
```

This relies on the monad `m` providing access to the source block state and `t m` providing access to the target block state. So `lift (refLoad hb)` loads from the old state and `refMake` creates a reference in the new state. When we add the extra `StateT` layer to our stack, we want `t = StateT . t0` and `m = m0` rather than `t = StateT` and `m = t0 m0`. So we need to transform the monad transformer.

drsk0: Makes sense, thanks!

PR #1234, opened 2024-09-11
Consensus exception handling

### LINKED ISSUE #1233: Exception Handling
When an unrecoverable exception occurs in the consensus, ensure the following:

* The exception is logged at ERROR level.
* Consensus does not continue to execute past that point, since we cannot trust that it will behave correctly.

Note: such unrecoverable exceptions are generally a result of bugs, and as such should not occur. (They could also potentially occur if the node's state is somehow tampered with or corrupted externally.) These can include observing an invariant violation in the state. While it is not good for everything to shut down, it is generally better than the alternative of continuing in some way, which could end up causing wider damage to the chain.

td202:
## Purpose

Closes #1233

This improves the handling of exceptions in consensus so that, if an unrecoverable exception occurs (e.g. an invariant violation), then consensus cannot continue actively.

## Changes

- When an (unrecoverable) exception occurs in a thread holding the global write lock, the lock is put into a poisoned state. When another thread attempts to acquire the lock, it will also raise an exception.
- Ensure that such exceptions are handled appropriately.
- Add a `ResultConsensusFailure` return code to the FFI `receive` functions indicating that an exception has occurred and consensus has been stopped.
- On the Rust side, observing `ConsensusFailure` results in a panic.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/MultiVersion.hs
+++ b/concordium-consensus/src/Concordium/MultiVersion.hs
@@ -518,7 +518,8 @@ data GlobalLockPoisonError = GlobalLockPoisonError
     deriving (Eq, Show, Typeable)
 
 instance Exception GlobalLockPoisonError where
-    displayException _ = "The global state lock was poisoned by an error in another thread."
+    displayException _ =
+        "The global state lock was poisoned by an unrecoverable error in another thread."
 
 -- | The context for managing multi-version consensus.
 data MultiVersionRunner finconf = MultiVersionRunner
--- a/concordium-consensus/src/Concordium/MultiVersion.hs
+++ b/concordium-consensus/src/Concordium/MultiVersion.hs
@@ -23,9 +23,9 @@
 module Concordium.MultiVersion where
 
 import Control.Concurrent
-import Control.Exception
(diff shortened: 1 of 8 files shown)


concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -500,6 +501,25 @@ data CatchUpStatusBufferState
...
+    deriving (Eq)
+
+-- | An exception indicating that the global state lock was poisoned by an error in another thread.
+--  This indicates that some invariants of the global state may have been violated, and is not
+--  recoverable.
+data GlobalLockPoisonError = GlobalLockPoisonError
+    deriving (Eq, Show, Typeable)
+
+instance Exception GlobalLockPoisonError where
+    displayException _ = "The global state lock was poisoned by an error in another thread."
```
drsk0, line 521: Maybe worth pointing out, that the error is unrecoverable here as well.

drsk0 (approved):
Looks good :fire: .

eb-concordium (approved):

PR #1238, opened 2024-09-12
7.0 fixes

### LINKED ISSUE #1236: "Missing finalization entry" on restart after protocol update
**Bug Description**

After a protocol update, when no further blocks have been processed, restarting the node can result in the error:

> Missing finalization entry for last finalized block

**Steps to Reproduce**

- Stop the node just after a protocol update to P7/P6 (without letting it connect to peers or catch-up, so that it cannot progress beyond the regenesis block).
- Restart the node.

**Expected Result**
The node should behave normally.

**Actual Result**
The node crashes with the error described above.

**Versions**
- Software Version: 6.*, 7.0.0-7.0.2
- OS: all

### LINKED ISSUE #1237: Miscomputation of the genesis height after protocol update
**Bug Description**

The genesis height is not computed correctly when the node is restarted after a second protocol update in the new consensus version (e.g. P5->P6->P7).

**Steps to Reproduce**

Restart the node after a second protocol update in the new consensus version. (Enable TRACE logging.)

**Expected Result**

The node continues executing normally.

**Actual Result**

The node rejects the next block as having the wrong result hash. Subsequent blocks are logged as `pending its parent`.

```
2024-09-12T15:00:45.857009600+02:00 TRACE concordium_node::consensus_ffi::ffi - Konsensus: Block 1c610f55333878623e9a6dce085d32e0614e73a81d4c9747d436c0e77a2656be stated result hash (46343c5600e186fb43bdcf4e2f2aa733df6e7ac8d12e1afaea1343e673a90485) does not match computed value (c21ae7c1edfa8f9448f7654d23df42cd85db972d63f4c9f7a6d70f9c933419fa).
```

**Versions**
- Software Version: 6.*, 7.0.0-7.0.2
- OS: all

td202:
## Purpose

Fixes #1236 and #1237. 7.0.3 release.

## Changes

- On startup, when determining the current epoch and finalization entry for entering the current epoch, if there is no last finalization entry then it actually means that the last finalized block is the genesis block, so we should not throw an error. Fixes #1236
- When iteratively loading the consensuses, correctly compute the regenesis absolute block height (rather than simply coercing a relative block height). Fixes #1237
- Bump version to 7.0.3

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased changes
 
+## 7.0.3
+
+- Fix a bug in the computation of the genesis height after the second protocol update. (#1237)
+- Fix a bug where an error was incorrectly thrown when loading the consenus state immediately
+  after a protocol update (in the new consensus version) (#1236).
+
 ## 7.0.2
 
 - Fix the timing of paydays after protocol update from version 6 to 7.
--- a/concordium-node/Cargo.lock
+++ b/concordium-node/Cargo.lock
@@ -787,7 +787,7 @@ dependencies = [
 
 [[package]]
 name = "concordium_node"
-version = "7.0.2"
+version = "7.0.3"
 dependencies = [
  "anyhow",
  "app_dirs2",
--- a/concordium-node/Cargo.toml
+++ b/concordium-node/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "concordium_node"
-version = "7.0.2" # must be kept in sync with 'is_compatible_version' in 'src/configuration.rs'
+version = "7.0.3" # must be kept in sync with 'is_compatible_version' in 'src/configuration.rs'
 description = "Concordium Node"
 authors = ["Concordium <developers@concordium.com>"]
 exclude = [".gitignore", ".gitlab-ci.yml", "test/**/*","**/**/.gitignore","**/**/.gitlab-ci.yml"]


concordium-consensus/src/Concordium/KonsensusV1/TreeState/StartUp.hs
```diff
@@ -247,8 +247,8 @@ loadSkovData _genesisBlockHeight _runtimeParameters didRollback = do
             && not consensusIsShutdown
             then case mLatestFinEntry of
                 Nothing ->
-                    throwM . TreeStateInvariantViolation $
-                        "Missing finalization entry for last finalized block"
+                    -- In this case, by the above check, the last finalized block is the genesis block.
+                    return (blockEpoch lastFinBlock, Absent)
```
td202, line 251: The referenced above check is:
```haskell
        case mLatestFinEntry of
            Nothing
                | blockRound lastFinBlock == 0 ->
                    return
                        CertifiedBlock
                            { cbQuorumCertificate = genesisQuorumCertificate (getHash genesisBlock),
                              cbQuorumBlock = genesisBlock
                            }
                | otherwise ->
                    throwM . TreeStateInvariantViolation $
"Missing finalization entry for last finalized block"
```
i.e. if the round wasn't 0 (indicating genesis) we'd have already thrown an exception.

limemloh (approved):

PR #1244, opened 2024-10-08
Start github actions

lassemand:
## Purpose

Change to be using github actions instead of jenkins
This is supposed to be a big bang integration with no expectation on keeping jenkins builds alive, should we encounter issues then we revert this PR.
I am not changing the current workflows unless errors exists. 

## Changes

Two different releases, namely rc and alpha. Furthermore using workflow dispatch one can build an individual artifact or image. 
Alpha releases: macOS, Windows, Ubuntu stagenet, Docker stagenet, bootstrapper, and the database exporter.
Rc releases: Everything
Making individual build be using --release
Move into build variable definition into the caller
Use prebuilt binaries on linux instead of building it for flatbuffers

## TODO
Validate artifact and images:

- [x] macOS
- [x] Windows
- [x] Ubuntu stagenet
- [x] Ubuntu testnet
- [x] Ubuntu mainnet
- [x] Ubuntu flynet
- [x] Bootstrapper
- [x] Database exporter
- [x] Docker stagenet
- [x] Docker testnet
- [x] Docker mainnet

## Checklist

      hard-to-understand areas.


scripts/database-exporter/Dockerfile
```diff
@@ -1,25 +0,0 @@
-ARG base_image_tag
```
lassemand, line 1: These files seems to have been forgotten to be removed after some duplication occurred.

scripts/distribution/windows/build-all.ps1
```diff
@@ -5,19 +5,23 @@ Write-Output "cargo version: $(cargo --version)"
 Write-Output "flatc version: $(flatc --version)"
 Write-Output "protoc version: $(protoc --version)"
 
-# Set the default rust toolchain so that consensus rust dependencies use it.
-rustup default $rustVersion-x86_64-pc-windows-gnu
```
lassemand, line 9: Using default version default in workflow instead

scripts/distribution/windows/build-all.ps1
```diff
@@ -5,19 +5,23 @@ Write-Output "cargo version: $(cargo --version)"
...
-
 Write-Output "Building consensus..."
-stack build
+
+& {
+    Set-Location concordium-consensus
```
lassemand, line 11: May not be needed

.github/workflows/release.yaml
```diff
@@ -0,0 +1,437 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.VERSION }}
+          submodules: recursive
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: 1.73.0
```
lassemand, line 258: Needs to be controlled using env

.github/workflows/release.yaml
```diff
@@ -0,0 +1,568 @@
...
+     - '*.*.*-*-rc'
+     - '*.*.*-*-alpha'
+
+env:
+  UBUNTU_VERSION: '20.04'
+  STATIC_LIBRARIES_IMAGE_TAG: 'rust-1.73.0_ghc-9.6.4'
+  RUST_VERSION: '1.73'
+  VERSION_TAG: '7.0.5-1-alpha'
+  STACK_VERSION: '2.13.1'
+  FLATBUFFERS_VERSION: '24.3.25'
```
lassemand, line 19: Bumped to most recent because https://github.com/google/flatbuffers/blob/master/CHANGELOG.md#23526-may-26-2023 gave issues with the make on the docker container on linux.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,569 @@
...
+  RUST_VERSION: '1.73'
+  VERSION_TAG: '7.0.5-1-alpha'
+  STACK_VERSION: '2.13.1'
+  FLATBUFFERS_VERSION: '23.5.26'
+  GHC_VERSION: '9.6.4'
+  CABAL_VERSION: '3.6.2.0'
+  PROTOC_VERSION: '25.2'
+  STATIC_NODE_BINARY_IMAGE_NAME: 'static-node-binaries'
+  AWS_ROLE_TO_ASSUME: 'arn:aws:iam::192549843005:role/github_concordium-node'
+  S3_OBJECTS_TEMPLATE: '{
```
lassemand, line 25: We want to ensure that all our outputs are not there already

.github/workflows/release.yaml
```diff
@@ -0,0 +1,569 @@
...
+#             genesis_path: mainnet/2021-06-09
+#             grpc_port: 20000
+#             listen_port: 8888
+#             exclude_from_release_type: alpha
+           - env: flynet
+             tld: com
+             genesis_path: flynet/2023-02-01/genesis_data
+             grpc_port: 20002
+             listen_port: 8890
+#         exclude:
```
lassemand, line 424: This is supposed to cover the different release types.

Those did not seem to work for now, so lets skip that

scripts/distribution/windows/build-all.ps1
```diff
@@ -5,19 +5,16 @@ Write-Output "cargo version: $(cargo --version)"
...
 stack build
 if ($LASTEXITCODE -ne 0) { throw "Failed building consensus" }
 
 Write-Output "Building node..."
-stack exec -- cargo build --manifest-path concordium-node\Cargo.toml --release
+stack exec -- cargo build --manifest-path concordium-node\Cargo.toml --release --locked
 if ($LASTEXITCODE -ne 0) { throw "Failed building node" }
 
 Write-Output "Building the collector..."
-cargo +$rustVersion-x86_64-pc-windows-msvc build --manifest-path collector\Cargo.toml --release
```
lassemand, line 20: This seems to be wrong according to README

scripts/static-binaries/build-on-ubuntu.sh
```diff
@@ -46,12 +50,12 @@ rm protoc.zip
...
 source "$HOME/.cargo/env"
 rustup set profile minimal
-rustup default "$rust_toolchain_version"
+rustup default "$RUST_TOOLCHAIN_VERSION"
 
 # Install flatbuffers.
 
-git clone https://github.com/google/flatbuffers.git
-( cd flatbuffers && git checkout "$flatbuffers_version" && cmake -G "Unix Makefiles" && make -j"$(nproc)" && make install )
+curl -L "https://github.com/google/flatbuffers/releases/download/v${FLATBUFFERS_VERSION}/Linux.flatc.binary.g++-10.zip" -O
```
lassemand, line 57: We should obtain the g++ version dynamically

scripts/distribution/windows/build-all.ps1
EmilKWarmdahl: Seems like this is to force the toolchain according to the cargo [documentation](https://doc.rust-lang.org/cargo/commands/cargo.html#common-options) I think it's best to keep it.

lassemand: This is to force a different toolchain. According to docs:

`  - For building the node runner service (optional), the toolchain `1.73-x86_64-pc-windows-msvc`  is required, which can be installed with the command: `rustup toolchain install 1.73-x86_64-pc-windows-msvc`.
`

This is not any such service. I want to use the default service here.

.github/workflows/release.yaml
```diff
@@ -285,7 +288,8 @@ jobs:
   node-windows:
     runs-on: windows-latest
     needs: [validate-preconditions]
-    if: contains(fromJSON('["rc", "alpha", "node-windows"]'), needs.validate-preconditions.outputs.release_type)
+    #if: contains(fromJSON('["rc", "alpha", "node-windows"]'), needs.validate-preconditions.outputs.release_type)
+    if: false
```
EmilKWarmdahl, line 292: Revert this as well as on the macos and docker jobs

.github/workflows/release.yaml
```diff
@@ -79,7 +82,7 @@ jobs:
         run: |
           CARGO_VERSION=$(yq .package.version concordium-node/Cargo.toml)
           if [ -z "${{ env.SERVICE }}" ]; then
-            IFS='-' read -r VERSION BUILD RELEASE_TYPE <<< "${{ github.ref }}"
+            IFS='-' read -r VERSION BUILD RELEASE_TYPE <<< "${{ env.VERSION_TAG }}"
```
EmilKWarmdahl, line 85: Use the actual tag

.github/workflows/release.yaml
```diff
@@ -15,6 +15,8 @@ on:
           - database-exporter
 
   push:
+    branches:
+      - lma/release/github_actions
```
EmilKWarmdahl, line 19: Remove this

.github/workflows/release.yaml
```diff
@@ -33,20 +35,21 @@ env:
...
-      \"docker-stagenet\": \"concordium/stagenet-node:${VERSION}-dev\",
-      \"docker-testnet\": \"concordium/testnet-node:${VERSION}-dev\",
-      \"docker-mainnet\": \"concordium/mainnet-node:${VERSION}-dev\"
+      \"docker-stagenet\": \"concordium/stagenet-node:${VERSION}\",
+      \"docker-testnet\": \"concordium/testnet-node:${VERSION}\",
+      \"docker-mainnet\": \"concordium/mainnet-node:${VERSION}\"
     }'
   REGISTRY: docker.io
   SERVICE: "${{ inputs.service }}"
+  VERSION_TAG: "7.0.5-4-rc"
```
EmilKWarmdahl, line 52: Remove this

EmilKWarmdahl (approved):
LGTM, but should also get an approve by a dev

.github/workflows/release.yaml
```diff
@@ -0,0 +1,623 @@
...
+          role-to-assume: ${{ env.AWS_ROLE_TO_ASSUME }}
+          role-session-name: ValidatePreconditionsSession
+          aws-region: "eu-west-1"
+
+      - name: Validate version
+        id: versions_derivation
+        run: |
+          CARGO_VERSION=$(yq .package.version concordium-node/Cargo.toml)
+          if [ -z "${{ env.SERVICE }}" ]; then
+            IFS='-' read -r VERSION BUILD RELEASE_TYPE <<< "${{ env.VERSION_TAG }}"
```
drsk0, line 82: Is the `VERSION_TAG` set by github actions itself? Or where is it coming from?

lassemand: That is a mistake yes. I used it as part of my last test, and forgot to change it back. 

Using: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs

drsk0 (approved):
Lgtm, thanks!

PR #1252, opened 2024-10-24
suspension via computeBakerStakesAndCapital

drsk0:
## Purpose

Set stake to zero of suspended validators.

## Changes

As described above. To facilitate the change, a suspended flag and validator id field are added to `ActiveBakerInfo`.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
@@ -235,7 +235,7 @@ doEpochTransition True epochDuration theState0 = do
                     applyPendingChanges av (<= epochEnd)
                         <$> bsoGetActiveBakersAndDelegators theState7
                 let BakerStakesAndCapital{..} =
-                        computeBakerStakesAndCapital @m
+                        computeBakerStakesAndCapital
                             (chainParams ^. cpPoolParameters)
                             activeBakers
                             passiveDelegators
--- a/concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
@@ -235,7 +235,7 @@ doEpochTransition True epochDuration theState0 = do
                     applyPendingChanges av (<= epochEnd)
                         <$> bsoGetActiveBakersAndDelegators theState7
                 let BakerStakesAndCapital{..} =
-                        computeBakerStakesAndCapital
+                        computeBakerStakesAndCapital @m
                             (chainParams ^. cpPoolParameters)
                             activeBakers
                             passiveDelegators
@@ -244,8 +244,7 @@ doEpochTransition True epochDuration theState0 = do
                         theState7
                         bakerStakes
                         (chainParams ^. cpFinalizationCommitteeParameters)
-                capDist <- capitalDistributionM
-                theState9 <- bsoSetNextCapitalDistribution theState8 capDist
+                theState9 <- bsoSetNextCapitalDistribution theState8 capitalDistribution
                 -- From P7 onwards, we transition pre-pre-cooldowns into pre-cooldowns, so that
                 -- at the next payday they will enter cooldown.
                 case sSupportsFlexibleCooldown (sAccountVersionFor (protocolVersion @(MPV m))) of
--- a/concordium-consensus/src/Concordium/Kontrol/Bakers.hs
+++ b/concordium-consensus/src/Concordium/Kontrol/Bakers.hs
@@ -1,9 +1,9 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE AllowAmbiguousTypes #-}
 -- We suppress redundant constraint warnings since GHC does not detect when a constraint is used
 -- for pattern matching. (See: https://gitlab.haskell.org/ghc/ghc/-/issues/20896)
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
@@ -191,16 +191,15 @@ computeBakerStakesAndCapital poolParams activeBakers passiveDelegators = BakerSt
     capLimit = takeFraction (theCapitalBound capitalBound) totalCapital
     makeBakerStake ActiveBakerInfo{..} poolCap =
         ( activeBakerInfoRef,
-          if activeBakerIsSuspended
-            then 0
-            else
-                minimum
-                    [ poolCap,
-                      applyLeverageFactor leverage activeBakerEquityCapital,
-                      capLimit
-                    ]
+          minimum
+            [ poolCap,
+              applyLeverageFactor leverage activeBakerEquityCapital,
+              capLimit
+            ]
         )
-    bakerStakes = zipWith makeBakerStake activeBakers poolCapitals
+    activeBakers' = [abi | abi@ActiveBakerInfo{..} <- activeBakers, not activeBakerIsSuspended]
+    poolCapitals' = poolCapital <$> activeBakers'
+    bakerStakes = zipWith makeBakerStake activeBakers' poolCapitals'
     delegatorCapital ActiveDelegatorInfo{..} = DelegatorCapital activeDelegatorId activeDelegatorStake
     bakerCapital ActiveBakerInfo{..} =
         BakerCapital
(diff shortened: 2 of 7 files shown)


concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -190,13 +192,19 @@ computeBakerStakesAndCapital poolParams activeBakers passiveDelegators = BakerSt
...
+            bakerEx <- derefBakerInfoEx activeBakerInfoRef
+            return $
+                if fromCondDef (_bieAccountIsSuspended bakerEx) False
+                    then 0
+                    else
+                        minimum
+                            [ poolCap,
+                              applyLeverageFactor leverage activeBakerEquityCapital,
+                              capLimit
+                            ]
```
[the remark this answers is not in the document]
drsk0: Is `ActiveBakerInfo` purely in memory such that we don't need to guard the `activeBakerSuspended` field with a `conditionally`?

drsk0:
Tests are missing at the moment, is there a good place to add them @td202 ?

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/LeaderElectionTest.hs
```diff
@@ -157,40 +157,40 @@ testComputeMissedRounds :: Spec
 testComputeMissedRounds =
     describe "computeMissedRounds" $ do
         it "no timeout" $
-            ( Map.toList $
-                computeMissedRounds
+            Map.toList
+                ( computeMissedRounds
```
drsk0, line 161: This is just a lint unrelated to the change.

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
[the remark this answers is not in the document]
td202: Yes. `ActiveBakerInfo` is really just an intermediary for data from `bsoGetActiveBakersAndDelegators` to be passed to `computeBakerStakesAndCapital` (via `applyPendingChanges`).

td202:
> Tests are missing at the moment, is there a good place to add them @td202 ?

Probably either in EpochTransition, if they fit there, or else add a new test file (probably in `SchedulerTests`).

td202 (changes requested):

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -190,26 +190,26 @@ computeBakerStakesAndCapital poolParams activeBakers passiveDelegators = BakerSt
...
-              capLimit
-            ]
+          if activeBakerIsSuspended
+            then 0
+            else
+                minimum
+                    [ poolCap,
+                      applyLeverageFactor leverage activeBakerEquityCapital,
+                      capLimit
+                    ]
```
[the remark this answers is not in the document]
drsk0: I changed this as described. I don't have a hard opinion on this, I think both approaches suffer a bit from being not very explicit.
> I'm not 100% certain that we never rely on validators' stakes being non-zero.

Are we 100% sure that we never rely on validators being present in `bakerStakes`?

> Even 0-weighted validators could end up as finalizers, and potentially end up signing blocks even though their signatures contribute nothing. This seems weird to me.

Agreed.

> If we remove the suspended validators from the epoch bakers/capital distribution, that means that when they become un-suspended then their missed block counters will be reset to 0, which I think is what we want. Maybe?

Yes, this is what we want. I'm not clear however why this would happen as a consequence from them being removed from the epoch bakers/capital distribution.

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
drsk0, line 1: I haven't found a way to avoid this.

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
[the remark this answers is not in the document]
drsk0: :+1:  makes sense, thanks!

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -179,7 +180,7 @@ computeBakerStakesAndCapital ::
     PoolParameters' 'PoolParametersVersion1 ->
     [ActiveBakerInfo m] ->
     [ActiveDelegatorInfo] ->
-    BakerStakesAndCapital m
+    BakerStakesAndCapital (BakerInfoRef m)
```
td202, line 183: Change the signature to:
```
computeBakerStakesAndCapital ::
    PoolParameters' 'PoolParametersVersion1 ->
    [ActiveBakerInfo' bakerInfoRef] ->
    [ActiveDelegatorInfo] ->
    BakerStakesAndCapital bakerInfoRef
```
Then you don't need to `AllowAmbiguousTypes`.

Because `BakerInfoRef` is not injective, `m` is an ambiguous type variable: i.e. there can be multiple different `m`s for which the types reduce to the same thing.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -235,7 +235,7 @@ doEpochTransition True epochDuration theState0 = do
                     applyPendingChanges av (<= epochEnd)
                         <$> bsoGetActiveBakersAndDelegators theState7
                 let BakerStakesAndCapital{..} =
-                        computeBakerStakesAndCapital
+                        computeBakerStakesAndCapital @m
```
td202, line 238: After changing the type signature for `computeBakerStakesAndCapital`, this type application shouldn't be needed.

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -196,20 +197,19 @@ computeBakerStakesAndCapital poolParams activeBakers passiveDelegators = BakerSt
               capLimit
             ]
         )
-    bakerStakes = zipWith makeBakerStake activeBakers poolCapitals
+    activeBakers' = [abi | abi@ActiveBakerInfo{..} <- activeBakers, not activeBakerIsSuspended]
+    poolCapitals' = poolCapital <$> activeBakers'
```
td202, line 201: *Suggestion (non-blocking):* consider `filteredActiveBakers` and `filteredPoolCapitals` rather than `activeBakers'` and `poolCapitals'`.

concordium-consensus/tests/scheduler/SchedulerTests/KonsensusV1/EpochTransition.hs
```diff
@@ -464,19 +463,23 @@ testEpochTransitionPaydayOnly accountConfigs = runTestBlockState @P7 $ do
             ^. cpCooldownParameters . cpUnifiedCooldown
 
 -- | Test an snapshot epoch transition.
-testEpochTransitionSnapshotOnly :: [AccountConfig 'AccountV3] -> Assertion
+testEpochTransitionSnapshotOnly ::
+    forall m.
+    (m ~ PersistentBlockStateMonad P7 (PersistentBlockStateContext P7) (BlobStoreT (PersistentBlockStateContext P7) (LoggerT IO))) =>
+    [AccountConfig 'AccountV3] ->
+    Assertion
```
td202, line 470: I think changing the signature of `computeBakerStakesAndCapital` should allow simplifying this back to the original signature (and elsewhere).

```suggestion
testEpochTransitionSnapshotOnly :: [AccountConfig 'AccountV3] -> Assertion
```

concordium-consensus/tests/scheduler/SchedulerTests/KonsensusV1/EpochTransition.hs
```diff
@@ -763,6 +778,78 @@ testMissedRoundsUpdate accountConfigs = runTestBlockState @P8 $ do
     startEpoch = 10
     startTriggerTime = 1000
 
+-- Test that suspended validators have zero stake.
```
td202, line 781:
```suggestion
-- | Test that suspended validators have zero stake.
```

td202 (approved):
Looks good!

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
drsk0: Changed it and removed the pragma.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
drsk0: yep, this has been changed.

PR #1258, opened 2024-11-19
sketch for auto suspension

drsk0:
## Purpose

Discuss how to trigger suspension at snapshot. I filled in most of the missing pieces, but would like to be sure, that `processSuspensions` is doing the right thing.

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/PoolRewards.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/PoolRewards.hs
@@ -15,6 +15,28 @@ import Concordium.Types.Conditionally
 import Concordium.Types.HashableTo
 import Concordium.Utils.Serialization
 
+-- | Information needed to determine whether to suspend a validator.
+data SuspensionInfo = SuspensionInfo
+    { -- | The number of missed rounds since the validator most recently
+      --  became a (non-suspended) member of the validator committee.
+      missedRounds :: !Word64,
+      -- | Flag indicating that the validator should be suspended at the coming
+      --  snapshot epoch, because its missed rounds have crossed the threshold
+      --  given in the chain parameters in the previous reward period.
+      primedForSuspension :: !Bool
+    }
+    deriving (Eq, Show)
+
+emptySuspensionInfo :: SuspensionInfo
+emptySuspensionInfo = SuspensionInfo{missedRounds = 0, primedForSuspension = False}
+
+instance Serialize SuspensionInfo where
+    put SuspensionInfo{..} = do
+        putWord64be missedRounds
+        putBool primedForSuspension
+    get =
+        SuspensionInfo <$> get <*> getBool
+
 -- | 'BakerPoolRewardDetails' tracks the rewards that have been earned by a baker pool in the current
 --  reward period. These are used to pay out the rewards at the payday.
 data BakerPoolRewardDetails (av :: AccountVersion) = BakerPoolRewardDetails
@@ -24,8 +46,8 @@ data BakerPoolRewardDetails (av :: AccountVersion) = BakerPoolRewardDetails
       transactionFeesAccrued :: !Amount,
       -- | Whether the pool contributed to a finalization proof in the reward period
       finalizationAwake :: !Bool,
-      -- | The number of missed rounds in the reward period
-      missedRounds :: !(Conditionally (SupportsValidatorSuspension av) Word64)
+      -- | Information for deciding whether a validator will be suspended the next snapshot epoch.
+      suspensionInfo :: !(Conditionally (SupportsValidatorSuspension av) SuspensionInfo)
     }
     deriving (Eq, Show)
 
@@ -34,7 +56,7 @@ instance forall av. (IsAccountVersion av) => Serialize (BakerPoolRewardDetails a
         putWord64be blockCount
         put transactionFeesAccrued
         putBool finalizationAwake
-        mapM_ putWord64be missedRounds
+        mapM_ put suspensionInfo
 
     get =
         BakerPoolRewardDetails
@@ -55,7 +77,8 @@ emptyBakerPoolRewardDetails =
         { blockCount = 0,
           transactionFeesAccrued = 0,
           finalizationAwake = False,
-          missedRounds = conditionally (sSupportsValidatorSuspension (accountVersion @av)) 0
+          suspensionInfo =
+            conditionally (sSupportsValidatorSuspension (accountVersion @av)) emptySuspensionInfo
         }
 
 -- | Migrate BakerPoolRewardDetails with different account versions.
@@ -66,9 +89,9 @@ migrateBakerPoolRewardDetails ::
     BakerPoolRewardDetails av1
 migrateBakerPoolRewardDetails BakerPoolRewardDetails{..} =
     BakerPoolRewardDetails
-        { missedRounds =
+        { suspensionInfo =
             conditionally
                 (sSupportsValidatorSuspension (accountVersion @av1))
-                (fromCondDef missedRounds 0),
+                (fromCondDef suspensionInfo emptySuspensionInfo),
           ..
         }
--- a/concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
@@ -9,6 +9,7 @@ module Concordium.KonsensusV1.Scheduler where
 import Control.Monad
 import Data.Bool.Singletons
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 import Data.Time
 import Data.Word
 import Lens.Micro.Platform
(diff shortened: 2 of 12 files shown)


concordium-consensus/src/Concordium/GlobalState/PoolRewards.hs
```diff
@@ -35,13 +36,15 @@ instance forall av. (IsAccountVersion av) => Serialize (BakerPoolRewardDetails a
         put transactionFeesAccrued
         putBool finalizationAwake
         mapM_ putWord64be missedRounds
+        putBool suspensionPrimed
```
td202, line 39: Could combine the boolean flags with a bitmap.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -231,6 +231,11 @@ doEpochTransition True epochDuration theState0 = do
                 -- This is the start of the last epoch of a payday, so take a baker snapshot.
                 let epochEnd = newSeedState ^. triggerBlockTime
                 let av = accountVersionFor (demoteProtocolVersion (protocolVersion @(MPV m)))
+
+                -- TODO (drsk) suspend validators that are primed for suspension and whose 
+                -- missed rounds are above the threshold set by the score chain parameter.
+                -- Clear all active bakers from being primed for suspension
```
td202, line 237: I think this is good and in line with the proposal.

I mentioned that we should add special transaction outcomes for suspending accounts. My main concern with this is that the special transaction outcomes are treated as happening after the regular transactions. For instance, if an account receives a reward in a block (which is in a special transaction outcome), that reward cannot be spent in the same block by a regular transaction. This means that if we have a special transaction outcome suspending a validator and a regular transaction resuming the validator in the same block, then we have to be a bit careful about handling this. In particular:
- The validator should be considered suspended for the purposes of calculating the next epoch bakers.
- There will be a transaction outcome indicating the validator was resumed manually.
- There should be a special transaction outcome indicating the validator was suspended automatically.
- The interpretation of special transaction outcomes happening after the regular transactions means we should actually do the suspension when executing the block epilogue.

concordium-consensus/src/Concordium/GlobalState/PoolRewards.hs
```diff
@@ -25,7 +25,8 @@ data BakerPoolRewardDetails (av :: AccountVersion) = BakerPoolRewardDetails
       -- | Whether the pool contributed to a finalization proof in the reward period
       finalizationAwake :: !Bool,
       -- | The number of missed rounds in the reward period
-      missedRounds :: !(Conditionally (SupportsValidatorSuspension av) Word64)
+      missedRounds :: !(Conditionally (SupportsValidatorSuspension av) Word64),
+      suspensionPrimed :: !Bool
```
td202, line 29: Document this.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
drsk0: I implemented this as discussed above (I think). @td202 could you take a look at `processSuspension` to tell me that this is the right direction?

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -363,7 +370,13 @@ processPaydayRewards (Just PaydayParameters{..}) theState0 = do
     -- in which the rewards are distributed.
     foundationAddr <- getAccountCanonicalAddress =<< bsoGetFoundationAccount theState0
     theState1 <- doMintingP6 paydayMintRate foundationAddr theState0
-    distributeRewards foundationAddr paydayCapitalDistribution paydayBakers paydayPoolRewards theState1
+    theState2 <- distributeRewards foundationAddr paydayCapitalDistribution paydayBakers paydayPoolRewards theState1
+    -- We prime all active validators of the payday for possible suspension the next snapshot.
+    case hasValidatorSuspension of
+        STrue -> bsoPrimeForSuspension theState2 (Map.keys paydayPoolRewards)
+        SFalse -> return theState2
```
td202, line 377: Don't we only want to prime the validators that are over the threshold for missed blocks? Also, I would suggest using `paydayBakers` as the source of the list of `BakerID`s rather than `paydayPoolRewards` (e.g. `paydayBakers ^.. each . bakerIdentity`).

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -402,10 +415,75 @@ processBlockRewards ParticipatingBakers{..} TransactionRewardParameters{..} miss
...
+        SomeParam (ValidatorScoreParameters{..}) -> do
+            -- get baker pool reward details
+            bprd <- bsoGetBakerPoolRewardDetails bs0
+            -- account indexes that will be suspended
+            let ixs =
+                    [ bakerAccountIndex bid
+                      | (bid, rd) <- Map.toList bprd,
+                        fromCondDef (suspensionPrimed rd) False,
+                        fromCondDef (missedRounds rd) 0 > _vspMaxMissedRounds
+                    ]
```
td202, line 445: If we reset the `primed` flag when a validator contributes, then we should only need to check that flag here, and not the number of missed rounds. Assuming that we only set the primed flag in the first place when the validator exceeds the number of missed rounds by the payday.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
td202: I think we should also consider generating a special outcome when a validator is primed to help with notifying inactive validators that they face suspension.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -402,10 +415,75 @@ processBlockRewards ParticipatingBakers{..} TransactionRewardParameters{..} miss
...
+            (upds, bs1) <-
+                foldM
+                    ( \(upds, bs) accIx -> do
+                        res <- bsoUpdateValidator bs timestamp accIx updSuspend
+                        case res of
+                            -- TODO(drsk) how to deal with error?
+                            Left _err -> return (upds, bs)
+                            Right (upds', bs') -> return (upds ++ upds', bs')
```
td202, line 453: I would consider introducing a new operation `bsoSuspendValidator` that just suspends the validator, so that you don't have to worry about impossible error cases here. I'm also a bit suspicious that the precondition that the account is actually a validator may not be met here, because it could be that the validator has been removed since the last snapshot. `bsoSuspendValidator` could be safe to call on accounts that are not validators, and even return whether the validator was actually suspended. When we add the special outcomes, we can probably limit to those that were actually suspended, leaving out any that had already been manually suspended or removed since the last snapshot.

td202 (commented):
This is looking good. I think it would be helpful to refine what the missed block count and primed for suspension actually mean.

- Missed block count: rounds that a validator won but failed to bake a block since the last round that a validator contributed to as a baker or finalizer, since it most recently became a (non-suspended) member of the validator committee.
- Primed for suspension: should be set on a payday when the missed block count exceeds the threshold. Should be cleared when the validator contributes again. At the snapshot epoch, validators that are primed will not be included in the next committee and will be set as suspended (if still an active validator).

If the payday length is 1 epoch, then each payday transition is also the snapshot for the next payday. For this to make sense, we need to determine the set of validators that should get suspended (i.e. the ones that are already marked as primed) before marking any validators as primed. I think this is consistent with the order of operations discussed elsewhere in this review.

concordium-consensus/src/Concordium/GlobalState/BakerInfo.hs
```diff
@@ -291,8 +291,8 @@ data BakerConfigureUpdateChange
     | BakerConfigureTransactionFeeCommission !AmountFraction
     | BakerConfigureBakingRewardCommission !AmountFraction
     | BakerConfigureFinalizationRewardCommission !AmountFraction
-    | BakerConfigureSuspended
-    | BakerConfigureResumed
+    | BakerConfigureSuspended !BakerId
+    | BakerConfigureResumed !BakerId
```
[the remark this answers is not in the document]
drsk0: You're right, I found a way to remove this.

concordium-consensus/src/Concordium/GlobalState/PoolRewards.hs
```diff
@@ -25,7 +25,9 @@ data BakerPoolRewardDetails (av :: AccountVersion) = BakerPoolRewardDetails
       -- | Whether the pool contributed to a finalization proof in the reward period
       finalizationAwake :: !Bool,
       -- | The number of missed rounds in the reward period
-      missedRounds :: !(Conditionally (SupportsValidatorSuspension av) Word64)
+      missedRounds :: !(Conditionally (SupportsValidatorSuspension av) Word64),
```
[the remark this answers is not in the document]
drsk0: I changed the docs.
>  I think we don't reset this counter at the payday, but just when it becomes a validator, right?

Yes, it's reset when it becomes a validator.

concordium-consensus/src/Concordium/GlobalState/PoolRewards.hs
```diff
@@ -25,7 +25,9 @@ data BakerPoolRewardDetails (av :: AccountVersion) = BakerPoolRewardDetails
...
+      -- | A flag indicating whether suspension can be triggered the next snapshot epoch.
+      suspensionPrimed :: !(Conditionally (SupportsValidatorSuspension av) Bool)
```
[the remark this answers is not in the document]
drsk0: I moved `suspensionPrimed` and `missedRounds`  into a data structure `SuspensionInfo`.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
drsk0: this has been changd.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
drsk0: Implemented a `bsoSuspendValidator` block state operation.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
drsk0: Added the special outcome for the validator when it gets primed.

drsk0:
This has been fully implemented now. I'll open two separate PR to bring the changes in. First one is https://github.com/Concordium/concordium-node/pull/1264. This are only the new data structures. The second one will contain the implementation.

PR #1260, opened 2024-11-22
Queries for accounts with scheduled releases or cooldowns

td202:
## Purpose

Closes: https://github.com/Concordium/concordium-node/issues/1263
Depends on: https://github.com/Concordium/concordium-grpc-api/pull/67 https://github.com/Concordium/concordium-base/pull/575

Add gRPC endpoints for getting the lists of accounts with scheduled releases or cooldowns.

## Changes

-    GetScheduledReleaseAccounts endpoint
-    GetCooldownAccounts endpoint
-    GetPreCooldownAccounts endpoint
-    GetPrePreCooldownAccounts endpoint

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Queries.hs
+++ b/concordium-consensus/src/Concordium/Queries.hs
@@ -956,6 +956,8 @@ getCooldownAccounts = liftSkovQueryStateBHI query
     query bs = do
         cooldowns <- BS.getCooldownAccounts bs
         let processAtTimestamp ts accs = ([AccountPending acc ts | acc <- Set.toList accs] ++)
+        -- Right-fold here ensures that the set at each timestamp is converted lazily as
+        -- the resulting list is consumed.
         return $ Map.foldrWithKey processAtTimestamp [] cooldowns
 
 -- | Get the index of accounts in pre-cooldown.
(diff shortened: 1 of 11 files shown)


concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -934,6 +934,44 @@ getNextUpdateSequenceNumbers = liftSkovQueryStateBHI query
...
+-- | Get the index of accounts with stake in cooldown.
+getCooldownAccounts ::
+    forall finconf.
+    BlockHashInput ->
+    MVR finconf (BHIQueryResponse [AccountPending])
+getCooldownAccounts = liftSkovQueryStateBHI query
+  where
+    query bs = do
+        cooldowns <- BS.getCooldownAccounts bs
+        let processAtTimestamp ts accs = ([AccountPending acc ts | acc <- Set.toList accs] ++)
```
drsk0, line 958: Is the O(n) complexity of (++) a problem in this fold?

drsk0 (approved):

concordium-consensus/src/Concordium/Queries.hs
td202: It shouldn't be. The complexity is in the first argument of `(++)`, so it's important that we use `foldr`. When we consume this list produced by the fold, it is lazily constructed by consuming each set of accounts in turn.

drsk0: Maybe it's worth adding a comment here so that future reviewers of the code know that this has been taken into consideration.

PR #1293, opened 2024-12-05
Primed for suspension

drsk0:
## Purpose

This is the implementation of automatic suspension for validators that missed too many rounds. Suspension happens at snapshot, while priming new validators for suspension at payday.

We compute the to be suspended validators during epoch transition at snapshot before we prime validators for future suspension. This is to avoid priming and then immediately suspending validators when the length of the payday is one.

## Changes
- two block state operations: `bsoPrimeValidators` and `bsoSuspendValidators`
- automatic suspend logic in the scheduler
- change `computeBakerStakesAndCapital` to ignore newly suspended validators at snapshot
- a bit of boilerplate code for the chain parameters version 3 slipped in this PR
- tests
## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1532,7 +1532,7 @@ class (BlockStateQuery m) => BlockStateOperations m where
 
     -- | Mark given validators for possible suspension at the next snapshot
     --  epoch. Returns the subset of the given validator ids whose missed rounds
-    --  exceeded the given threshold and are now priMed for suspension.
+    --  exceeded the given threshold and are now primed for suspension.
     bsoPrimeForSuspension :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Word64 -> [BakerId] -> m ([BakerId], UpdatableBlockState m)
 
     -- | Suspend validators with the given account indices, if
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1535,12 +1535,13 @@ class (BlockStateQuery m) => BlockStateOperations m where
     --  exceeded the given threshold and are now priMed for suspension.
     bsoPrimeForSuspension :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Word64 -> [BakerId] -> m ([BakerId], UpdatableBlockState m)
 
-    -- \| Suspend validators with the given account indices, if
+    -- | Suspend validators with the given account indices, if
     --  1) the account index points to an existing account
     --  2) the account belongs to a validator
     --  3) the account was not already suspended
-    --  Returns the subset of account indices that were suspended.
-    bsoSuspendValidators :: (PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> [AccountIndex] -> m ([AccountIndex], UpdatableBlockState m)
+    --  Returns the subset of account indices that were suspended together with their canonical
+    --  addresses.
+    bsoSuspendValidators :: (PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> [AccountIndex] -> m ([(AccountIndex, AccountAddress)], UpdatableBlockState m)
 
     -- | A snapshot of the block state that can be used to roll back to a previous state.
     type StateSnapshot m
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1530,7 +1530,9 @@ class (BlockStateQuery m) => BlockStateOperations m where
     --  round did timeout.
     bsoUpdateMissedRounds :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Map.Map BakerId Word64 -> m (UpdatableBlockState m)
 
-    -- | Mark given validators for possible suspension at the next snapshot epoch.
+    -- | Mark given validators for possible suspension at the next snapshot
+    --  epoch. Returns the subset of the given validator ids whose missed rounds
+    --  exceeded the given threshold and are now priMed for suspension.
     bsoPrimeForSuspension :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Word64 -> [BakerId] -> m ([BakerId], UpdatableBlockState m)
 
     -- \| Suspend validators with the given account indices, if
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1533,7 +1533,16 @@ class (BlockStateQuery m) => BlockStateOperations m where
     -- | Mark given validators for possible suspension at the next snapshot
     --  epoch. Returns the subset of the given validator ids whose missed rounds
     --  exceeded the given threshold and are now primed for suspension.
-    bsoPrimeForSuspension :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Word64 -> [BakerId] -> m ([BakerId], UpdatableBlockState m)
+    bsoPrimeForSuspension ::
+        (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) =>
+        UpdatableBlockState m ->
+        -- | The threshold for maximal missed rounds
+        Word64 ->
+        -- | The set of validators that are considered for suspension. This
+        --  should be the current payday validators.
+        [BakerId] ->
+        -- | Returns the subset of primed validator ids and the updated block state
+        m ([BakerId], UpdatableBlockState m)
 
     -- | Suspend validators with the given account indices, if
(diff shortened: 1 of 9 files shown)


concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -178,8 +179,9 @@ computeBakerStakesAndCapital ::
     PoolParameters' 'PoolParametersVersion1 ->
     [ActiveBakerInfo' bakerInfoRef] ->
     [ActiveDelegatorInfo] ->
+    Set.Set BakerId ->
```
td202, line 182: Add some documentation on parameters.

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
```diff
@@ -219,9 +221,10 @@ generateNextBakers ::
     ) =>
     -- | The payday epoch
     Epoch ->
+    Set.Set BakerId ->
     UpdatableBlockState m ->
     m (UpdatableBlockState m)
-generateNextBakers paydayEpoch bs0 = do
+generateNextBakers paydayEpoch suspendedBids bs0 = do
```
td202, line 227: This function is only called in one place, and there with `Set.empty`, so I don't think there's any need to parametrise it with the set.

concordium-consensus/src/Concordium/Scheduler/TreeStateEnvironment.hs
td202, line 1: These changes can be reverted if `generateNextBakers` doesn't require the set argument.

td202 (dismissed):
Looks good. Just a few changes.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -416,12 +498,20 @@ executeBlockEpilogue ::
...
-executeBlockEpilogue participants paydayParams transactionRewardParams missedRounds theState0 = do
+executeBlockEpilogue participants paydayParams transactionRewardParams missedRounds snapshotSuspendedBids theState0 = do
     theState1 <- processPaydayRewards paydayParams theState0
     theState2 <- processBlockRewards participants transactionRewardParams missedRounds theState1
-    freezeBlockState theState2
+    theState3 <- case hasValidatorSuspension of
+        STrue
+            | Just suspendedBids <- snapshotSuspendedBids -> processSuspensions suspendedBids theState2
+            | otherwise -> return theState2
+        SFalse -> return theState2
```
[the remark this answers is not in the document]
drsk0: I think the more explicit case distinction is clearer.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -350,9 +377,11 @@ doMintingP6 mintRate foundationAddr theState0 = do
 
 -- | If a payday has elapsed, this mints and distributes rewards for the payday.
```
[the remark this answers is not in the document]
drsk0: Updated the docs.

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
drsk0: ok I removed the parameter and updated the docs, that this assumes no suspended validators.

concordium-consensus/src/Concordium/Scheduler/TreeStateEnvironment.hs
drsk0: done

concordium-consensus/src/Concordium/Kontrol/Bakers.hs
drsk0: documented the parameters

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -225,9 +243,16 @@ doEpochTransition True epochDuration theState0 = do
...
+    let isSnapshot = newEpoch + 1 == newNextPayday
+    (suspendedBids, theState8) <- do
+        if isSnapshot
             then do
+                snapshotPoolRewards <- bsoGetBakerPoolRewardDetails theState7
+                -- account indexes that will be suspended
+                let suspendedBids =
+                        Set.fromList
+                            [ bid | (bid, rd) <- Map.toList snapshotPoolRewards, primedForSuspension $ fromCondDef (suspensionInfo rd) emptySuspensionInfo
+                            ]
```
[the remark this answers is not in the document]
drsk0: the computation is only done if the protocols supports suspension.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -363,7 +392,37 @@ processPaydayRewards (Just PaydayParameters{..}) theState0 = do
...
+                            address <- getAccountCanonicalAddress account
+                            let outcome =
+                                    ValidatorPrimedForSuspension
+                                        { vpfsBakerId = bid,
+                                          vpfsAccount = address
+                                        }
+                            bsoAddSpecialTransactionOutcome theState outcome
+                    foldM addOutcome theState3 bids
+  where
+    hasValidatorSuspension = sSupportsValidatorSuspension (sAccountVersionFor (protocolVersion @pv))
```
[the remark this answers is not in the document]
drsk0: factored this out.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1530,6 +1530,19 @@ class (BlockStateQuery m) => BlockStateOperations m where
     --  round did timeout.
     bsoUpdateMissedRounds :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Map.Map BakerId Word64 -> m (UpdatableBlockState m)
 
+    -- | Mark given validators for possible suspension at the next snapshot
+    --  epoch. Returns the subset of the given validator ids whose missed rounds
+    --  exceeded the given threshold and are now primed for suspension.
+    bsoPrimeForSuspension :: (PVSupportsDelegation (MPV m), PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> Word64 -> [BakerId] -> m ([BakerId], UpdatableBlockState m)
```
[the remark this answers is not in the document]
drsk0: Added some more docs for the parameters.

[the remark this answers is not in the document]
drsk0: I think it's better to pass both parameters explicitly. Because this function is meant to be called with the current paydays validators. If they are not passed in explicitly, we would have to gather them (most likely from the `BakerPoolRewardDetails`) during the call. The same holds for the threshold, we could also gather it by inspecting the chain parameters. But I think being explicit makes the intent a lot clearer and doesn't rely on implicit assumptions on the implementations.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3509,6 +3509,87 @@ doUpdateMissedRounds pbs rds = do
...
+--  suspension.
+doPrimeForSuspension ::
+    ( PVSupportsDelegation pv,
+      SupportsPersistentState pv m
+    ) =>
+    PersistentBlockState pv ->
+    Word64 ->
+    [BakerId] ->
+    m ([BakerId], PersistentBlockState pv)
+doPrimeForSuspension pbs threshold bids = do
```
eb-concordium, line 3523: Why only suspend those bakers given in `bids` and not all bakers for which `missedRounds > threshold`?

drsk0: @eb-concordium We only want to prime validators of the current payday.

drsk0: The other variant would be to remove both arguments and just look them up from the block state.

eb-concordium: I see 👍

eb-concordium (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
drsk0: @eb-concordium I researched a little and we could look up both arguments in the `bsoPrimeValidators` computation itself. It would just mean that we end up doing unnecessary work, because both arguments are already available at the call site. Not sure what's better, a simpler API or a faster algorithm.

drsk0: Turns out that `bsoGetBakerPoolRewardDetails` also returns the current validators in the keys. So I dropped them from the arguments and kept the threshold.

concordium-consensus/src/Concordium/KonsensusV1/Scheduler.hs
```diff
@@ -363,7 +392,37 @@ processPaydayRewards (Just PaydayParameters{..}) theState0 = do
     -- in which the rewards are distributed.
     foundationAddr <- getAccountCanonicalAddress =<< bsoGetFoundationAccount theState0
     theState1 <- doMintingP6 paydayMintRate foundationAddr theState0
-    distributeRewards foundationAddr paydayCapitalDistribution paydayBakers paydayPoolRewards theState1
+    theState2 <- distributeRewards foundationAddr paydayCapitalDistribution paydayBakers paydayPoolRewards theState1
+    case hasValidatorSuspension of
+        SFalse -> return theState2
+        STrue -> do
+            cps <- bsoGetChainParameters theState1
```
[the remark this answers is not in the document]
drsk0: changed.

Issue #1294, opened 2024-12-06
Implement Protocol 7->8 update


td202:
Implement the protocol update. Payload should be the new maximum missed rounds threshold.

* Redefine Regenesis for P8 to include migration parameter.
* Implement update Concordium.ProtocolUpdate.P7.ProtocolP8 and related modules.
* Ensure that migration sets the new maximum missed rounds from the threshold.

PR #1301, opened 2024-12-11
Introduce the P7 to P8 update.

eb-concordium:
## Purpose

To implement what is left of https://github.com/Concordium/concordium-node/issues/1294.
So far a dummy hash for the update is given. This MUST be updated when https://github.com/Concordium/concordium-update-proposals/pull/69 has been merged.

## Changes

- P7 to P8 update introduced.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/ProtocolUpdate/P7.hs
+++ b/concordium-consensus/src/Concordium/ProtocolUpdate/P7.hs
@@ -8,7 +8,7 @@ module Concordium.ProtocolUpdate.P7 (
     updateNextProtocolVersion,
 ) where
 
-import Control.Monad.State
+import Control.Monad.State hiding (get)
 import qualified Data.HashMap.Strict as HM
 import Data.Serialize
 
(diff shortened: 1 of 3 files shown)


concordium-consensus/src/Concordium/ProtocolUpdate/P7.hs
```diff
@@ -8,29 +8,33 @@ module Concordium.ProtocolUpdate.P7 (
...
 
+import qualified Concordium.Genesis.Data.P8 as P8
 import Concordium.GlobalState.BlockState
 import qualified Concordium.GlobalState.Persistent.BlockState as PBS
 import Concordium.GlobalState.Types
 import qualified Concordium.GlobalState.Types as GSTypes
 import Concordium.KonsensusV1.TreeState.Implementation
 import Concordium.KonsensusV1.TreeState.Types
+import qualified Concordium.ProtocolUpdate.P7.ProtocolP8 as ProtocolP8
 import qualified Concordium.ProtocolUpdate.P7.Reboot as Reboot
```
drsk0, line 27: I think this should be P8.Reboot.

concordium-consensus/src/Concordium/ProtocolUpdate/P7.hs
```diff
@@ -53,9 +57,11 @@ updateRegenesis ::
     BlockPointer (MPV m) ->
     m (PVInit m)
 updateRegenesis Reboot = Reboot.updateRegenesis
```
drsk0, line 59: this should reboot P8, not P7. I think the constraint in line 49 is probably also wrong.

drsk0 (changes requested):
I think there is a problem with rebooting the wrong protocol version. Other than that it looks good.

concordium-consensus/src/Concordium/ProtocolUpdate/P7.hs
eb-concordium: In this file we specify all the protocol updates that are possible from P7. Concretely we specify two of such updates, namely a P7 -> P7 update (the `Reboot` update) and a P7 -> P8 update (the update that introduces P8).

I there believe that we should indeed use ``P7.Reboot` and not `P8.Reboot` as you suggest.

drsk0: You're right, I sought I was looking at P8.hs.

drsk0 (approved):

td202 (approved):

PR #1307, opened 2025-01-10
Fix the computation of missed rounds on timeout.

td202:
## Purpose

Fix the computation of missed rounds on timeout.

## Changes

- Revise `computeMissedRounds` to use the QC and TC of a block to compute the missed rounds. The missed rounds will be from `qcRound + 1` to `tcRound` (inclusive) if the TC is present. (No missed rounds if the TC is absent.)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix a bug in computing the number of missed rounds in the event of a timeout.
+
 ## 8.0.0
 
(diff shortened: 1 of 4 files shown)


eb-concordium (approved):
Good catch!

drsk0 (commented):
This definition of missed rounds makes sense to me. @td202 what's the reasoning, that there are no other timeout certificates between the round of the QC and the observed TC?

PR #1320, opened 2025-01-22
Fix bug in calculation of absolute block height

### LINKED ISSUE #1319: BUG: Abolute block height may be incorrect on restart after protocol update
**Bug Description**

In certain circumstances, after a protocol update (in consensus version 1), when a node is restarted it may calculate absolute block heights incorrectly. This leads to the node rejecting blocks or producing incorrect blocks. If all nodes are restarted at or around the same time, then they may all begin producing erroneous blocks.

This issue arises from the start up process assuming that the last finalized block of a shut-down consensus is the terminal block. The terminal block is the block that is taken to be the parent of the genesis block for the new consensus. It is commonly the case that the last finalized block is the terminal block. However, if round timeouts occur around the time of the protocol update, the two may be distinct.

The terminal block is the first finalized block with a timestamp at/after the protocol update trigger time. Typically, a block is finalized by having a child block in the next round that has a QC. However, it is also possible for a block to be implicitly finalized by having a child that's not in the next round that is finalized (either explicitly or implicitly). If the terminal block is implicitly finalized in this way, then it won't end up being the last finalized block.

Related to this, the `GetBlockInfo` query on the regenesis block will return the parent as being the last finalized block of the previous genesis index, rather than the terminal block.

**Steps to Reproduce**

- Start a local network with 4 nodes with equal stake in protocol version 7.
- Take one of the nodes down to ensure that timeouts occur frequently.
- Execute a P7.Reboot update. The goal is that the round after the terminal block should time-out. If this does not happen, then the update can be repeated as necessary. (This may be observed in the logs, or by querying blocks at specific heights as described below.)
- Restart the 4th node and allow it to catch-up.
- Stop and restart one of the nodes.

To determine if the timeout occurred at the right point (note, this assumes the node has not been restarted after the protocol update) **These instructions are for 8.0.2**:
- Use `concordium-client consensus status` to get the current era genesis block.
- Use `concordium-client raw GetBlockInfo <current-era-genesis-block>` to observe the `blockParent`. This will be the last finalized block of the previous consensus. (Not necessarily the terminal block.)
- Use `concordium-client raw GetBlockSpecialEvents <last-finalized-block>`. If this shows `[]`, then a timeout occurred after the terminal block. Otherwise, there will always be at least one event with tag `BlockAccrueReward` (except for a genesis block).

**For a node with a fix for `GetBlockInfo`, the following instructions apply**:
- Use `concordium-client consensus status` to get the current era genesis block.
- Use `concordium-client raw GetBlockInfo <current-era-genesis-block>` to observe the `blockParent`. This should be the terminal block of the previous consensus.
- Use `concordium-client raw GetBlockInfo <terminal-block>` to observe the terminal block. In particular, note the `eraBlockHeight` and `genesisIndex`, as well as the `round`.
- Use `concordium-client raw GetBlocksAtHeight <eraBlockHeight+1> --genesis-index <genesisIndex> --restrict`. If the result is `[]` then the timeout did not occur at the desired point. Otherwise, note the hash of the block provided.
- Inspect the block with `GetBlockInfo` and observe that the round is more than `1+terminal block round`. (Otherwise, the timeout didn't occur, at least for that node.)

**Expected Result**

The restarted node should catch up and participate successfully.

**Actual Result**

The restarted node rejects blocks produced by its peers after it is restarted.

**Versions**
- Software Version 6.0.0 - 8.0.2
- OS: all

td202:
## Purpose

Closes #1319

- Fix a bug where, after a protocol update in consensus version 1 (P6 onwards), a node may miscalculate the absolute height of blocks when it is restarted. (#1319)
- Fix a bug where `GetBlockInfo` reports the parent block of a genesis block to be the last finalized block of the previous genesis index, instead of the terminal block.

## Changes

- In `startupSkov`, use the terminal block instead of the last finalized block for calculating the height of a new genesis.
- Revise `loadSkovData` to return the height terminal block in the case of a protocol update. Update the `ExistingSkov` data structure to include this (and drop the height of the last finalized block).
- Revise `getBlockInfo` to use the terminal block as the parent of a genesis block.
- Bump version to 8.0.3

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased changes
 
+## 8.0.3
+
+- Fix a bug where, after a protocol update in consensus version 1 (P6 onwards), a node may
+  miscalculate the absolute height of blocks when it is restarted. (#1319)
+- Fix a bug where `GetBlockInfo` reports the parent block of a genesis block to be the last
+  finalized block of the previous genesis index, instead of the terminal block.
+
 ## 8.0.2
 
 - Fix a bug where the P7->P8 protocol update affects payday timing.
--- a/concordium-consensus/src/Concordium/KonsensusV1/SkovMonad.hs
+++ b/concordium-consensus/src/Concordium/KonsensusV1/SkovMonad.hs
@@ -465,10 +465,9 @@ data ExistingSkov pv m = ExistingSkov
       esState :: !(SkovV1State pv),
       -- | The hash of the current genesis block.
       esGenesisHash :: !BlockHash,
-      -- | The (relative) height of the last finalized block.
-      esLastFinalizedHeight :: !BlockHeight,
-      -- | The effective protocol update if one has occurred.
-      esProtocolUpdate :: !(Maybe ProtocolUpdate)
+      -- | The effective protocol update if one has occurred, together with the relative
+      --  block height of the terminal block.
+      esProtocolUpdate :: !(Maybe (ProtocolUpdate, BlockHeight))
     }
 
 -- | Internal type used for deriving 'HasDatabaseHandlers' and 'LMDBAccountMap.HasDatabaseHandlers'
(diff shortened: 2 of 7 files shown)


drsk0 (approved):
Looks good!

PR #1322, opened 2025-02-10
NOD-525: Optimize baker finalization awake updates

drsk0:
Prevent redundant writes in bsoMarkFinalizationAwakeBakers by only updating  BakerPoolRewardDetails when a baker's finalization state actually changes. This  avoids unnecessary tree rehashing on each block.

Technical details:
- Added check before updating BakerPoolRewardDetails

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
@@ -3823,13 +3823,21 @@ doMarkFinalizationAwakeBakers pbs bids = do
         case binarySearchI bcBakerId bpc bid of
             Nothing -> return bprs
             Just (i, _) -> do
-                mBPRs <- LFMBT.update setAwake (fromIntegral i) bprs
+                mBPRs <- MTL.runExceptT $ LFMBT.update setAwake (fromIntegral i) bprs
                 case mBPRs of
-                    Nothing ->
+                    -- error is used to signal that there is no change
+                    Left () -> return bprs
+                    Right Nothing ->
                         error "Invariant violation: unable to find baker in baker pool reward details tree"
-                    Just ((), newBPRs) ->
+                    Right (Just ((), newBPRs)) ->
                         return newBPRs
-    setAwake bpr =
+    setAwake bpr = do
+        -- If there's nothing to do, use throwError to skip the update.
+        when
+            ( finalizationAwake bpr
+                && all (== emptySuspensionInfo) (suspensionInfo bpr)
+            )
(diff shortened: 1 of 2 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -3823,12 +3823,20 @@ doMarkFinalizationAwakeBakers pbs bids = do
...
+                        | finalizationAwake oldBpr
+                            && suspensionInfo oldBpr == (emptySuspensionInfo <$ suspensionInfo oldBpr) ->
+                            return bprs
+                        | otherwise -> do
+                            mBPRs <- LFMBT.update setAwake (fromIntegral i) bprs
+                            case mBPRs of
+                                Nothing -> invariantErr
+                                Just ((), newBPRs) ->
+                                    return newBPRs
+    invariantErr = error "Invariant violation: unable to find baker in baker pool reward details tree"
```
td202, line 3839: This is fine, but we could avoid doing a traversal twice with something like:
```haskell
    markFinalizerAwake bpc bprs bid = do
        case binarySearchI bcBakerId bpc bid of
            Nothing -> return bprs
            Just (i, _) -> do
                mBPRs <- runExceptT $ LFMBT.update setAwake (fromIntegral i) bprs
                case mBPRs of
                    Left () -> return bprs
                    Right Nothing ->
                        error "Invariant violation: unable to find baker in baker pool reward details tree"
                    Right (Just ((), newBPRs)) ->
                        return newBPRs
    setAwake bpr = do
        -- If there's nothing to do, use throwError to skip the update.
        when
            ( finalizationAwake bpr
                && all (== emptySuspensionInfo) (suspensionInfo bpr)
            )
            $ throwError ()
        return
            ( (),
              bpr
                { finalizationAwake = True,
                  suspensionInfo = emptySuspensionInfo <$ suspensionInfo bpr
                }
            )
```

td202 (approved):

td202:
If you haven't, could you double check that we have appropriate test coverage for this? e.g. if you break the test for skipping the update then it gets caught by some test.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
drsk0: Switched to this solution.

drsk0:
I added a test to `KonsensusV1/EpochTransition.hs`.

PR #1328, opened 2025-02-20
Be lazier when loading modules.

### LINKED ISSUE #1325: Module table improvements 1: Lazier module loading
Currently, when a `CachedModule` (a reference to a module) is loaded from the blob store, it will also load the module itself. This is because the module hash is loaded eagerly as part of the reference. Loading the module can be an expensive operation, even involving recompiling the module. However, the module reference (hash) can potentially be loaded quickly, as it is stored directly at the start of the module record. We should allow just loading the module hash when only the hash is needed. Additionally, when we want to get the module source, we should be able to bypass recompiling the module.

td202:
## Purpose

Closes #1325

This revises how the module table is loaded to avoid eagerly recompiling modules. While loading the modules does reconstruct the module map (from `ModuleRef`s to `ModuleIndex`es), it no longer fully loads each module to do so (which may incur the cost of recompiling the module, and also loads the module into the cache). Also, when querying the module source, we bypass recompiling the module and do not cache the module.

## Changes

- Introduce the `DirectBlobHashable` class. This provides one function `loadHash :: BlobRef a -> m h` that directly obtains a hash from a `BlobRef`. There is a blanket, overlappable implementation of this as `loadHash = getHashM <=< loadDirect`. This allows overlapping instances to be defined where there is a more efficient way to get the hash directly.
- Define an instance `DirectBlobHashable m Hash Module` that simply deserializes the module hash without loading the module (and potentially recompiling it).
- Factor `loadModuleDirect` out of the `loadDirect` implementation for `Module`. `loadModuleDirect` loads the module, but will not recompile it (as might be necessary to use the artifact). `loadDirect` will recompile if necessary.
- Define `openHashedCachedRef` that will return the value stored at the reference if it is in memory/cached, and otherwise return the `BlobRef` for it.
- Revise `getSource` to use `openHashedCachedRef` and `loadModuleDirect` to bypass recompiling the module.
- Define `mfoldRef` on `LFMBTree` to allow folding over the leaf references of the tree.
- Revise `load` for `HashedCachedRef` to use `loadHash` instead of loading and caching the value at the reference.
- Revise `load` for `Modules` to build the module map by using `mfoldRef` over the table and `getHashM` on the `HashedCachedRef`s in the leaves. This avoids loading the full module and only loads the hash.

Note, the change to the way `load` works for `HashedCachedRef` has implications for the accounts table. Specifically, when the account table fully loaded into memory, it won't cache each account.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
@@ -9,6 +9,7 @@ module Concordium.GlobalState.Persistent.CachedRef where
 
 import Control.Monad
 import Control.Monad.Trans
+import Data.Functor
 import Data.IORef
 import Data.Proxy
 import Data.Serialize (put)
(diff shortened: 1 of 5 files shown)


eb-concordium (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/CachedRef.hs
```diff
@@ -437,9 +438,32 @@ makeFlushedHashedCachedRef val = do
...
+    liftIO (readIORef hcrUnflushed) >>= \case
+        HCRMem val -> return (Right val)
+        HCRMemHashed val _ -> return (Right val)
+        HCRDisk r -> openHashedCachedRef r
+openHashedCachedRef HCRFlushed{..} =
+    lookupCachedValue (Proxy @c) hcrBlob <&> \case
+        Nothing -> Left hcrBlob
+        Just val -> Right val
+
+-- | Loading a 'HashedCachedRef' does not cache the value, but it does load the hash via 'loadHash'.
```
drsk0, line 462: So are we still caching them at all?

PR #1330, opened 2025-02-28
Correctly construct account difference map for certified blocks

### LINKED ISSUE #1329: BUG: Account difference map is not reconstructed correctly
**Bug Description**

When the node is started, the account difference map is reconstructed for certified blocks. This records accounts that have been created since the last finalized block. However, at present, while the difference map is constructed, it is not correctly updated in the block state of these certified blocks. This means that if those accounts are then used in transactions, the scheduler will reject them (as the accounts are not seen to exist).

This is likely to cause problems for nodes that get stopped while they have certified blocks that contain account creation transactions. However, this will only become apparent when the accounts are involved in other transactions. Moreover, if the node is subsequently restarted, it is possible the error will be detected and fixed as the number of accounts in the last finalized block will not match the number in the account table.

**Steps to Reproduce**

1. Set up a local chain with 5 equal validator nodes. (It may be helpful if the genesis time is far in the past, as this can result in blocks being in the certified state longer as epoch transitions occur.)
2. Send a number of account creations so that at least one account is created per block. The `create-initial-accounts` example in the rust SDK can do this. (Tool requires first creating a `created-accounts` directory to output the new account details.)
3. While these transactions are being included in blocks, shut down one of the nodes.
4. Stop the account creations.
5. Send a nominal transfer to each of the new accounts. (e.g. use `concordium-client account list` to get a list of all accounts, then the `transfer-to-accounts` example from the SDK to send to each of them.)
6. Restart the stopped node.

**Expected Result**

The node should catch up successfully with its peers and block production should continue.

**Actual Result**

The node eventually rejects a block that contains the transfer transactions and subsequent blocks are reported as `pending its parent`:

```
INFO: Konsensus: Block 0a38394f551aa1c917bb0ce6738924d5338ffd01048f42ddfbb12e0b1c24ce9a received.
INFO: Konsensus: Block 9e3ffdba160acf5036a6829852e4c9fc2a89b6b4ff5b389bf3fff74e4c596383 is pending its parent 0a38394f551aa1c917bb0ce6738924d5338ffd01048f42ddfbb12e0b1c24ce9a.
```

Restarting the node likely causes the node to successfully catch up again.

**Versions**

* Software Version: 6.2.1 - 8.0.3
* OS: all

td202:
## Purpose

Closes #1329 

This fixes a bug where the account difference map for certified blocks that contain account credential deployments is incorrect. The problem was caused by the fact that the block state was not cached before the account difference map was constructed. This meant that, while the difference map was reconstructed, it was dropped, since before caching only a `BlobRef` is held for the block state. By caching first, the `BlockStatePointers` are loaded into memory, which ensures that subsequent reads will use the same pointer for the account difference map.

## Changes

- In `loadCertifiedBlocks`, ensure that `cacheBlockState` is called before `reconstructAccountDifferenceMap`.
- Add a check that the account difference map is constructed correctly.
- Refine the documentation of `reconstructAccountDifferenceMap`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,13 @@
 # Changelog
 
 ## Unreleased changes
+
 - Replace `BufferedRef` with `HashedBufferedRef` in `PoolRewards`
   `bakerPoolRewardDetails::LFMBTree` field to cache computed hashes.
-
 - Improvements to the loading of modules. This particularly improves the performance of
   `GetModuleSource` in certain cases, and can also reduce start-up time.
+- Fix a bug that affects setting up the account map correctly for non-finalized certified blocks
+  that contain account creations (#1329).
 
 ## 8.0.3
 
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1608,6 +1608,8 @@ class (BlockStateOperations m, FixedSizeSerialization (BlockStateRef m)) => Bloc
     --
     --  Preconditions:
     --  * This function MUST only be called on a certified block.
+    --  * The block state MUST already be cached (with 'cacheBlockState'). Otherwise the update to
+    --    the difference map may not be retained.
     --  * This function MUST only be called on a block state that does not already
     --    have a difference map.
     --  * The provided list of accounts MUST correspond to the accounts created in the block,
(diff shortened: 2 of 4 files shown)


drsk0 (approved):
The fix looks good. I wonder if we could  avoid this class of subtle bugs, by enforcing to load the block state from cache for any ephemeral data.

eb-concordium (approved):

PR #1353, opened 2025-03-19
Delay handler execution till after state update

### LINKED ISSUE #1352: BUG: Data race can lead to crash from GetBlockInfo
**Bug Description**

It is possible that, during a protocol update, a GRPC call to `GetBlockInfo` can lead to the error "Updated consensus is missing terminal block." This results in the node crashing. The issue is that during the protocol update, the new consensus state is written before the old consensus state is updated. The `GetBlockInfo` query expects that if the new state is present, then the old consensus must have a terminal block. However, if the query executes between the new consensus state being written and the old state being updated, then it will not see the terminal block for the old consensus.

The following scenario describes how this can occur:

* The node receives a finalization message; `receiveFinalizationMessage` is called.
* The write lock is acquired.
* The finalization message is successfully received in the current consensus state.
* A thread is forked to process the finalization message, sampling the current consensus state.
* Processing the finalization message triggers finalization of the terminal block. This invokes `onFinalize` which eventually invokes `checkForProtocolUpdateV1`.
* Check for protocol update instantiates a new consensus and begins shutting down the old consensus.
* The new consensus state is added to `mvVersions`.
* *At this point, the old consensus state has not been updated in its* `vc1State` `IORef`*.*
* `GetBlockInfo` is called to query the last finalized block.
  * The new consensus state is available, so it is run against that, getting the genesis block of the new consensus.
  * To determine the parent block, it tries to query the terminal block of the previous consensus version. However, this has not yet been written. Consequently the `Updated consensus is missing terminal block.` error is raised, causing the node to crash.

**Steps to Reproduce**

(Repeatedly) invoke `GetBlockInfo` (querying the best or last finalized block) while the node is executing a protocol update. Note, it may require a significant amount of luck to trigger the bug.

**Expected Result**

There is no error.

**Actual Result**

```
2025-03-17 11:02:04.744	
FATAL: exception not rethrown
	
	
2025-03-17 11:02:04.733	
concordium-mainnet-node-8.0.3: receiveFinalizationMessage: interrupted
	
	
2025-03-17 11:02:04.704	
[DEBUG concordium_node::plugins::consensus] Regenesis occurred; marking all peers as pending.
	
	
2025-03-17 11:02:04.630	
[DEBUG concordium_node::consensus_ffi::ffi] Konsensus: Advancing round: round 5541608 certified block 13b82ef6b945daa8e36e905fd9adb4e6fc18efe98612984303f7b2c10fe15b91
	
	
2025-03-17 11:02:04.630	
  error, called at src/Concordium/Queries.hs:761:47 in concordium-consensus-0.1.0.0-BOf50nu8VUQBAbmJmSPmvN:Concordium.Queries
	
	
2025-03-17 11:02:04.630	
CallStack (from HasCallStack):
	
	
2025-03-17 11:02:04.630	
concordium-mainnet-node-8.0.3: Updated consensus is missing terminal block.
	
	
2025-03-17 11:02:04.629	
[DEBUG tower_http::trace::make_span] request; method=POST uri=http://localhost:20000/concordium.v2.Queries/GetBlockInfo version=HTTP/2.0
	
	
...
	
	
2025-03-17 11:00:29.471	
[INFO  concordium_node::consensus_ffi::ffi] Kontrol: Starting protocol update.
	
	
2025-03-17 11:00:29.471	
[INFO  concordium_node::consensus_ffi::ffi] Konsensus: Block 0f582b5850c511cbb8abc4d3ee9392ea6770d329b89dd4d711fbd2c6e8387056 (round 5541607) finalized at height 5418563
	
	
2025-03-17 11:00:29.471	
[INFO  concordium_node::consensus_ffi::ffi] Konsensus: Block 76b9b415063f647dda7c639bdb9a47cb5c74fcb71169127438e4defa388bb510 (round 5541605) finalized at height 5418562
	
	
2025-03-17 11:00:29.469	
[INFO  concordium_node::consensus_ffi::ffi] Konsensus: Shutdown triggered in block 0f582b5850c511cbb8abc4d3ee9392ea6770d329b89dd4d711fbd2c6e8387056 (round 5541607) finalized at height 5418563
	
	
2025-03-17 11:00:29.467	
[DEBUG concordium_node::consensus_ffi::ffi] Konsensus: Quorum certificate generated for block 13b82ef6b945daa8e36e905fd9adb4e6fc18efe98612984303f7b2c10fe15b91 in round 5541608
	
	
2025-03-17 11:00:29.467	
[DEBUG concordium_node::plugins::consensus] Sent a broadcast containing a finalization message
	
	
2025-03-17 11:00:29.466	
[DEBUG concordium_node::plugins::consensus] Processed a finalization message from 0000000000011ee1
```

**Versions**

* Software Version: 8.0.3
* OS: all

td202:
## Purpose

Closes #1352

This defers the execution of on-block and on-finalization handlers until after the consensus state has been committed. This in particular ensures that in the case of a protocol update, the consensus state is committed before the protocol update occurs (as this is part of the on-finalization handling). Consequently, a `GetBlockInfo` that queries the genesis block of a new consensus (created in the protocol update) will always be able to see the terminal block (which is available once the consensus state is committed).

This also ensures that if gRPC clients are notified of block arrivals, queries about those blocks will not fail, which was previously a theoretical possibility.

## Changes

- The `SkovT` implementation uses the `MonadWriter` aspect to produce a sequence of `HandlerEvent`s rather than directly invoking a handler callback.
- The handlers `onBlockHandler`, `onFinalizationHandler` and `onPendingLiveHandler` are removed from the `HandlerContext`, as the events will be used instead. (`onPendingLiveHandler` was not used anyway.)
- The handler implementations are promoted from being part of the `HandlerContext` implementation to separate functions `skovV1BlockHandler` and `skovV1FinalizeHandler`.
- These handlers are called as part of `liftSkovV1Update`, and are run after updating the state.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/MultiVersion.hs
+++ b/concordium-consensus/src/Concordium/MultiVersion.hs
@@ -258,11 +258,9 @@ data Callbacks = Callbacks
 --  be populated with a configuration for the genesis index prior to calling this function.
 skovV1Handlers ::
     forall pv finconf.
-    (IsConsensusV1 pv, IsProtocolVersion pv) =>
     GenesisIndex ->
-    AbsoluteBlockHeight ->
     SkovV1.HandlerContext pv (MVR finconf)
-skovV1Handlers gi genHeight = SkovV1.HandlerContext{..}
+skovV1Handlers gi = SkovV1.HandlerContext{..}
   where
     _sendTimeoutHandler timeoutMsg = MVR $ \mvr -> do
         broadcastFinalizationMessage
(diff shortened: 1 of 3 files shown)


concordium-consensus/src/Concordium/MultiVersion.hs
```diff
@@ -280,50 +278,63 @@ skovV1Handlers gi genHeight = SkovV1.HandlerContext{..}
...
-                        liftIO (notifyCallback (getHash bp) height isHomeBaked)
-        checkForProtocolUpdateV1
-
-    _onPendingLiveHandler = do
-        -- Notify peers of our catch-up status, since they may not be aware of the now-live pending
-        -- blocks.
-        bufferedSendCatchUpStatus
+                forM_ finalizedBlocks $ \bp -> do
+                    let height = localToAbsoluteBlockHeight genHeight (SkovV1.blockHeight bp)
+                    let isHomeBaked = case nodeBakerIdMaybe of
```
drsk0, line 331: could we factor this out and reuse in `skovV1BlockHandler`

drsk0 (approved):

eb-concordium (approved):

drsk0:
@td202 are there tests for this?

PR #1354, opened 2025-03-21
Prepare protocol version 9

eb-concordium:
## Purpose

Prepare for Protocol 9 in the node. 

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
@@ -129,6 +129,13 @@ migratePersistentBakerInfoEx StateMigrationParametersP7ToP8{} = migrateReference
         BakerInfoEx av1 ->
         m' (BakerInfoEx av2)
     migrateBakerInfoExV1 BakerInfoExV1{..} = return BakerInfoExV1{_bieIsSuspended = CTrue False, ..}
+migratePersistentBakerInfoEx StateMigrationParametersP8ToP9{} = migrateReference migrateBakerInfoExV1
+  where
+    migrateBakerInfoExV1 ::
+        (AVSupportsDelegation av1, AVSupportsDelegation av2, SupportsValidatorSuspension av2 ~ 'True, Monad m') =>
+        BakerInfoEx av1 ->
+        m' (BakerInfoEx av2)
+    migrateBakerInfoExV1 BakerInfoExV1{..} = return BakerInfoExV1{_bieIsSuspended = CTrue False, ..}
 
 -- | Migrate a 'V0.PersistentBakerInfoEx' to a 'PersistentBakerInfoEx'.
 --  See documentation of @migratePersistentBlockState@.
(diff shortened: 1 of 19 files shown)


concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -129,6 +129,13 @@ migratePersistentBakerInfoEx StateMigrationParametersP7ToP8{} = migrateReference
         BakerInfoEx av1 ->
         m' (BakerInfoEx av2)
     migrateBakerInfoExV1 BakerInfoExV1{..} = return BakerInfoExV1{_bieIsSuspended = CTrue False, ..}
+migratePersistentBakerInfoEx StateMigrationParametersP8ToP9{} = migrateReference migrateBakerInfoExV1
+  where
+    migrateBakerInfoExV1 ::
+        (AVSupportsDelegation av1, AVSupportsDelegation av2, SupportsValidatorSuspension av1 ~ SupportsValidatorSuspension av2, Monad m') =>
```
drsk0, line 135: Given you want to migrate from P8 -> P9, maybe the constraint on validator suspension should read that both protocol support it.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -218,6 +225,34 @@ migratePersistentAccountStakeEnduring PersistentAccountStakeEnduringBaker{..} =
...
     return $! PersistentAccountStakeEnduringDelegator{..}
 
+migratePersistentAccountStakeEnduringAV3 ::
+    forall m t av1 av2.
+    ( SupportMigration m t,
+      IsAccountVersion av1,
+      IsAccountVersion av2,
+      AVSupportsDelegation av1,
+      AVSupportsDelegation av2,
+      AVSupportsValidatorSuspension av1 ~ AVSupportsValidatorSuspension av2
```
drsk0, line 235: Same for this constraint.

concordium-consensus/src/Concordium/GlobalState/Persistent/Bakers.hs
```diff
@@ -156,7 +156,8 @@ migratePersistentEpochBakers migration PersistentEpochBakers{..} = do
                 SomeParam $ P6.updateFinalizationCommitteeParameters migrationProtocolUpdateData
             StateMigrationParametersP6ToP7{} -> _bakerFinalizationCommitteeParameters
             StateMigrationParametersP7ToP8{} ->
-                SomeParam $ unOParam $ _bakerFinalizationCommitteeParameters
+                SomeParam $ unOParam _bakerFinalizationCommitteeParameters
```
drsk0, line 159: Don't we have a linter in the CI that should catch these things?

drsk0 (approved):
Looks good, thanks!

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
eb-concordium: Well, the constrains are there, because they are needed, and in this case what is needed is `SupportsValidatorSuspension av1 ~ SupportsValidatorSuspension av2`. 

If we want to be concrete we could just contraint that `av1 ~ 'AccountV4, av2 ~ 'AccountV5` since this is necessarily the case for `StateMigrationParametersP8ToP9` and likewise above. 

So I think we should either keep the generality or make it concrete.

concordium-consensus/src/Concordium/GlobalState/Persistent/Bakers.hs
eb-concordium: I don't think the CI will complain about too many $'s

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
drsk0: To me it would make more sense for the constraint to be more narrow, the same as for delegation. I guess it's a general question. Should we make the constraints as concrete as possible or as general as possible.

eb-concordium: I realized this case just boils down to `migrateReference (return . coerceBakerInfoExV1)` which we also do in other places in the file.

td202 (approved):
Looks good. Please merge with the `plt` feature branch instead of main.

PR #1362, opened 2025-04-09
Use binary flatc in CI

td202:
## Purpose

Download a binary image of `flatc` instead of building it from source in the CI pipeline. This should reduce the runtime for the pipeline by a few (~3-4) minutes.


drsk0 (approved):
:crossed_fingers:

eb-concordium (approved):

PR #1365, opened 2025-04-10
Add PLT block state operations for token-level state

td202:
## Purpose

This adds the global PLT index to the block state (for P9) and adds block state operations for accessing it.

## Changes

- Add protocol level tokens to `BlockStatePointers`.
- Include protocol level tokens in `BlockStateHash` calculation.
- Define and implement `PLTQuery` type class that provides the following operations for getting global PLT state: `getPLTList`, `getTokenIndex`, `getTokenState`, `getTokenConfiguration` and `getTokenCirculatingSupply`. The type class is parametrised by the block state so that it can be used for both `BlockState` and `UpdatableBlockState`.
- Add `bsoSetTokenState`, `bsoSetTokenCirculatingSupply` and `bsoCreateToken` to `BlockStateOperations`.
- Since `MonadProtocolVersion` is a constraint for `PLTQuery`, many `MonadProtocolVersion` constraints are now redundant and have been removed.
- Add derived `PLTQuery` instances where necessary (`ActiveFinalizationM`, `MaybeT`, `ExceptT`, `PersistentTreeStateMonad`, `SkovV1T`, `InitMonad`, `TestMonad`, `AccountNonceQueryT`, `SkovQueryMonadT`, `SkovT`, `PersistentBSM` (testing)). These are all cases where `BlockStateQuery` is already derived.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState.hs
```diff
@@ -4293,6 +4305,58 @@ doGetPrePreCooldownAccounts pbs = case sSupportsFlexibleCooldown sav of
   where
     sav = sAccountVersionFor (protocolVersion @pv)
 
+-- | Set the token-level state of a token for a given 'TokenStateKey'. If the value is
+--  @Nothing@, the key removed from the token state. Otherwise the key is mapped to the
```
eb-concordium, line 4309:
```suggestion
--  @Nothing@, the key is removed from the token state. Otherwise the key is mapped to the
```

eb-concordium (approved):

limemloh (approved):

PR #1368, opened 2025-04-15
Basic support for token holder transactions in scheduler (close NOD-697)

limemloh:
## Purpose

Basic support for token holder transactions in scheduler

Related PR for grpc: https://github.com/Concordium/concordium-grpc-api/pull/88
Related PR for base https://github.com/Concordium/concordium-base/pull/621

## Changes

- Add handler in scheduler for token holder transactions.

### DIFF
--- a/concordium-consensus/src/Concordium/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/Scheduler.hs
@@ -101,7 +101,6 @@ import Lens.Micro.Platform
 import qualified Concordium.GlobalState.ContractStateV1 as StateV1
 import Concordium.Scheduler.WasmIntegration.V1 (ReceiveResultData (rrdCurrentState))
 import Concordium.Types.Accounts
-import Concordium.Types.Execution (RejectReason (..))
 import Concordium.Wasm (IsWasmVersion)
 import qualified Concordium.Wasm as GSWasm
 import Data.Proxy
--- a/concordium-consensus/src/Concordium/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/Scheduler.hs
@@ -2691,7 +2691,7 @@ handleTokenHolder depositContext tokenId tokenOperations =
         -- Fetch the token configuration to find the reference for the token module.
         configuration <- lift $ getTokenConfiguration tokenIndex
         let moduleRef = Token._pltModule configuration
-        -- TODO Tick energy for loading the module into memory based on the module size.
+        -- TODO Tick energy for loading the module into memory based on the module size. (Issue https://linear.app/concordium/issue/COR-1337)
         -- Invoke the token module with operations.
         events <- invokeTokenHolderOperations moduleRef tokenIndex senderAccount tokenOperations
         -- Map the produced events adding the token ID.
--- a/concordium-consensus/src/Concordium/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/Scheduler.hs
@@ -75,6 +75,7 @@ import Concordium.GlobalState.BlockState (
     ModuleQuery (..),
     NewInstanceData (..),
  )
+import qualified Concordium.GlobalState.Persistent.BlockState.ProtocolLevelTokens as Token
 import Concordium.GlobalState.Types
 
 import Concordium.Logger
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -2656,6 +2664,49 @@ handleUpdateCredentialKeys wtc cid keys sigs =
...
+        tickEnergy Cost.tokenHolderBaseCost
+        -- Fetch the token from state
+        tokenIndex <-
+            lift (getTokenIndex tokenId) >>= \case
+                Just tokenIndex -> return tokenIndex
+                Nothing -> rejectTransaction $ NonExistentTokenId tokenId
+        -- Fetch the token configuration to find the reference for the token module.
+        configuration <- lift $ getTokenConfiguration tokenIndex
+        let moduleRef = Token._pltModule configuration
+        -- TODO Tick energy for loading the module into memory based on the module size.
```
drsk0, line 2694: Would be nice to reference and track these in the backlog.

drsk0 (approved):
We can't test this, but maybe we should track the missing tests in the backlog as well.

concordium-consensus/src/Concordium/Scheduler.hs
limemloh: I have created [this task](https://linear.app/concordium/issue/COR-1337/tick-energy-for-loading-wasm-token-module) and added a reference as well in the comment.

limemloh:
> We can't test this, but maybe we should track the missing tests in the backlog as well.

I have created this [task about it](https://linear.app/concordium/issue/COR-1338/implement-scheduler-tests-for-token-holder-transactions)

DOBEN (approved):

concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -2656,6 +2664,49 @@ handleUpdateCredentialKeys wtc cid keys sigs =
...
+        tokenIndex <-
+            lift (getTokenIndex tokenId) >>= \case
+                Just tokenIndex -> return tokenIndex
+                Nothing -> rejectTransaction $ NonExistentTokenId tokenId
+        -- Fetch the token configuration to find the reference for the token module.
+        configuration <- lift $ getTokenConfiguration tokenIndex
+        let moduleRef = Token._pltModule configuration
+        -- TODO Tick energy for loading the module into memory based on the module size. (Issue https://linear.app/concordium/issue/COR-1337)
+        -- Invoke the token module with operations.
+        events <- invokeTokenHolderOperations moduleRef tokenIndex senderAccount tokenOperations
```
td202, line 2696: It would be good if this also handled the failure case.

concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -2656,6 +2664,49 @@ handleUpdateCredentialKeys wtc cid keys sigs =
...
+        -- Charge the base cost for this transaction type
+        tickEnergy Cost.tokenHolderBaseCost
+        -- Fetch the token from state
+        tokenIndex <-
+            lift (getTokenIndex tokenId) >>= \case
+                Just tokenIndex -> return tokenIndex
+                Nothing -> rejectTransaction $ NonExistentTokenId tokenId
+        -- Fetch the token configuration to find the reference for the token module.
+        configuration <- lift $ getTokenConfiguration tokenIndex
+        let moduleRef = Token._pltModule configuration
```
td202, line 2693: I'm not sure we should do this here. At the moment, it's not used, and I think it could be handled further down the stack when it eventually is.

td202 (approved):

concordium-consensus/src/Concordium/Scheduler.hs
limemloh: `invokeTokenHolderOperations` can reject the transaction directly.
Also not sure what type of reject reasons we intend for the token modules.

concordium-consensus/src/Concordium/Scheduler.hs
limemloh: That is also an option. Don't really have an opinion on that, so feel free to change it as we introduce the token module.

limemloh: I mainly wanted to prepare for charging the cost of loading the module, but that mights as well be further down the stack as well

PR #1369, opened 2025-04-17
initial plt kernel

drsk0:
Draft PR to get feedback on plt kernel design.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


concordium-consensus/src/Concordium/GlobalState/Account.hs
```diff
@@ -135,6 +136,47 @@ instance HashableTo TokenStateTableHash InMemoryTokenStateTable where
...
+      tasModuleStateDelta :: ![(TokenAccountStateKey, TokenAccountStateValueDelta)]
+    }
+    deriving (Eq, Show)
+
+newtype TokenAmountDelta = TokenAmountDelta {tokenAmountDelta :: Integer} deriving (Eq, Show)
+
+data TokenAccountStateValueDelta
+    = TASVDelete
+    | TASVCreate TokenAccountStateValue
+    | TASVUpdate TokenAccountStateValue
```
td202, line 150: I don't think it's worthwhile to distinguish these cases.

concordium-consensus/src/Concordium/GlobalState/Account.hs
```diff
@@ -601,14 +643,16 @@ data AccountUpdate = AccountUpdate
       -- | Optionally update the locked stake on the account by adding scheduled releases.
       --  Each entry in the list MUST have a non-empty list of releases, and the releases must be
       --  in ascending order of timestamp.
-      _auReleaseSchedule :: !(Maybe [([(Timestamp, Amount)], TransactionHash)])
+      _auReleaseSchedule :: !(Maybe [([(Timestamp, Amount)], TransactionHash)]),
+      -- | Optionally update the PLT state of an account.
+      _auTokenAccountStateTableDelta :: !(Maybe [(TokenIndex, TokenAccountStateDelta)])
```
td202, line 648: I am not sure that it's best to add the token state updates into `AccountUpdate` like this. (I'm already a bit uncomfortable with the release schedule additions being part of this.) I think the main benefit of doing this would be that it allows batching multiple operations which can potentially reduce look-ups and maybe hashing (if it's not done lazily). However, I don't think we're likely to batch account PLT state updates with other operations, so this seems moot. On the other hand, it forces all types of update through a single function, which results in `updateAccount` being large and potentially more difficult to maintain.  My suggestion would be to have one or more separate functions for updating the token state.

drsk0: That would mean adding new block state operations for updating the token account state?

PR #1370, opened 2025-04-22
Token Module: PLT Initialization

td202:
## Purpose

Closes [COR-698](https://linear.app/concordium/issue/COR-698/token-module-basic-plt-initialization)

Depends on: https://github.com/Concordium/concordium-grpc-api/pull/90
Depends on: https://github.com/Concordium/concordium-base/pull/623

Implement PLT initialization in the Token Module.

## Changes

- Add the governance account index to the `PLTConfiguration`.
- `handleCreatePLT` invokes token module.
- `SchedulerMonad` functions:
  - `withBlockStateRollback`: execute some operation allowing for rolling back the block state changes in the event of failure.
  - `runPLT`: run operations that use the PLT kernel.
  - `createToken`: create a new PLT with the specified configuration.
- Introduce token kernel classes and provide an implementation based on the scheduler.
- Token module implementation of `initializeToken`.
- Testing for the token module implementation.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Scheduler.hs
+++ b/concordium-consensus/src/Concordium/Scheduler.hs
@@ -2860,41 +2860,32 @@ handleChainUpdate (WithMetadata{wmdData = ui@UpdateInstruction{..}, ..}, maybeVe
 -- Unlike the other chain updates there is no support for queuing the update and the effective time
 -- is required to be zero.
 handleCreatePLT :: (SchedulerMonad m, PVSupportsPLT (MPV m)) => UpdateHeader -> CreatePLT -> m (Either FailureKind ValidResult)
-handleCreatePLT updateHeader payload =
-    if updateEffectiveTime updateHeader /= 0
-        then return $ Left InvalidUpdateTime -- TODO: Since this is invalid, the transaction verifier should be updated to reject this
-        else do
-            -- TODO Check signature when relevant update keys are introduced (Issue https://linear.app/concordium/issue/NOD-701).
-            -- TODO Verify the TokenId not already exists.
-            -- verify the governance account exists.
-            let governanceAccountAddress = payload ^. cpltGovernanceAccount
-            maybeGovernanceAccount <- getStateAccount governanceAccountAddress
-            case maybeGovernanceAccount of
-                Nothing -> do
-                    return $ Right $ TxReject $ InvalidAccountReference governanceAccountAddress
-                Just (governanceAccountIndex, _governanceAccount) -> do
-                    let tokenSymbol = payload ^. cpltTokenSymbol
-                    getTokenIndex tokenSymbol >>= \case
-                        Just _existingTokenIndex -> do
-                            return $ Right $ TxReject $ TokenExists $ tokenSymbol
-                        Nothing -> do
-                            createResult <- withBlockStateRollback $ do
-                                let config =
-                                        PLTConfiguration
-                                            { _pltTokenId = tokenSymbol,
-                                              _pltModule = payload ^. cpltTokenModule,
-                                              _pltDecimals = payload ^. cpltDecimals,
-                                              _pltGovernanceAccountIndex = governanceAccountIndex
-                                            }
-                                tokenIx <- createToken config
-                                runPLT tokenIx $ TokenModule.initializeToken (payload ^. cpltInitializationParameters)
-                            return $ Right $ case createResult of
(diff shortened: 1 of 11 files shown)


concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
```diff
@@ -0,0 +1,378 @@
...
+import Concordium.GlobalState.Persistent.BlockState.ProtocolLevelTokens
+import Concordium.Scheduler.ProtocolLevelTokens.Kernel
+import Concordium.Scheduler.ProtocolLevelTokens.Module (InitializeTokenError (..), initializeToken)
+import Concordium.Types.ProtocolLevelTokens.CBOR
+import Concordium.Types.Tokens
+
+-- | A value of type @PLTKernelQueryCall acct ret@ represents an invocation of an operation
+--  in a 'PLTKernelQuery' monad where @PLTAccount m ~ acct@. The parameter @ret@ is the type
+--  of the return value expected.
+data PLTKernelQueryCall acct ret where
```
limemloh, line 35: **Question (non-blocking):** Can we get away with hardcoding `AccountIndex` for `acct`?

Do we expect any other type for `acct`?

concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
```diff
@@ -0,0 +1,378 @@
...
+-- | Assert that calling a given operation results in the given trace.
+assertTrace :: (Eq ret, Show ret) => TraceM call (Either TraceException ()) ret ret -> Trace call ret -> IO ()
+assertTrace op trace = case checkTrace trace op of
+    Left e -> assertFailure e
+    Right () -> return ()
+
+-- | Helper to generate a trace that aborts as a result of a 'PLTError' call.
+abortPLTError :: (Show e, Show acct, Typeable e, Typeable acct) => e -> Trace (PLTCall e acct) ret
+abortPLTError = Abort . AbortCall . PLTF . PLTError
+
```
limemloh, line 252: **Suggestion (non-blocking):** Consider moving the code above here into a separate module, separating it from the test spec (or move the code below into a separate spec file)

concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -2844,44 +2852,49 @@ handleChainUpdate (WithMetadata{wmdData = ui@UpdateInstruction{..}, ..}, maybeVe
...
 -- | Handler for processing chain update creating a new protocol level token.
 --
 -- Unlike the other chain updates there is no support for queuing the update and the effective time
 -- is required to be zero.
-handleCreatePLT :: (SchedulerMonad m) => UpdateHeader -> CreatePLT -> m TxResult
+handleCreatePLT :: (SchedulerMonad m, PVSupportsPLT (MPV m)) => UpdateHeader -> CreatePLT -> m (Either FailureKind ValidResult)
 handleCreatePLT updateHeader payload =
     if updateEffectiveTime updateHeader /= 0
-        then return $ TxInvalid InvalidUpdateTime
+        then return $ Left InvalidUpdateTime -- TODO: Since this is invalid, the transaction verifier should be updated to reject this
```
limemloh, line 2865: This makes sense, do we have a task for that? Maybe link to it from here

limemloh (approved):
Great work

concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -2844,44 +2852,49 @@ handleChainUpdate (WithMetadata{wmdData = ui@UpdateInstruction{..}, ..}, maybeVe
...
-                    --             TxInvalid $
-                    --                 NonExistentAccount governanceAccountAddress
-                    --     Just tokenModuleInterface -> do
-                    --         -- Create new PLT in token state with initial state.
-                    --         tokenState <- createToken tokenModuleInterface governanceAccountState
-                    --         -- Invoke the initialization call of token module with init state and the initialization parameter and collect the events for the summary.
-                    --         invokeInitializeToken tokenState
-                    --         buildValidTxSummary
-                    error "Not implemented yet"
+                Nothing -> do
```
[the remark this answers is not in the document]
td202: I think there's a perception at least that `ExceptT` is inefficient because bind always pattern-matches even when the preceding action cannot fail. (Compared to a continuation-based approach.) I think that's not really a worry here, so I'll use `ExceptT`.

PR #1372, opened 2025-04-23
Persistent account level state

drsk0:
This adds protocol level token state to the accounts.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


drsk0:
I'm working on the tests while waiting for the review.

limemloh (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/ProtocolLevelTokens.hs
```diff
@@ -0,0 +1,82 @@
...
+instance (MonadBlobStore m) => MHashableTo m Hash.Hash TokenAccountStateTable where
+    getHashM (TokenAccountStateTable tast) = do
+        bs <-
+            mapM
+                ( \(tIx, ref) -> do
+                    h <- getHashM ref
+                    return (Hash.hashOfHashes (getHash tIx) h)
+                )
+                $ Map.toAscList tast
+        return $ hashAsLFMBTV1 (getHash BS.empty) bs
```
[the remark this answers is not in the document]
drsk0: The usual style in the Haskell community is to use abbreviations in lambdas, in particular when they are small like this one. I personally find this actually easier to read than long variable names. I try to choose abbreviations that should be immediately clear from the context. 
I think you're right however in that `bs` should probably better be named to something like `hashes`

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
```diff
@@ -1019,6 +1022,8 @@ data PersistentAccount av = PersistentAccount
       -- | The actively staked balance of the account.
       --  INVARIANT: This is 0 if the account is not a baker or delegator.
       accountStakedAmount :: !Amount,
+      -- | The state table of the protocol level tokens of the account in ascending order of the TokenIndex.
+      accountTokenStateTable :: !(Conditionally (SupportsPLT av) (HashedBufferedRef TokenAccountStateTable)),
```
td202, line 1026: I think it might be better to use `Nullable (HashedBufferedRef TokenAccountStateTable)` with `Null` being used for the empty account state table, since this will reduce storage overhead for the many accounts that have no PLTs. I'd be fine with opening an issue for this instead of addressing it in this PR.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/ProtocolLevelTokens.hs
```diff
@@ -0,0 +1,83 @@
...
+import Data.Serialize
+
+-- | The table of PLT account states. The table is indexed by the token index
+--  into the global token table.
+newtype TokenAccountStateTable = TokenAccountStateTable
+    { tokenAccountStateTable :: Map.Map TokenIndex (HashedBufferedRef TokenAccountState)
+    }
+    deriving newtype (Show)
+
+instance (MonadBlobStore m) => MHashableTo m Hash.Hash TokenAccountStateTable where
```
td202, line 29: Should hash to `TokenStateTableHash` instead of `Hash.Hash`.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
td202: It may also be useful to use an `EagerlyHashedBufferedRef` so that it is never necessary to read from the blob store to compute the hash of an account that is loaded (i.e. there can be a straight `HashableTo` implementation). This can also be deferred if we create an issue to reassess exactly how we want to hash and store the PLT state on an account.

td202 (approved):
Looks good.

For the tests, I don't think there are actually good existing tests for the `Account`. I would add them in a new `GlobalStateTests.Account` probably. Some functionality from `GlobalStateTests.Accounts` could be used as a starting point (though e.g. `randomizeAccount` does not cover a lot of the features of accounts).

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/ProtocolLevelTokens.hs
drsk0: done.

concordium-consensus/src/Concordium/GlobalState/Basic/BlockState/Account.hs
```diff
@@ -31,6 +35,21 @@ import Concordium.Types
...
+newtype InMemoryTokenStateTable = InMemoryTokenStateTable
+    { inMemoryTokenStateTable :: Map.Map TokenIndex TokenAccountState
+    }
+    deriving (Eq, Show)
+
+instance HashableTo TokenStateTableHash InMemoryTokenStateTable where
+    getHash (InMemoryTokenStateTable m) =
+        TokenStateTableHash $
+            hashAsLFMBTV1
+                (getHash ("EmptyTokenAccountStateTable" :: BS.ByteString))
```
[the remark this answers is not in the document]
drsk0: I factored the constant out. I'd prefer factoring the shared implementation out, but I can't put it in `GlobalState/Accounts.hs`, because it would yield cyclic imports (because the TokenIndex) type is used.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/StructureV1.hs
drsk0: I created #1373 and #1374 for this.

PR #1375, opened 2025-04-25
Fix missing increment update sequence number

limemloh:
## Purpose

Closes COR-1343

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1523,7 +1523,7 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
 
     -- | Increment the update sequence number for Protocol Level Tokens (PLT).
     -- Unlike the other chain updates this is a separate function, since there is no queue associated with PLTs.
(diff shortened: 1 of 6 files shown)


concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1521,6 +1521,10 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
         (UpdateValue (ChainParametersVersionFor (MPV m))) ->
         m (UpdatableBlockState m)
 
+    -- | Increment the update sequence number for Protocol Level Tokens (PLT).
+    -- Unlike the other chain updates this is a separate function, since there is no queue associated with PLTs.
+    bsoIncrementPltUpdateSequenceNumber :: UpdatableBlockState m -> m (UpdatableBlockState m)
```
td202, line 1526: Minor gripe: I'd prefer the style `bsoIncrementPLTUpdateSequenceNumber`, which is more consistent with what we use elsewhere.

td202 (approved):
Looks good.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
limemloh: I'll fix

PR #1377, opened 2025-04-29
Fetch account tokens in AccountInfo gRPC query

limemloh:
## Purpose

Closes COR-680

### DIFF
--- a/concordium-consensus/src/Concordium/Queries.hs
+++ b/concordium-consensus/src/Concordium/Queries.hs
@@ -47,6 +47,7 @@ import Concordium.Types.IdentityProviders
 import Concordium.Types.Parameters
 import Concordium.Types.Queries hiding (PassiveCommitteeInfo (..), bakerId)
 import qualified Concordium.Types.Queries.KonsensusV1 as QueriesKonsensusV1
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/Queries.hs
```diff
@@ -1244,8 +1248,28 @@ getAccountInfoHelper getASI getCooldowns acct bs = do
...
+                                }
+                    return
+                        Tokens.Token
+                            { tokenId = BlockState._pltTokenId pltConfiguration,
+                              tokenAccountState =
+                                Tokens.TokenAccountState
+                                    { balance = accountBalance,
+                                      -- TODO Support allow/deny list state in account info query (Issue https://linear.app/concordium/issue/COR-1349)
+                                      memberAllowList = False,
+                                      memberDenyList = False
```
td202, line 1269: I'm fine to leave it for now, but allow list and deny list shouldn't be part of the protobuf state, but we should have a CBOR-encoded state representation instead that includes them if relevant.

td202 (approved):

drsk0 (approved):

PR #1385, opened 2025-05-06
`GetAccountInfo` returns state from the token module.

td202:
## Purpose

Closes COR-1371
Depends on https://github.com/Concordium/concordium-base/pull/634

Add module-defined state to `GetTokenInfo` query.

## Changes

- `queryTokenModuleState` added to Token Module.
- Add module `Concordium.Scheduler.ProtocolLevelTokens.Queries`:
  - Add `QueryT` monad transformer that implements `PLTKernelQuery` and `PLTKernelFail` based on a non-updatable view of a block state.
  - Add `queryTokenInfo`, refactoring most of the functionality from `getTokenInfo`, and invoking `queryTokenModuleState`.
- Implement `getAccountState` for `KernelT`.
- Update `getTokenInfo` to call `queryTokenInfo`.
- Update `getTokenInfoV2` for error handling.
- Tests.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
@@ -161,8 +161,9 @@ getTokenInfoV2 cptr blockType blockHashPtr tokenIdPtr tokenIdLen outHash outVec 
                 Q.BQRBlock _ (Left QTIEUnknownToken) -> do
                     copyHashTo outHash res
                     return $ queryResultCode QRNotFound
-                Q.BQRBlock _ (Left e) -> do
-                    mvLog mvr Logger.External Logger.LLError $ "Internal error processing GetTokenInfo: " ++ show e
+                Q.BQRBlock _ (Left e@QTIEInternal) -> do
+                    mvLog mvr Logger.External Logger.LLError $
+                        "Internal error processing GetTokenInfo: " ++ show e
                     return $ queryResultCode QRInternalError
                 Q.BQRBlock _ (Right r) ->
                     returnMessageWithBlock (copier outVec) outHash (res $> r)
--- a/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
@@ -16,6 +16,7 @@ import Control.Concurrent
 import qualified Data.ByteString.Unsafe as BS
 import qualified Data.FixedByteString as FBS
 import Data.Foldable (toList)
+import Data.Functor
 import Data.Int
 import qualified Data.ProtoLens as Proto
 import qualified Data.ProtoLens.Combinators as Proto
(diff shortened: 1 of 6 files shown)


concordium-consensus/src/Concordium/Scheduler/EnvironmentImplementation.hs
```diff
@@ -507,9 +507,13 @@ instance (BS.BlockStateOperations m, PVSupportsPLT (MPV m)) => PLTKernelQuery (K
...
                 Just acct -> BS.getAccountTokenBalance acct tokenIx
-    getAccountState _acct _key = do
-        -- TODO: implement
-        return Nothing
+    getAccountState acctIndex key = do
+        tokenIx <- ask
+        bs <- use ssBlockState
+        lift $
+            BS.bsoGetAccountByIndex bs acctIndex >>= \case
+                Nothing -> error "getAccountBalance: Account does not exist"
```
drsk0, line 515: The error should be "getAccountState: ... " ?

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
```diff
@@ -155,7 +157,17 @@ getTokenInfoV2 cptr blockType blockHashPtr tokenIdPtr tokenIdLen outHash outVec
         Left _ -> return $ queryResultCode QRInvalidArgument
         Right tokenId -> do
             res <- runMVR (Q.getTokenInfo bhi tokenId) mvr
-            returnMaybeMessageWithBlock (copier outVec) outHash res
+            case res of
+                Q.BQRBlock _ (Left QTIEUnknownToken) -> do
+                    copyHashTo outHash res
+                    return $ queryResultCode QRNotFound
+                Q.BQRBlock _ (Left e) -> do
```
drsk0, line 164: Maybe this shouldn't be a catch all case and instead explicitly use `QTIEInternal` in case we expand the error cases in the future.

drsk0 (approved):
This looks good once we settled on the right CBOR representation in base.

concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/Queries.hs
```diff
@@ -0,0 +1,128 @@
...
+    { -- | Index of the token. This must be valid in the context of the block state.
+      qcTokenIndex :: !TokenIndex,
+      -- | The block state.
+      qcBlockState :: !(BlockState m)
+    }
+
+-- | @QueryT fail ret@ is a monad transformer that supports 'PLTKernelQuery' and 'PLTKernelFail'
+--  on top of an underlying monad that implements 'BS.BlockStateQuery'.
+newtype QueryT fail ret m a = QueryT
+    { runQueryT' :: ReaderT (QueryContext m) (ContT (Either fail ret) m) a
```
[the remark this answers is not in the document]
td202: Bind (`>>=`) is more expensive with `ExceptT` because each one requires pattern matching on the `Either` from the first action. With `ContT` you don't construct the `Either` at all, because you either just pass the result to the continuation or call the failure continuation instead.

PR #1391, opened 2025-05-16
Test for TokenHolderOperation and allow any transaction in Scheduler tests

eb-concordium:
## Purpose

- Add some test cases for the `TokenHolder`, namely:
  - Test that a token holder transaction fails with reject reason `NonExistentTokenId` when the token does not exist
  - Test that token holder transaction does not fail with reject reason `NonExistentTokenId` when the token does actually exist
- Add support for testing any kind of transaction (i.e. account transaction, account creation or chain update) in the scheduler tests.

## Changes

- Tests testing the above have been added
- The `TransactionAndAssertion` type have been generalized with a new type `BlockItemAndAssertion` to have any BlockItem (account transaction, account creation or chain update) together with an assertion. Accordingly a `runSchedulerTestAssertIntermediateStatesBlockItem` have been intruced so that one can combine different kinds of transactions (with assertions) in a test, e.g. in something like

```haskell
...
do
        let transactionsAndAssertions :: [Helpers.BlockItemAndAssertion pv]
            transactionsAndAssertions =
                [ Helpers.BlockItemAndAssertion
                    { biaaTransaction =
                        Runner.AccountTx ... -- a normal account transaction
                      ataaAssertion = \result _ -> do
                        return $ Helpers.assertSuccess result
                    },
                  Helpers.BlockItemAndAssertion
                    { biaaTransaction =
                        Runner.ChainUpdateTx ... -- a chain update transaction
                      biaaAssertion = \result _ -> do
                        return $ Helpers.assertSuccess result
                    }
                ]
        Helpers.runSchedulerTestAssertIntermediateStatesBlockItem
            @pv
            Helpers.defaultTestConfig
            initialBlockState
            transactionsAndAssertions
```

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Scheduler/Runner.hs
+++ b/concordium-consensus/src/Concordium/Scheduler/Runner.hs
@@ -41,9 +41,9 @@ signTxSingle key TransactionHeader{..} encPayload = Types.signTransactionSingle 
   where
     header = Types.TransactionHeader{thPayloadSize = Types.payloadSize encPayload, ..}
 
-signChainTx :: ChainTransaction -> Types.UpdateInstruction
+signChainTx :: ChainUpdateTransaction -> Types.UpdateInstruction
 signChainTx
-    ChainTransaction
+    ChainUpdateTransaction
         { ctSeqNumber = ruiSeqNumber,
           ctEffectiveTime = ruiEffectiveTime,
           ctTimeout = ruiTimeout,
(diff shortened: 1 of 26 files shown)


concordium-consensus/src/Concordium/Scheduler/Runner.hs
```diff
@@ -282,3 +321,19 @@ data TransactionJSON = TJSON
...
+data ChainTransaction = ChainTransaction
+    { ctSeqNumber :: Updates.UpdateSequenceNumber,
+      ctEffectiveTime :: Types.TransactionTime,
+      ctTimeout :: Types.TransactionTime,
+      ctPayload :: Updates.UpdatePayload,
+      ctKeys :: [(Updates.UpdateKeyIndex, KeyPair)]
+    }
+    deriving (Show, Generic)
+
+data AnyTransaction = AccountTx TransactionJSON | ChainTx ChainTransaction
```
limemloh, line 334: **Suggestion (non-blocking):** A small thing here, but maybe use `BlockItem` instead of `Any` to be more consistent with what we currently are naming similar types, and `ChainTx` should maybe be `UpdateTx` or `ChainUpdateTx` :shrug: 

```suggestion
data BlockItemDescription = AccountTx TransactionJSON | UpdateTx ChainTransaction
```

limemloh (approved):

td202 (approved):
Looks good. Perhaps a little bit more documentation in Runner.hs could be useful.

PR #1408, opened 2025-06-25
Local rust build

allanbrondum:
## Purpose

Update README.md for building and running the node. Also fix haskell library runtime binding for running the node.

## Changes

Fixed haskell library runtime binding for running the node to look in the GHC variant subfolder also, where we also look at compile time.

Also changed binding to happen agains the `.stack-work` in root and not `.stack-work` inside `concordium-consensus`. This is because Haskell language server will look in and build to root `.stack-work`, hence it mean less confusion and likely better build performance (and the build files for the language server and those build for the node cannot get out of sync).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased changes
 
+## 9.0.5 (DevNet)
+
+- Protocol-level tokens:
+  - Simplified token transaction handling: Consolidated all token holder and
+    governance operations for a single `TokenUpdateTransaction` type.
+  - Moved authorization from the scheduler to the token module.
+  - Fix a bug in checking authorization for CreatePLT update.
+
 ## 9.0.4 (DevNet)
 
 - Protocol-level tokens:
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/ProtocolLevelTokens.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/ProtocolLevelTokens.hs
@@ -46,9 +46,7 @@ data PLTConfiguration = PLTConfiguration
       -- | The token module reference.
       _pltModule :: !TokenModuleRef,
       -- | The number of decimal places used in the representation of the token.
-      _pltDecimals :: !Word8,
-      -- | The index of the account authorized to perform token governance operations.
-      _pltGovernanceAccountIndex :: !AccountIndex
+      _pltDecimals :: !Word8
     }
     deriving (Eq, Ord, Show)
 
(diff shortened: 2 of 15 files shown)


concordium-node/build.rs
```diff
@@ -67,7 +67,7 @@ fn main() -> std::io::Result<()> {
                 // otherwise auto-discover the directories via stack
                 let stack_install_root_command = command_output(Command::new("stack").args([
                     "--stack-yaml",
-                    "../concordium-consensus/stack.yaml",
+                    "../stack.yaml",
```
allanbrondum, line 72: look for build output in root like Haskell language server

concordium-node/build.rs
```diff
@@ -110,10 +110,11 @@ fn main() -> std::io::Result<()> {
                 };
 
                 println!(
-                    "cargo:rustc-env={}={}:{}",
+                    "cargo:rustc-env={}={}:{}:{}",
                     lib_path,
                     ghc_lib_dir.as_path().to_string_lossy(),
-                    stack_install_lib.as_path().to_string_lossy()
+                    stack_install_lib.as_path().to_string_lossy(),
+                    local_package.as_path().to_string_lossy()
```
allanbrondum: Some libraries are in this subfolder, allow linked at runtime to them also like we do at compile time

hie.yaml
```diff
@@ -0,0 +1,70 @@
+cradle:
```
drsk0: I wonder if this is really needed. Because hls is working for me sometimes/sometimes not.

allanbrondum: Using haskell 9.6.7 where the language server runs much more stable, it does not seems to be needed. I'll remove it again together with the haskell upgrade. I have some experiment with it here: https://github.com/Concordium/concordium-node/pull/1409. Seems to work fine, but need to get it planned into next cycle to finish it.

allanbrondum: On 9.6.6, I'm not sure, the language server seems to unstable that it is hard to determine, at least it does not hurt from the testing I've done.

limemloh: Did you use a prebuild HLS? or do did you compile it for the GHC version?

allanbrondum: I installed it with ghcup: hls 2.9.0.1. The binary I run is then `haskell-language-server-9.6.6~2.9.0.1`. For 9.6.7 it is a newer package: hls 2.10.0.0, and binary `haskell-language-server-9.6.7~2.10.0.0`

allanbrondum: @limemloh In this in relation to whether `hie.yaml` is needed or whether it hangs?

allanbrondum: You can try to run it like described here: https://concordium.atlassian.net/wiki/spaces/cor/pages/1897988099/Haskell+development+environment. Just run the language server with no arguments in the project folder. Sometimes it hangs, some times not. I have a very simple test project where it even hangs if there are compile errors.

concordium-node/build.rs
limemloh: Not sure I fully understand the implication of adding this, but it would be good to add some context in a comment, so we know why this needs to be there in the future.

hie.yaml
allanbrondum: Removed `hie.yaml`, does not seem needed. About Haskell language server, see https://concordium.atlassian.net/wiki/spaces/cor/pages/1897988099/Haskell+development+environment#Installation

limemloh (approved):

PR #1410, opened 2025-06-26
Use rust trie state for token state and remove the account state.

limemloh:
## Purpose

Use rust-implemented state trie for token modules.

Related PR in base: TBA

## Changes

- Change the token module, so it no longer have a separate account state, instead it uses the token state and prefixes the account related state with magic value + account index.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -280,15 +280,6 @@ class (BlockStateTypes m, Monad m) => AccountOperations m where
         TokenIndex ->
         m TokenRawAmount
 
-    -- | Look up the 'TokenStateValue' associated with a particular token and key on an account.
-    -- This is only available at protocol versions that support protocol-level tokens.
-    getAccountTokenState ::
-        (PVSupportsPLT (MPV m)) =>
-        Account m ->
-        TokenIndex ->
-        TokenStateKey ->
-        m (Maybe TokenStateValue)
-
 -- * Active, current and next bakers/delegators
 
 --
@@ -516,16 +507,13 @@ class (MonadProtocolVersion m, Monad m) => PLTQuery bs m where
     -- | Get the 'TokenIndex' associated with a 'TokenId' (if it exists).
     getTokenIndex :: (PVSupportsPLT (MPV m)) => bs -> TokenId -> m (Maybe TokenIndex)
 
-    -- | Get the state of a token for a given 'TokenStateKey'. Returns @Nothing@ if the token
-    --  does not have a state for the given key.
+    -- | Convert a persistent state to a mutable one that can be updated by the scheduler.
     --
     --  PRECONDITION: The token identified by 'TokenIndex' MUST exist.
-    getTokenState ::
-        (PVSupportsPLT (MPV m)) =>
-        bs ->
-        TokenIndex ->
-        TokenStateKey ->
-        m (Maybe TokenStateValue)
+    thawTokenState :: (PVSupportsPLT (MPV m)) => bs -> TokenIndex -> m StateV1.MutableState
+
+    -- | Read key from the mutable token state.
+    getTokenState :: (PVSupportsPLT (MPV m)) => bs -> TokenStateKey -> StateV1.MutableState -> m (Maybe TokenStateValue)
 
     -- | Get the configuration of a protocol-level token.
     --
@@ -1547,6 +1535,14 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
     -- Unlike the other chain updates this is a separate function, since there is no queue associated with PLTs.
     bsoIncrementPLTUpdateSequenceNumber :: (PVSupportsPLT (MPV m)) => UpdatableBlockState m -> m (UpdatableBlockState m)
 
+    -- | Convert a mutable state to a persistent one that can be stored in the block state.
+    --
+    --  PRECONDITION: The token identified by 'TokenIndex' MUST exist.
+    bsoFreezeTokenState :: (PVSupportsPLT (MPV m)) => UpdatableBlockState m -> TokenIndex -> StateV1.MutableState -> m (UpdatableBlockState m)
+
+    -- | Write value to key in the mutable token state.
+    bsoSetTokenState :: (PVSupportsPLT (MPV m)) => TokenStateKey -> Maybe TokenStateValue -> StateV1.MutableState -> m (Maybe Bool)
+
     -- | Overwrite the election difficulty, removing any queued election difficulty updates.
     --  This is intended to be used for protocol updates that affect the election difficulty in
     --  tandem with the slot duration.
@@ -1632,24 +1628,6 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
     --  addresses.
     bsoSuspendValidators :: (PVSupportsValidatorSuspension (MPV m)) => UpdatableBlockState m -> [AccountIndex] -> m ([(AccountIndex, AccountAddress)], UpdatableBlockState m)
 
-    -- | Set the token-level state of a token for a given 'TokenStateKey'. If the value is
-    --  @Nothing@, the key is removed from the token state. Otherwise the key is mapped to the
-    --  specified value.
-    --
-    --  PRECONDITION: The token identified by 'TokenIndex' MUST exist.
-    bsoSetTokenState ::
-        (PVSupportsPLT (MPV m)) =>
-        -- | The current block state.
-        UpdatableBlockState m ->
-        -- | The token index to update.
-        TokenIndex ->
-        -- | The key to set or remove.
-        TokenStateKey ->
-        -- | The value to set.
-        Maybe TokenStateValue ->
-        -- | The updated block state.
-        m (UpdatableBlockState m)
-
     -- | Set the recorded total circulating supply for a protocol-level token.
     --  This should always be kept up-to-date with the total balance held in accounts
     --  (and smart contracts).
@@ -1681,20 +1659,6 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
         PLTConfiguration ->
         -- | The index of the new token and the updated block state.
         m (TokenIndex, UpdatableBlockState m)
-
-    -- | Update the token module state.
-    --
-    --  PRECONDITION: The token identified by 'TokenIndex' MUST exist.
-    bsoUpdateTokenAccountModuleState ::
-        (PVSupportsPLT (MPV m)) =>
-        UpdatableBlockState m ->
-        -- | The token index to update.
-        TokenIndex ->
-        -- | The account to update.
-        AccountIndex ->
-        -- | The list of updates to the module state.
-        [(TokenStateKey, GSAccount.TokenAccountStateValueDelta)] ->
-        m (UpdatableBlockState m)
 
     -- Update the token balance. Returns 'Nothing' if the update would overflow or underflow
     -- the token balance on the account.
@@ -1846,7 +1810,8 @@ instance (Monad (t m), MonadTrans t, ModuleQuery m) => ModuleQuery (MGSTrans t m
 instance (Monad (t m), MonadTrans t, PLTQuery bs m) => PLTQuery bs (MGSTrans t m) where
     getPLTList = lift . getPLTList
     getTokenIndex bs = lift . getTokenIndex bs
-    getTokenState bs ti = lift . getTokenState bs ti
+    thawTokenState blockState = lift . thawTokenState blockState
+    getTokenState bs key = lift . getTokenState bs key
     getTokenConfiguration bs = lift . getTokenConfiguration bs
     getTokenCirculatingSupply bs = lift . getTokenCirculatingSupply bs
 
@@ -1963,7 +1928,6 @@ instance (Monad (t m), MonadTrans t, AccountOperations m) => AccountOperations (
     getAccountCooldowns = lift . getAccountCooldowns
     getAccountTokens = lift . getAccountTokens
     getAccountTokenBalance acct tokenIx = lift $ getAccountTokenBalance acct tokenIx
-    getAccountTokenState acct tokenIx key = lift $ getAccountTokenState acct tokenIx key
     {-# INLINE getAccountCanonicalAddress #-}
     {-# INLINE getAccountAmount #-}
     {-# INLINE getAccountAvailableAmount #-}
@@ -1981,7 +1945,6 @@ instance (Monad (t m), MonadTrans t, AccountOperations m) => AccountOperations (
     {-# INLINE getAccountCooldowns #-}
     {-# INLINE getAccountTokens #-}
     {-# INLINE getAccountTokenBalance #-}
-    {-# INLINE getAccountTokenState #-}
 
 instance (Monad (t m), MonadTrans t, ContractStateOperations m) => ContractStateOperations (MGSTrans t m) where
     thawContractState = lift . thawContractState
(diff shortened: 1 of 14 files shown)


concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1547,6 +1535,14 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
     -- Unlike the other chain updates this is a separate function, since there is no queue associated with PLTs.
     bsoIncrementPLTUpdateSequenceNumber :: (PVSupportsPLT (MPV m)) => UpdatableBlockState m -> m (UpdatableBlockState m)
 
+    -- | Convert a mutable state to a persistent one that can be stored in the block state.
+    --
+    --  PRECONDITION: The token identified by 'TokenIndex' MUST exist.
+    bsoFreezeTokenState :: (PVSupportsPLT (MPV m)) => UpdatableBlockState m -> TokenIndex -> StateV1.MutableState -> m (UpdatableBlockState m)
```
td202: Can the MutableState still be used afterwards? I guess so. I guess it's important to note that future changes won't be reflected in the block state (unless this is called again).

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1547,6 +1535,14 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
...
+
+    -- | Write value to key in the mutable token state.
+    bsoSetTokenState :: (PVSupportsPLT (MPV m)) => TokenStateKey -> Maybe TokenStateValue -> StateV1.MutableState -> m (Maybe Bool)
```
td202: I think it should be made explicit that changes are not propagated to the `UpdatableBlockState` unless `bsoFreezeTokenState` is called.

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
```diff
@@ -70,6 +74,130 @@ newMutableState msContext ptr = do
 withMutableState :: MutableState -> (Ptr MutableStateInner -> IO a) -> IO a
 withMutableState MutableState{msInner = MutableStateInner fp} = withForeignPtr fp
 
+-- | Look up entry using key and read the value in a mutable state.
+foreign import ccall "lookup_entry_value_mutable_state"
+    lookup_entry_value_mutable_state ::
```
td202: I think it's generally best to rename into camelCase.
```suggestion
    lookupEntryValueMutableState ::
```

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
```diff
@@ -70,6 +74,130 @@ newMutableState msContext ptr = do
...
+            then return Nothing
+            else do
+                len <- peek outPtr
+                Just
+                    <$> BSU.unsafePackCStringFinalizer
+                        (castPtr response)
+                        (fromIntegral len)
+                        (rs_free_array_len response (fromIntegral len))
+
+-- | Insert value into entry of the provided key.
```
td202:
```suggestion
-- | Insert entry into the mutable state, overwriting the value if already
--  present.
```

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
```diff
@@ -70,6 +74,130 @@ newMutableState msContext ptr = do
...
+        CSize ->
+        -- | Location of the value.
+        Ptr Word8 ->
+        -- | Length of the value.
+        CSize ->
+        -- | Reference to the mutable state.
+        Ptr MutableStateInner ->
+        IO Word8
+
+-- | Insert value into entry at key.
```
td202:
```suggestion
-- | Insert a value into the mutable state at a specified key.
```

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
```diff
@@ -70,6 +74,130 @@ newMutableState msContext ptr = do
...
+--
+-- Returns
+-- * @Just True@ if an entry had a value which got replaced.
+-- * @Just False@ if no entry was overwritten.
+-- * @Nothing@ signals error due to the entry being locked.
+insertMutableState :: BS.ByteString -> BS.ByteString -> MutableState -> IO (Maybe Bool)
```
td202: Document the arguments - particularly which is key and which is value.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -516,16 +507,13 @@ class (MonadProtocolVersion m, Monad m) => PLTQuery bs m where
...
-    getTokenState ::
-        (PVSupportsPLT (MPV m)) =>
-        bs ->
-        TokenIndex ->
-        TokenStateKey ->
-        m (Maybe TokenStateValue)
+    thawTokenState :: (PVSupportsPLT (MPV m)) => bs -> TokenIndex -> m StateV1.MutableState
+
+    -- | Read key from the mutable token state.
+    getTokenState :: (PVSupportsPLT (MPV m)) => bs -> TokenStateKey -> StateV1.MutableState -> m (Maybe TokenStateValue)
```
td202: I would suggest refactoring the state in a couple of ways:

- Abstract the `MutableState` type. 
- Rename `thawTokenState` to `getMutableTokenState`.
- Rename `bsoFreezeTokenState` to something like `bsoSetTokenState` maybe?
- `getTokenState` and `bsoSetTokenState` -> put into a separate interface, and rename to `lookupTokenState` and `updateTokenState` (or similar). There's not really any reason that `getTokenState` should depend on the block state.

My assumption is that "thawing" the state gets a mutable copy of the current state in the block state, and "freezing" updates the block state to match the given mutable state.

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
```diff
@@ -70,6 +74,130 @@ newMutableState msContext ptr = do
...
+                    (castPtr keyPtr)
+                    (fromIntegral keyLen)
+                    (castPtr valuePtr)
+                    (fromIntegral valueLen)
+                    inner
+            return $ case out of
+                0 -> Just False
+                1 -> Just True
+                2 -> Nothing
+                _ -> error "Unexpected FFI status code"
```
td202:
```suggestion
                _ -> error "insertMutableState: Unexpected FFI status code"
```

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
```diff
@@ -70,6 +74,130 @@ newMutableState msContext ptr = do
...
+                delete_entry_mutable_state
+                    (msContext mutableState)
+                    (castPtr keyPtr)
+                    (fromIntegral keyLen)
+                    inner
+            return $ case out of
+                0 -> Just False
+                1 -> Just True
+                2 -> Nothing
+                _ -> error "Unexpected FFI status code"
```
td202:
```suggestion
                _ -> error "deleteEntryMutableState: Unexpected FFI status code"
```

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/ProtocolLevelTokens.hs
```diff
@@ -63,26 +63,21 @@ emptyTokenAccountStateTable = TokenAccountStateTable{tokenAccountStateTable = Ma
...
+        { tasBalance = TokenRawAmount 0
         }
 
 -- | Token state at the account level
 data TokenAccountState = TokenAccountState
     { -- | The available balance for the account.
-      tasBalance :: !TokenRawAmount,
-      -- | The token module state for the account, represented as a key-value map.
-      tasModuleState :: !(Map.Map TokenStateKey TokenStateValue)
+      tasBalance :: !TokenRawAmount
```
td202: Convert to `newtype`.

concordium-consensus/src/Concordium/GlobalState/Persistent/Account/ProtocolLevelTokens.hs
```diff
@@ -95,9 +90,7 @@ instance (MonadBlobStore m) => BlobStorable m TokenAccountState
 -- | An update to the token account state.
 data TokenAccountStateDelta = TokenAccountStateDelta
     { -- | A change to the token balance.
-      tasBalanceDelta :: !(Maybe TokenAmountDelta),
-      -- | A change to the token module state.
-      tasModuleStateDelta :: ![(TokenStateKey, TokenAccountStateValueDelta)]
+      tasBalanceDelta :: !(Maybe TokenAmountDelta)
```
td202: At this point, I think this type is useless - delete.

concordium-consensus/src/Concordium/Scheduler/EnvironmentImplementation.hs
```diff
@@ -475,43 +479,53 @@ instance
...
                           _pltecConfiguration = config,
-                          _pltecEnergy = Nothing
+                          _pltecEnergy = Nothing,
+                          _pltecMutableState = mutState
                         }
             runKernelT op context initialExecutionState
+        when (isRight res) $ do
+            let outBlockState = finalExecutionState ^. plteBlockState
+            newBlockState <- lift $ BS.bsoFreezeTokenState outBlockState tokenIx mutState
+            ssBlockState .= newBlockState
```
td202: I think it could be useful to track whether the token state was updated at all in order to avoid this update in the (relatively common) case where the state is not updated. i.e. have a `_plteStateIsDirty :: !Bool` in the `PLTExecutionState` that is initially `False`, but set to `True` in `setTokenState`. It's not necessarily that freezing the token state is expensive (I presume it's not in this case) but that parts of the block state may end up being duplicated in the blob store needlessly.

concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
```diff
@@ -62,40 +63,45 @@ import qualified SchedulerTests.Helpers as Helpers
...
+    GetAccountIndex :: AccountIndex -> PLTKernelQueryCall AccountIndex
+    GetAccountCanonicalAddress :: AccountIndex -> PLTKernelQueryCall AccountAddress
+    GetGovernanceAccount :: PLTKernelQueryCall AccountIndex
+    GetCirculatingSupply :: PLTKernelQueryCall TokenRawAmount
+    GetDecimals :: PLTKernelQueryCall Word8
+
+deriving instance Show (PLTKernelQueryCall ret)
+deriving instance Eq (PLTKernelQueryCall ret)
+
+getAcccountStateCall :: AccountIndex -> TokenStateKey -> PLTKernelQueryCall (Maybe TokenStateValue)
```
td202:
```suggestion
getAccountStateCall :: AccountIndex -> TokenStateKey -> PLTKernelQueryCall (Maybe TokenStateValue)
```

td202: Add doc.

td202: Since the test is depending on the behaviour of `accountStateKey`, please ensure there's also unit tests for `accountStateKey`.

concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
```diff
@@ -62,40 +63,45 @@ import qualified SchedulerTests.Helpers as Helpers
...
-    GetAccountBalance :: acct -> PLTKernelQueryCall acct TokenRawAmount
-    GetAccountState :: acct -> TokenStateKey -> PLTKernelQueryCall acct (Maybe TokenStateValue)
-    GetAccountCanonicalAddress :: acct -> PLTKernelQueryCall acct AccountAddress
-    GetGovernanceAccount :: PLTKernelQueryCall acct acct
-    GetCirculatingSupply :: PLTKernelQueryCall acct TokenRawAmount
-    GetDecimals :: PLTKernelQueryCall acct Word8
-
-deriving instance (Show acct) => Show (PLTKernelQueryCall acct ret)
-deriving instance (Eq acct) => Eq (PLTKernelQueryCall acct ret)
+data PLTKernelQueryCall ret where
```
td202: I see no good reason for unnecessarily reifying `acct` here, and you haven't updated the documentation. Please revert. (OK, I see that the `PLTKernelQuery` instance more or less needs `acct ~ AccountIndex`, but still I would keep the definitions as abstract as possible.)

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1547,6 +1535,14 @@ class (BlockStateQuery m, PLTQuery (UpdatableBlockState m) m) => BlockStateOpera
     -- Unlike the other chain updates this is a separate function, since there is no queue associated with PLTs.
     bsoIncrementPLTUpdateSequenceNumber :: (PVSupportsPLT (MPV m)) => UpdatableBlockState m -> m (UpdatableBlockState m)
 
+    -- | Convert a mutable state to a persistent one that can be stored in the block state.
+    --
+    --  PRECONDITION: The token identified by 'TokenIndex' MUST exist.
+    bsoFreezeTokenState :: (PVSupportsPLT (MPV m)) => UpdatableBlockState m -> TokenIndex -> StateV1.MutableState -> m (UpdatableBlockState m)
+
+    -- | Write value to key in the mutable token state.
```
td202: Document the return value.

concordium-consensus/src/Concordium/GlobalState/ContractStateV1.hs
limemloh: Some of the names clash with the non-ffi version, so I'll suffix these with FFI then

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
limemloh: I'll align with this.

> - Abstract the MutableState type.

Not sure what you are suggesting here, introduce a newtype wrapper for mutable state with functions? or make it a type parameter and provide some type class for the functions?

concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
limemloh: Reverted

concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
limemloh: Sure, the complexity of the test will be equal to the implementation though... but I guess this "locks" the behavior

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
td202: The latter. I would have it as a parameter to `PLTQuery`, and add it as an associated type to `BlockStateTypes`. Then have a type class for the functions (`getTokenState` and `bsoSetTokenState`).

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
limemloh: Good question... it seems that we are actually deallocating it on the rust side after the freeze, so using it again would be undefined behavior, I think.

limemloh: Looking further into this, we are not deallocating here.
It can actually be reused for further changes after a freeze.

concordium-consensus/src/Concordium/Scheduler/EnvironmentImplementation.hs
limemloh: Makes sense

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
limemloh: For better future-proofing we should probably not reuse it, I have added a comment.

PR #1417, opened 2025-07-01
benchtxns

allanbrondum:
## Purpose

Create benchmarks of PLT transactions against CCD transfer

## Changes

Implemented benchmarks. Notice that in each benchmark run, the number of operations/transaction we measure on is scaled up by the factor `transactionCount` in order to be able to ignore overhead from running the scheduler etc

## Checklist

      hard-to-understand areas.


concordium-consensus/benchmarks/transactions/SchedulerTests/Helpers.hs
allanbrondum, line 1: I copied this file from tests. I would like to hear what the best approach would be here. To cut the contents in this module down to what is actually used in the benchmark. Or to rewrite it in a way that may be more suitable for benchmarking instead of running tests and doing assertions (if so, any guidance for doing this the best way is appreciated, especially, what level in the scheduler should we hook in at)

td202: I would cut it down to what's needed.

td202 (commented):
Could you please add some more documentation to be clear about what each benchmark is actually measuring. I think there are quite a few subtleties and caveats in interpreting the results, potentially.

PR #1420, opened 2025-07-03
Plt token pause scheduler implementation

drsk0:
This adds the 'pause' logic to `executeTokenUpdateTransaction` in the scheduler.

### DIFF
--- a/concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/Module.hs
+++ b/concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/Module.hs
@@ -359,7 +359,7 @@ executeTokenUpdateTransaction TransactionContext{..} tokenParam = do
             failTH
                 OperationNotPermitted
                     { trrOperationIndex = opIndex,
-                      trrAddressNotPermitted = Just (accountTokenHolder tcSenderAddress),
+                      trrAddressNotPermitted = Nothing,
                       trrReason = Just $ Text.pack $ "token operation " ++ op ++ " is paused"
                     }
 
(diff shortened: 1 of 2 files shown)


concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
```diff
@@ -847,6 +864,29 @@ testExecuteTokenUpdateTransactionMintBurn = describe "executeTokenUpdateTransact
                                 }
                          )
         assertTrace (executeTokenUpdateTransaction (sender 1) (encodeTransaction transaction)) trace
+    it "paused operation" $ do
+        let transaction =
+                TokenUpdateTransaction . Seq.fromList $
+                    [TokenMint (TokenAmount 10_000_000 6)]
```
soerenbf, line 886: It might make sense to add tests covering the other operations as well, also to verify that only the intended operations are blocked during the token pause.

soerenbf (approved):
I think it would make sense to have either @td202 or @limemloh review these changse as well 😄

concordium-consensus/tests/scheduler/SchedulerTests/TokenModule.hs
drsk0: I added tests for `burn` and `transfer`.

concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/Module.hs
```diff
@@ -350,6 +353,15 @@ executeTokenUpdateTransaction TransactionContext{..} tokenParam = do
     tokenParamLBS =
         BS.Builder.toLazyByteString $ BS.Builder.shortByteString $ parameterBytes tokenParam
     failTH = pltError . encodeTokenRejectReason
+    checkPaused opIndex op = do
+        paused <- isJust <$> getTokenState "paused"
+        when paused $
+            failTH
+                OperationNotPermitted
+                    { trrOperationIndex = opIndex,
+                      trrAddressNotPermitted = Just (accountTokenHolder tcSenderAddress),
```
td202: I think the address should be `None` here, because the global pause is not a permission pertaining to a particular account.

PR #1421, opened 2025-07-03
Require TokenModuleRef to match defined value

td202:
## Purpose

Closes [COR-1553](https://linear.app/concordium/issue/COR-1553/require-tokenmoduleref-to-match-defined-value)

This adds a check that the `TokenModuleRef` used when creating a PLT is a specific value, namely the SHA256 hash of the String `TokenModuleV0`.

## Changes

- Introduce `tokenModuleV0Ref` constant to reference the current token module implementation.
- `handleCreatePLT` tests for the appropriate reference and fails if it does not match `tokenModuleV0Ref`.
- Tests are updated to use the correct ref and test for failure with a different ref.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
+++ b/concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
@@ -17,12 +17,12 @@ import qualified Concordium.GlobalState.DummyData as DummyData
 import qualified Concordium.GlobalState.Persistent.Account as BS
 import qualified Concordium.GlobalState.Persistent.BlobStore as Blob
 import qualified Concordium.GlobalState.Persistent.BlockState as BS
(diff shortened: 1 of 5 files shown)


concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
```diff
@@ -149,7 +149,7 @@ testTokenHolder _ pvString =
               tipBurnable = True
             }
     tp = Types.TokenParameter $ BSS.toShort $ CBOR.tokenInitializationParametersToBytes params
-    createPLT = Types.CreatePLT gtu (TokenModuleRef dummyHash) 0 tp
```
DOBEN, line 152: ```dummyHash = Hash.hashShort BSS.empty``` can be removed as well.

DOBEN (approved):

PR #1423, opened 2025-07-07
feat: add Windows code signing to the release pipeline

nb-ccd:
## Purpose

This branch starts work to add code signing for Windows builds into the Github Actions release pipeline

* Adds DigiCert signing manager controller tool to the Windows runner
* Decodes base64 certificate and writes it to filesystem
* Configures the smctl tool with relevant credentials

Here is a signed Windows build produced by these steps: https://github.com/Concordium/concordium-node/actions/runs/16150140527


.github/workflows/release.yaml
```diff
@@ -314,6 +345,89 @@ jobs:
...
+      
+      - name: Rename files to prepare for signing 
+        run: |
+          mv ./Node/ConcordiumConsensusDLL ./Node/ConcordiumConsensusDLL.dll
+          mv ./Node/ConcordiumBaseDLL ./Node/ConcordiumBaseDLL.dll
+          mv ./Node/ConcordiumSmartContractEngineDLL ./Node/ConcordiumSmartContractEngineDLL.dll
+          mv ./Node/Sha2DLL ./Node/Sha2DLL.dll
+          mv ./Node/NodeRunnerService ./Node/NodeRunnerService.exe
+          mv ./Node/NodeCollector ./Node/NodeCollector.exe 
+          mv ./Node/ConcordiumNode ./Node/ConcordiumNode.exe
```
Thahara: Maybe add a comment saying this is necessary because `smctl` only supports signing files with certain extensions, see [here](https://docs.digicert.com/it/digicert-keylocker/client-tools/signing-tools/files-supported-for-signing.html)

.github/workflows/release.yaml
```diff
@@ -254,17 +255,47 @@ jobs:
...
+          SM_CLIENT_CERT_FILE: ${{ steps.windows_certificate.outputs.CERTIFICATE_PATH }}
+          SM_CLIENT_CERT_PASSWORD: ${{ secrets.WINDOWS_SM_CLIENT_CERT_PASSWORD }}
+        run: |
+          smctl healthcheck --all
+        shell: cmd
+
+      - name: Install dependencies
+        run: |
+          choco install yq jq -y
+        shell: bash
```
td202: Surely this needs to go before `Extrapolate artifact name`, which is where `jq` is actually used?

.github/workflows/release.yaml
```diff
@@ -314,6 +345,90 @@ jobs:
...
+      - name: Extract files to prepare for signing 
+        run: |
+          dir service\windows\installer
+          "C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x86\MsiDb.exe" -d service\windows\installer/Node.msi -x Node.cab
+          mkdir Node
+          dir
+          expand -d Node.cab
+          expand -F:* Node.cab ./Node
+          dir Node
+        shell: cmd
```
td202, line 358: I'm sure this will work OK, but I really think it would be better to sign the binaries before we actually build the installer. For instance, modify the `service/windows/installer/build.ps1` script to sign the binaries, build the installer, then sign the installer. That way, we don't need to unpack the files from the MSI, rename them, repack them, etc.

td202 (approved):

.github/workflows/release.yaml
nb-ccd: I do think this would be preferable, but I was not familiar enough with these build tools or Powershell to know how to do that in a way that's optional and not disruptive to external users or folks running the build locally who don't have the signing configuration set up.

It seemed as if the MSBuild SDK can include configuration for signing builds, but I don't believe we're currently using that? https://learn.microsoft.com/en-us/visualstudio/msbuild/signfile-task?view=vs-2022

nb-ccd: If you wanna talk me through how this script works happy to take a stab at it, but if you're planning to update build tooling for windows fairly soon maybe the pragmatic thing to do is

1. use this janky method SRE had come up with beforehand for now
2. update to a newer Wix version
3. incorporate code signing as part of that new version (in which I think candle/light are deprecated)

.github/workflows/release.yaml
nb-ccd: I see, that seems true. Updated.

.github/workflows/release.yaml
nb-ccd: Updated

.github/workflows/release.yaml
td202: I think we should merge this for now and we can come back later and improve it.

td202: I'm going to merge so that we get this in for building the next release. I've opened an issue for addressing this: https://linear.app/concordium/issue/COR-1627/streamline-windows-node-signing-process

nb-ccd: Thanks @td202 and apologies for not responding at the time, I don't work on Fridays.

PR #1430, opened 2025-07-17
Protocol update

eb-concordium:
## Purpose
Closes https://linear.app/concordium/issue/COR-634/implement-p9-protocol-update-data-migration.

IMPORTANT: Currently a dummy hash for the update is used. When the specification document is ready and approved, we should update the hash (to the hash of the document, i.e. the specification hash).

## Changes

- P8 -> P9 update added

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/ProtocolUpdate/P8/ProtocolP9.hs
+++ b/concordium-consensus/src/Concordium/ProtocolUpdate/P8/ProtocolP9.hs
@@ -78,7 +78,7 @@ import Concordium.Types.ProtocolVersion
 -- | The hash that identifies a update from P8 to P9 protocol.
 --  This is the hash of the published specification document.
 updateHash :: SHA256.Hash
-updateHash = read "000000000000000000000000000000000000000000000000000000000000009" -- FIXME: replace with hash of the published specification document
+updateHash = read "0000000000000000000000000000000000000000000000000000000000000009" -- FIXME: replace with hash of the published specification document
 
 -- | Construct the genesis data for a P8.ProtocolP9 update.
 --  This takes the terminal block of the old chain which is used as the basis for constructing


concordium-consensus/src/Concordium/ProtocolUpdate/P8/ProtocolP9.hs
```diff
@@ -0,0 +1,120 @@
...
+import qualified Concordium.KonsensusV1.TreeState.Implementation as TreeState
+import Concordium.KonsensusV1.TreeState.Types
+import Concordium.KonsensusV1.Types
+import Concordium.Types.HashableTo (getHash)
+import Concordium.Types.ProtocolVersion
+
+-- | The hash that identifies a update from P8 to P9 protocol.
+--  This is the hash of the published specification document.
+updateHash :: SHA256.Hash
+updateHash = read "dummy" -- FIXME: replace with hash of the published specification document
```
td202: This is probably going to cause a runtime error because this is not a valid hash. Fill in an actual dummy value. e.g.
```suggestion
updateHash = read "000000000000000000000000000000000000000000000000000000000000009" -- FIXME: replace with hash of the published specification document
```

concordium-consensus/src/Concordium/ProtocolUpdate/P8/ProtocolP9.hs
```diff
@@ -0,0 +1,120 @@
...
+--        is a 'GDP8Regenesis'.
+--
+--  * 'genesisPreviousGenesis' is the hash of the previous genesis block.
+--
+--  * 'genesisTerminalBlock' is the hash of the last finalized block of the previous chain.
+--
+--  * 'genesisStateHash' is the state hash of the last finalized block of the previous chain.
+--
+--  * 'genesisMigration' is derived from the data provided to the protocol update, which is the
+--     validator score parameters
```
td202: Nope!

td202:
Also, update the changelog.

concordium-consensus/src/Concordium/ProtocolUpdate/P8/ProtocolP9.hs
eb-concordium: Good catch. Fixed.

PR #1432, opened 2025-07-18
Fix the CI pipeline so as not to rebuild everything unnecessarily.

td202:
## Purpose

When building dependencies and the consensus, set the target as `concordium-consensus` so that running the tests won't trigger a rebuild. This also avoids building the tests/benchmarks for `concordium-base`.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -224,12 +224,12 @@ jobs:
     - name: Build haskell dependencies (unless they were cached)
       if: steps.stack-programs.outputs.cache-hit != 'true' || steps.stack-global.outputs.cache-hit != 'true'
       run: |
-        stack build --test --bench --only-dependencies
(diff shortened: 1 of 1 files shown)


eb-concordium (approved):

.github/workflows/build-test.yaml
```diff
@@ -173,11 +180,13 @@ jobs:
           ~/.cargo/git
           concordium-base/rust-src/target
           concordium-base/lib
-          concordium-consensus/smart-contracts/wasm-chain-integration/target
-          concordium-consensus/smart-contracts/lib
```
drsk0, line 177: was this moved recently?

drsk0 (approved):
Thanks!

.github/workflows/build-test.yaml
td202: No, I think this moved quite a long time ago.

PR #1437, opened 2025-07-22
Parallelbuild

allanbrondum:
## Purpose

Declare to build modules in parallel. Will have effect in development environment and in CI

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -246,7 +246,7 @@ jobs:
     # Compile Haskell sources. This must be done before running checks or tests on the Rust sources.
     - name: Build consensus and run tests
       run: |
-        stack build concordium-consensus --test --bench --force-dirty --no-run-benchmarks --ghc-options "-Werror" --ghc-options -j
+        stack build concordium-consensus --test --bench --force-dirty --no-run-benchmarks --ghc-options "-Werror"
 
     # RUST #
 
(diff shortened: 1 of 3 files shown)


td202:
I just don't think `-j` is appropriate in all circumstances, so I don't think we should configure it like this. For instance, it may be better not to use `-j` when building dependencies, since `stack` already parallelizes at the package level.

allanbrondum:
> I just don't think `-j` is appropriate in all circumstances, so I don't think we should configure it like this. For instance, it may be better not to use `-j` when building dependencies, since `stack` already parallelizes at the package level.

I tried timing building dependencies for the node (deleting `~/.stack-work/snapshots` before each build):
```
stack build --only-dependencies  696.41s user 128.75s system 428% cpu 3:12.57 total (plt branch)
stack build --only-dependencies --ghc-options -j  699.76s user 126.84s system 430% cpu 3:12.23 total (plt branch)
stack build --only-dependencies --ghc-options -j4  698.49s user 126.90s system 427% cpu 3:13.09 total (plt branch)
stack build --only-dependencies  811.14s user 142.65s system 583% cpu 2:43.57 total (plt branch, j4 ghc options in ~/.stack/config.yaml)
stack build --only-dependencies  865.71s user 237.69s system 625% cpu 2:56.54 total (this branch)
stack build --only-dependencies  869.05s user 236.49s system 616% cpu 2:59.44 total (this branch, j4 ghc options in ~/.stack/config.yaml)
```

PR #1443, opened 2025-07-23
Charge energy immediatly in token module (COR-1653)

DOBEN:
## Purpose

Energy should be charged as early as possible in the token module. To avoid that some failure cases are executed "for free" because the energy is not yet charged for the given operation.

## Changes

- Change to charge energy as early as possible in the token module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Changelog
 
 ## Unreleased changes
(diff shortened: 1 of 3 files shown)


squirer (approved):
looks sensible to me

td202 (approved):
Looks good!

PR #1445, opened 2025-07-24
Token id should be normalized (COR-1670)

DOBEN:


concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/Queries.hs
```diff
@@ -129,7 +129,7 @@ queryTokenInfo tokenId bs = case sSupportsPLT (accountVersion @(AccountVersionFo
                                       tsTotalSupply = toTokenAmount _pltDecimals totalSupply,
                                       tsModuleState = tms
                                     }
-                        return $ Right TokenInfo{tiTokenId = tokenId, tiTokenState = ts}
+                        return $ Right TokenInfo{tiTokenId = toUpperCase tokenId, tiTokenState = ts}
```
td202:
```suggestion
                        return $ Right TokenInfo{tiTokenId = _pltTokenId, tiTokenState = ts}
```
The canonical form is what's specified in `PLTConfiguration`. Not converted to uppercase.

td202 (approved):

PR #1453, opened 2025-08-18
COR-1627: streamline windows node signing process

drsk0:
- Moves the signing of the libraries and executables into the build script
- Avoids repacking of cabinet files


nb-ccd (approved):
Looks good to me, thanks for removing the very messy cabinet file manipulation!

PR #1459, opened 2025-09-09
fix: provide code coverage reporting on PRs (except for Windows)

nb-ccd:
## Purpose

QA want to introduce code coverage reporting for Rust projects - this implements it for the non-Windows parts of the node.

It will:

* Provide a comment on pull requests showing which modified lines are coverage by tests
* Provide a detailed report in the CodeCov SaaS platform (log in through Github account) of coverage overall 

## Changes

* Adds `-Cinstrument-coverage` rust flag
* Uses cargo-llvm-cov component to generate LCOV file from the instrumentation
* Uploads coverage report to codecov (free for open source projects) 
* Adds codecov.yml configuring the repo for informational only reporting.

cargo-llvm-cov seemed well maintained and well integrated with Cargo ecosystem compared to other reporting tools. Repo here provides a good overview: https://github.com/taiki-e/cargo-llvm-cov


.github/workflows/build-test.yaml
```diff
@@ -260,14 +273,28 @@ jobs:
...
         cargo clippy --locked --target x86_64-pc-windows-gnu --manifest-path service/windows/installer/custom-actions/Cargo.toml --all -- -D warnings
     - name: Test Rust crates (without extra features)
       run: |
-        cargo test --manifest-path concordium-node/Cargo.toml --all
+        cargo llvm-cov --manifest-path concordium-node/Cargo.toml --all --lcov --output-path lcov-default.info
     - name: Test Rust crates (with feature 'network_dump')
       run: |
-        cargo test --manifest-path concordium-node/Cargo.toml --all --features=network_dump
+        cargo llvm-cov --manifest-path concordium-node/Cargo.toml --all --features=network_dump  --lcov --output-path lcov-network-dump.info
+    - name: Upload coverage to Codecov
```
allanbrondum: I think we should merge the coverages for different feature enablements as described here: https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#merge-coverages-generated-under-different-test-conditions. I don't think a coverage report for each set of features makes so much sense?

.github/workflows/build-test.yaml
```diff
@@ -168,8 +171,12 @@ jobs:
         profile: minimal
         toolchain: ${{ matrix.plan.rust }}
         override: true
-        components: clippy
+        components: clippy, llvm-tools
         target: x86_64-pc-windows-gnu
+    # Install tool for code coverage reporting.
+    - name: Install cargo-llvm-cov
+      run: |
+        cargo +stable install cargo-llvm-cov --locked
```
allanbrondum: Detail: you should not need to specify `+stable` toolchain directly, just run `cargo` directly, it will then use the toolchain installed by actions-rs/toolchain

.github/workflows/build-test.yaml
```diff
@@ -130,6 +130,9 @@ jobs:
     # Use fixed OS version because we install packages on the system.
     runs-on: ubuntu-22.04
     if: ${{ !github.event.pull_request.draft }}
+    env:
+      RUSTFLAGS: "-Cinstrument-coverage"
```
allanbrondum: When using `cargo llvm-cov`, I don't think you need to set `RUSTFLAGS`, it will handle it automatically (https://crates.io/crates/cargo-llvm-cov). Tried it locally also without setting it.

.github/workflows/build-test.yaml
```diff
@@ -130,6 +130,9 @@ jobs:
...
+      LLVM_PROFILE_FILE: "concordium-node-%p-%m.profraw"
```
allanbrondum: Do we need this setting or is the "default" ok?

.github/workflows/build-test.yaml
allanbrondum: That is, don't generate a report (`--no-report`) when running tests, and then generate the report in "Upload coverage to Codecov" or in another separate step

.github/workflows/build-test.yaml
```diff
@@ -260,14 +273,28 @@ jobs:
...
       run: |
         cargo clippy --locked --target x86_64-pc-windows-gnu --manifest-path service/windows/installer/custom-actions/Cargo.toml --all -- -D warnings
     - name: Test Rust crates (without extra features)
       run: |
-        cargo test --manifest-path concordium-node/Cargo.toml --all
+        cargo llvm-cov --manifest-path concordium-node/Cargo.toml --all --lcov --output-path lcov-default.info
     - name: Test Rust crates (with feature 'network_dump')
       run: |
-        cargo test --manifest-path concordium-node/Cargo.toml --all --features=network_dump
+        cargo llvm-cov --manifest-path concordium-node/Cargo.toml --all --features=network_dump  --lcov --output-path lcov-network-dump.info
```
allanbrondum: Suggestion: Use `cargo llvm-cov test` instead of `cargo llvm-cov`. I think it makes it more explicit/descriptive for those not familiar with the tool, that we are actually running tests here.

.github/workflows/build-test.yaml
nb-ccd: All right - as I understand it the CodeCov platform does aggregate any reports within a repo itself (https://docs.codecov.com/docs/merging-reports), but doing that locally doesn't seem like a bad idea to me in case we wanted to do any analysis of this kind locally or move to another provider.

.github/workflows/build-test.yaml
nb-ccd: I guess the default works fine here, I named it explicitly to understand how the tool worked but no humans need to ever look at these / verify them so it's not needed now it's working.

.github/workflows/build-test.yaml
```diff
@@ -242,9 +247,15 @@ jobs:
       run: |
         cargo check --locked --manifest-path collector/Cargo.toml
     - name: Check that the Windows node runner service compiles (gnu toolchain, disable embedded resources)
+      # Setting instrument coverage flag for Windows targets results in error because profiler builtins aren't available.
+      env:
+        RUSTFLAGS: ""
```
allanbrondum: These sections should also no longer be needed now?

nb-ccd: Good point! Thanks.

nb-ccd:
> It looks like the CI run time is going from ~20 min to ~2hr 20 min - I don't think this is acceptable. It looks like the big change is in building and running the Haskell tests (now taking over 2 hours), which seems problematic, since that's not even testing any of the code in the coverage report.

Definitely agree that we don't want to rebuild the Haskell code unnecessarily rather than load it from cache. I'm not sure how this PR could be affecting that behaviour unless having the environment variable `LLVM_PROFILE_FILE` set for the overall job is enabling Haskell code coverage, it only touches Rust steps. I have removed that env var to see if that affects the runtime. 

> It looks like the CI run time is going from ~20 min to ~2hr 20 min - I don't think this is acceptable. It looks like the big change is in building and running the Haskell tests (now taking over 2 hours), which seems problematic, since that's not even testing any of the code in the coverage report.

@td202 heh, I hadn't realised this was a specific change in this PR rather than this project just being very slow to build still -  testing this morning it seems like it might be because setting `LLVM_PROFILE_FILE` globally triggers concordium-consensus being rebuilt or otherwise invalidates the cached version, since Haskell also supports that property. I think Allan is right that that property doesn't add anything in a CI context anyway so I've removed it.

.github/workflows/build-test.yaml
nb-ccd: This seems true in CI as well, good call.

PR #1575, opened 2026-03-06
POC: Use concrete types for block state in scheduler

limemloh:
## Purpose

This is restructuring the plt-scheduler codebase to use concrete types for block state in scheduler, removing traits such as the `BlockStateOperations` and `BlockStateQuery`.

The PR includes placeholder implementation for P10 and P11 scheduler/blockstate and changes the integration tests to abstract the protocol version, such that these can be run across multiple versions.

This PR is not in a ready-state, if this were to be merged, we still need to:
- Adjust documentation in a number of places.
- Adjust Haskell implementation (include the protocol version in FFI).
- ~~Change the remaining tests in plt_query.~~ Done
- ~~Merge in latest changes~~ Done


plt/plt-block-state/src/block_state/p10.rs
```diff
@@ -0,0 +1,392 @@
...
+
+    fn cache(&mut self, _loader: &mut impl blob_store::BackingStoreLoad) {
+        todo!()
+    }
+}
+
+/// Simplistic implementation of the block state that serializes as a flat sequence of bytes.
+// todo do real implementation as part of https://linear.app/concordium/issue/PSR-11/port-the-plt-block-state-to-rust
+#[derive(Debug, Default, Clone, Serialize)]
+pub struct Tokens {
```
limemloh, line 57: Maybe we should pull this into a separate module since it is shared across blockstate versions

plt/plt-block-state/src/block_state/p11.rs
```diff
@@ -0,0 +1,55 @@
+use super::{blob_store, p10, BlockStateOperations};
+
+/// Block state providing the various block state operations.
+#[derive(Debug, Clone)]
+pub struct PltBlockStateP11 {
+    // todo implement real block state as part of https://linear.app/concordium/issue/PSR-11/port-the-plt-block-state-to-rust
+    /// Simplistic state that is used as a temporary implementation of the block state
+    pub tokens: p10::Tokens,
```
limemloh, line 8: Just want to highlight that we are reusing the token state type from P10 here

plt/plt-block-state/src/ffi/block_state.rs
```diff
@@ -50,13 +96,40 @@ extern "C" fn ffi_free_plt_block_state(block_state: *mut PltBlockStateSavepoint)
...
+        | ProtocolVersion::P3
+        | ProtocolVersion::P4
+        | ProtocolVersion::P5
+        | ProtocolVersion::P6
+        | ProtocolVersion::P7
+        | ProtocolVersion::P8
+        | ProtocolVersion::P9 => {
+            unimplemented!()
+        }
+        ProtocolVersion::P10 => {
```
limemloh, line 119: Most of these matches on the protocol version could probably be removed using dynamic dispatch, and we can take that discussion if you want, but to me these are trivial enough to remain as static dispatch.

plt/plt-scheduler/tests/plt_create.rs
```diff
@@ -22,13 +23,22 @@ mod block_state_external_stubbed;
 /// Test create protocol-level token.
 #[test]
 fn test_plt_create() {
-    let mut stub = BlockStateWithExternalStateStubbed::new();
-    assert_eq!(stub.plt_update_instruction_sequence_number(), 0);
+    test_plt_create_worker::<p10::PltBlockStateP10>()
+}
```
limemloh, line 25: Here we trigger the test for P10 block state, we can add more tests like this for the protocol versions we find interesting to test

plt/plt-block-state/tests/block_state.rs
limemloh, line 1: These have been moved to be unit tests next to implementation

plt/plt-block-state/tests/block_state_no_external.rs
limemloh, line 1: Since we are not trying to combine haskell blockstate and rust block state into one big block state abstraction, there is probably no need for this stub

limemloh:
PR is not maintained

PR #1576, opened 2026-03-10
Allow token module to branch on RBAC support

limemloh:
## Purpose

Expose whether RBAC should be enabled to the token module.
[Linear task](https://linear.app/concordium/issue/PSR-71/support-running-multiple-protocol-version-in-the-rust-scheduler).

No unit tests were added as part of this, since this will be covered as part of the unit tests for the RBAC implementation.

## Changes

- Pass in the protocol version to the rust scheduler lib.
- Parameterize unit tests on the protocol version.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-scheduler/src/ffi/queries.rs
+++ b/plt/plt-scheduler/src/ffi/queries.rs
@@ -4,7 +4,7 @@
 
 use crate::queries;
 use crate::queries::QueryTokenInfoError;
-use concordium_base::base::AccountIndex;
+use concordium_base::base::{AccountIndex, ProtocolVersion};
 use concordium_base::common;
 use libc::size_t;
 use plt_block_state::block_state::{ExecutionTimePltBlockState, PltBlockStateSavepoint};
@@ -52,6 +52,7 @@ extern "C" fn ffi_query_plt_list(
     get_account_address_by_index_callback: GetCanonicalAddressByAccountIndexCallback,
     get_token_account_states_callback: GetTokenAccountStatesCallback,
     block_state: *const PltBlockStateSavepoint,
+    protocol_version: u64,
     return_data_out: *mut *mut u8,
     return_data_len_out: *mut size_t,
 ) -> u8 {
@@ -72,8 +73,11 @@ extern "C" fn ffi_query_plt_list(
         get_token_account_states_ptr: get_token_account_states_callback,
     };
 
+    let protocol_version =
+        ProtocolVersion::try_from(protocol_version).expect("Unknown protocol version");
     let internal_block_state = unsafe { &*block_state };
     let block_state = ExecutionTimePltBlockState {
+        protocol_version,
         internal_block_state,
         backing_store_load: load_callback,
         external_block_state: external_callbacks,
@@ -134,6 +138,7 @@ extern "C" fn ffi_query_token_info(
     get_account_address_by_index_callback: GetCanonicalAddressByAccountIndexCallback,
     get_token_account_states_callback: GetTokenAccountStatesCallback,
     block_state: *const PltBlockStateSavepoint,
+    protocol_version: u64,
     token_id: *const u8,
     token_id_len: size_t,
     return_data_out: *mut *mut u8,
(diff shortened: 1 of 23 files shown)


plt/plt-token-module/tests/token_module_list_updates.rs
```diff
@@ -94,7 +97,10 @@ fn test_allow_list_updates() {
 
 #[test]
 fn test_deny_list_updates() {
-    let mut stub = KernelStub::with_decimals(0);
+    test_deny_list_updates_worker(ProtocolVersion::P10);
+}
+fn test_deny_list_updates_worker(protocol_version: ProtocolVersion) {
+    let mut stub = KernelStub::with_decimals(0, protocol_version);
```
allanbrondum: I think there are two things here:
* I don't think we need two functions, in all the places where we only run the test on one protocol version. I don't see the need for that?
* How do we specify a "default" protocol version for tests: Wouldn't it be nicer to put it into the initialization methods for `KernelStub`. Alternatively create an options builder `KernelOptions` that has a default protocol version, and where you can change the default with `KernelOptions::protocol_version`. If we explicitly write `ProtocolVersion::P10` in the tests, to me that is a signal that it is important this runs on P10. So I would think it is better to not specify protocol version explicitly, but rely on a default in cases where we don't care which protocol version the tests runs on.

plt/plt-scheduler/src/ffi/queries.rs
```diff
@@ -134,6 +138,7 @@ extern "C" fn ffi_query_token_info(
     get_account_address_by_index_callback: GetCanonicalAddressByAccountIndexCallback,
     get_token_account_states_callback: GetTokenAccountStatesCallback,
     block_state: *const PltBlockStateSavepoint,
+    protocol_version: u64,
```
allanbrondum, line 141: To me it seems more natural, that the Rust maintained block state is initialized with a protocol version and knows that version inherently from initialization until being dropped. Is there a benefit in having the Haskell side specify the protocol version in each usage of the state?

limemloh: Haskell side already tracks it and this was the simplest place to add it.
We could put it the rust block state, but I don't think it is written to disk, so we would need the information when loading it from disk, which requires separate implementations of `blobstore::Loadable` per protocol version.

plt/plt-token-module/tests/token_module_list_updates.rs
limemloh: 1. yes, since we agreed that we will only run for the latest relevant protocol version I have made it a const per testsuite.
2. I think it is good for a testsuite to contain the all relevant information and avoid a global default, instead there is now a `LATEST_PROTOCOL_VERSION` const, which we need to keep up to date. I think a builder makes sense when there are obvious default values, which probably is the case for decimals, but not so much for the protocol version if you ask me.
    So I'll leave that for a later iteration.

plt/plt-scheduler/src/ffi/queries.rs
allanbrondum: The protocol version could also be in a field in a type wrapping what is loaded? That is how it is on the Haskell side also (essentially, just via the type system) and how the Rust implementation would also be, if all state was in Rust.

td202 (approved):

plt/plt-token-module/tests/token_module_update_general.rs
```diff
@@ -16,11 +16,13 @@ mod kernel_stub;
...
 
 const NON_EXISTING_ACCOUNT: AccountAddress = AccountAddress([2u8; 32]);
+/// Default protocol version used across the tests.
+const PROTOCOL_VERSION: ProtocolVersion = utils::LATEST_PROTOCOL_VERSION;
 
 /// Test failure to decode token operations
 #[test]
 fn test_update_token_decode_failure() {
-    let mut stub = KernelStub::with_decimals(0);
+    let mut stub = KernelStub::with_decimals(0, PROTOCOL_VERSION);
```
allanbrondum: Small detail, but why not? Then the test is more self contained (also visually/reading wise)
```suggestion
    let mut stub = KernelStub::with_decimals(0, LATEST_PROTOCOL_VERSION);
```

allanbrondum (approved):

plt/plt-token-module/tests/token_module_update_general.rs
limemloh: Just to factor out the common thing, but I don't have a good reason so I will change it then

plt/plt-scheduler/src/ffi/queries.rs
limemloh: We can iterate on this as we implement the block state.

PR #1587, opened 2026-03-13
Support RBAC operations in token module

limemloh:
## Purpose

Support assign+revoke+updateMetadata in token module

Query support is not included.
Tests specifically for assign+revoke is not implemented, since it needs the query to be implemented.

[Ref RBC-14](https://linear.app/concordium/issue/RBC-14/concordium-nodeplt-extend-token-module-with-rbac-functionality)

## Changes

- Update base and adjust according to recent changes.

## Checklist

      hard-to-understand areas.


plt/plt-token-module/src/token_module.rs
```diff
@@ -16,12 +16,6 @@ pub use initialize::*;
 pub use queries::*;
 pub use update::*;
 
-/// An invariant in the token module state that should be enforced
-/// is broken.
-#[derive(Debug, thiserror::Error)]
-#[error("Token module state invariant broken: {0}")]
-pub struct TokenModuleStateInvariantError(pub String);
```
limemloh, line 23: This seems to have been duplicated, so I removed it in favor of `TokenStateInvariantError`

plt/plt-token-module/src/token_module/initialize.rs
```diff
@@ -54,6 +53,14 @@ fn initialize_token_impl(
     kernel: &mut impl TokenKernelOperations,
     init_params: TokenModuleInitializationParameters,
 ) -> Result<(), TokenInitializationError> {
+    if !init_params.additional.is_empty() {
```
limemloh: This change from a CBOR decoding error to be something we have to check, due to the `additional` field being added to the parameter.

plt/plt-token-module/src/token_module/initialize.rs
```diff
@@ -70,8 +77,8 @@ fn initialize_token_impl(
         )
     })?;
     kernel.set_module_state(STATE_KEY_NAME, Some(name.into()));
-    let encoded_metadata = cbor::cbor_encode(&metadata);
-    kernel.set_module_state(STATE_KEY_METADATA, Some(encoded_metadata));
+    key_value_state::set_metadata_url(kernel, &metadata);
```
limemloh, line 74: Ideally we would ensure the `additional` field on metadata url here as well, but it would be a breaking change for P9 and P10, so we need to check no one is using it on mainnet and testnet

plt/plt-token-module/src/token_module/initialize.rs
```diff
@@ -91,6 +98,9 @@ fn initialize_token_impl(
         STATE_KEY_GOVERNANCE_ACCOUNT,
         Some(common::to_bytes(&governance_account_index.index)),
     );
+    if kernel.support_rbac() {
+        key_value_state::assign_account_roles(kernel, governance_account_index, roles::ALL)?;
```
td202: Only supported roles should be assigned, right?

plt/plt-token-module/src/token_module/update.rs
```diff
@@ -313,6 +317,18 @@ fn execute_token_update_operation<
...
+            execute_assign_admin_roles(transaction_execution, kernel, operation)
+        }
+        TokenOperation::RevokeAdminRoles(operation) if kernel.support_rbac() => {
+            execute_revoke_admin_roles(transaction_execution, kernel, operation)
+        }
+        TokenOperation::UpdateMetadata(operation) if kernel.support_updating_metadata() => {
+            execute_update_metadata(transaction_execution, kernel, operation)
+        }
+        _ => Err(TokenUpdateErrorInternal::UnsupportedOperation {
+            reason: "Operation not supported by the protocol",
```
td202: Maybe:
```suggestion
            reason: "Operation not supported at the current protocol version",
```

plt/plt-token-module/src/token_module/update.rs
```diff
@@ -609,3 +658,119 @@ fn execute_remove_deny_list<
...
+    operation: &TokenUpdateAdminRolesDetails,
+) -> Result<(), TokenUpdateErrorInternal> {
+    check_authorized(
+        transaction_execution,
+        kernel,
+        TokenAdminRole::UpdateAdminRoles,
+    )?;
+    check_roles_supported(kernel, &operation.roles)?;
+    let account = kernel.account_by_address(&operation.account.address)?;
+    kernel.touch_account(&account);
```
td202: Why do we touch the account? I think this would only be necessary if we expect the token account state to show the account's roles - but that's not the case, AFAIK. (The `GetTokenAuthorizations` is the way to determine roles for an account.) However, if we do want to touch accounts then we should probably also touch the governance account when the token is initialised.

plt/plt-token-module/src/token_module/initialize.rs
allanbrondum: I removed these exact lines on p11 branch. I think you base branch is out of sync with p11. The field `additional` does not exist anymore.

plt/plt-token-module/tests/token_module_burn.rs
allanbrondum, line 1: Shouldn't there be some tests running on the previous protocol version asserting the old behaviour?

plt/plt-token-module/src/token_module/update.rs
```diff
@@ -609,3 +658,119 @@ fn execute_remove_deny_list<
...
+    let account_index = kernel.account_index(&account);
+    if account_index == kernel.account_index(&transaction_execution.sender_account())
+        && operation.roles.contains(&TokenAdminRole::UpdateAdminRoles)
+    {
+        return Err(TokenUpdateErrorInternal::OperationNotPermitted {
+            account_address: Some(operation.account.address),
+            reason: "Sender not allowed to remove own admin role",
+        });
+    }
+    kernel.touch_account(&account);
```
td202: Touching is perhaps even more questionable here.

plt/plt-token-module/src/token_module/initialize.rs
allanbrondum: I think this is also unrelated to this PR in any case?

plt/plt-token-module/src/key_value_state.rs
```diff
@@ -102,45 +105,47 @@ pub fn is_burnable(kernel: &impl TokenKernelQueries) -> bool {
...
-) -> Result<String, TokenModuleStateInvariantError> {
+pub fn get_name(kernel: &impl TokenKernelQueries) -> Result<String, TokenStateInvariantError> {
     kernel
         .get_module_state(STATE_KEY_NAME)
-        .ok_or_else(|| TokenModuleStateInvariantError("Name not present".to_string()))
+        .ok_or_else(|| TokenStateInvariantError("Name not present".to_string()))
         .and_then(|value| {
             String::from_utf8(value).map_err(|err| {
-                TokenModuleStateInvariantError(format!("Stored name is invalid UTF8: {}", err))
+                TokenStateInvariantError(format!("Stored name is invalid UTF8: {}", err))
```
td202:
```suggestion
                TokenStateInvariantError(format!("Stored name is invalid UTF-8: {}", err))
```

plt/plt-token-module/tests/token_module_burn.rs
```diff
@@ -214,3 +215,92 @@ fn test_not_burnable() {
...
+
+    let reject_reason = utils::assert_reject_reason(&res);
+    assert_matches!(
+        reject_reason,
+        TokenModuleRejectReason::OperationNotPermitted(OperationNotPermittedRejectReason {
+            index: 0,
+            address: Some(address),
+            reason: Some(reason)
+        }) => {
+            assert_eq!(reason, "sender is not authorized to perform the operation for this token".to_string());
```
td202: Is `to_string` necessary?

plt/plt-token-module/src/key_value_state.rs
```diff
@@ -150,6 +155,50 @@ pub fn get_governance_account_index(
...
+pub fn assign_account_roles(
+    kernel: &mut impl TokenKernelOperations,
+    account: AccountIndex,
+    assigned_roles: &[TokenAdminRole],
+) -> Result<(), TokenStateInvariantError> {
+    let mut roles = get_account_roles(kernel, account)?;
+    for role in assigned_roles {
+        roles.assign(*role)
+    }
+    kernel.set_account_state(account, STATE_KEY_ACCOUNT_ROLES, roles.into_state_value());
```
td202: I don't think that account roles are stored in the account state (see [here](https://concordium.atlassian.net/wiki/spaces/ARC/pages/2425749512/Design+RBAC+in+the+node#RBAC-state-in-the-token-module)). The reason being that we need to be able to iterate over all of the accounts that have admin roles.

td202 (approved):

plt/plt-scheduler/tests/plt_create.rs
allanbrondum, line 1: It think we may want one at least one or two tests at scheduler level. E.g. a test creating a token and asserting roles have been assigned. And a token update test that also checks authorizations are checked when performing an operation that needs permissions.

plt/plt-token-module/src/token_module/update.rs
```diff
@@ -313,6 +317,18 @@ fn execute_token_update_operation<
...
+        TokenOperation::AssignAdminRoles(operation) if kernel.support_rbac() => {
+            execute_assign_admin_roles(transaction_execution, kernel, operation)
+        }
+        TokenOperation::RevokeAdminRoles(operation) if kernel.support_rbac() => {
+            execute_revoke_admin_roles(transaction_execution, kernel, operation)
+        }
+        TokenOperation::UpdateMetadata(operation) if kernel.support_updating_metadata() => {
+            execute_update_metadata(transaction_execution, kernel, operation)
+        }
+        _ => Err(TokenUpdateErrorInternal::UnsupportedOperation {
```
allanbrondum: Not sure I like that we have a fallback here. Since it means you don't get a compile error, if you introduce a new operation.

allanbrondum: Maybe e.g. putting `if kernel.support_rbac()` check into `execute_assign_admin_roles` will work better.

plt/plt-token-module/src/token_module/update.rs
```diff
@@ -609,3 +658,119 @@ fn execute_remove_deny_list<
...
+    transaction_execution: &mut TE,
+    kernel: &mut TK,
+    operation: &TokenUpdateAdminRolesDetails,
+) -> Result<(), TokenUpdateErrorInternal> {
+    check_authorized(
+        transaction_execution,
+        kernel,
+        TokenAdminRole::UpdateAdminRoles,
+    )?;
+    check_roles_supported(kernel, &operation.roles)?;
```
allanbrondum, line 750: Maybe allow revoking roles for features that are not enabled?

plt/plt-token-module/src/token_module/initialize.rs
limemloh: I'll make a [task](https://linear.app/concordium/issue/COR-2338/disallow-additional-data-during-plt-initialization) for the decision on this

plt/plt-token-module/tests/token_module_burn.rs
```diff
@@ -214,3 +215,92 @@ fn test_not_burnable() {
             if reason == "feature not enabled" && operation_type == "burn"
     );
 }
+
+/// Reject when governance account is not holding the burn role.
+#[test]
+fn test_reject_without_role() {
```
allanbrondum: Just a small thing: Maybe we can call the tests `test_authorization_xxx` or some other naming pattern to allow identifying them from the others.

plt/plt-token-module/tests/token_module_update_metadata.rs
```diff
@@ -0,0 +1,141 @@
...
+    let operations = vec![TokenOperation::UpdateMetadata(new_metadata_url.clone())];
+    token_module::execute_token_update_transaction(
+        &mut execution,
+        &mut stub,
+        RawCbor::from(cbor::cbor_encode(&operations)),
+    )
+    .expect("executes successfully");
+
+    let new_encoded_metadata = cbor::cbor_encode(&new_metadata_url);
+    assert_eq!(
```
allanbrondum: I think it will be better to make assertions based on querying the token info and reading the metadata that way. Because that behaviour is what we really want to have an assertion on, and that way don't force a specific storage via this test.

plt/plt-token-module/tests/token_module_pause.rs
```diff
@@ -351,3 +353,83 @@ fn test_unpause_multiple_ops() {
...
+    let mut execution = TransactionExecutionTestImpl::with_sender(account2);
+    let operations = vec![TokenOperation::Pause(TokenPauseDetails {})];
+    token_module::execute_token_update_transaction(
+        &mut execution,
+        &mut stub,
+        RawCbor::from(cbor::cbor_encode(&operations)),
+    )
+    .expect("execute");
+    assert_eq!(
+        stub.lookup_token_state_value(b"\0\0paused".into()),
```
allanbrondum: Also here, would it not be better to query token info and assert on that?

plt/plt-token-module/src/token_module/update.rs
limemloh: I agree that touch does not make sense here, I'll remove

plt/plt-token-module/src/token_module/update.rs
limemloh: Yes, I considered that as well.
I mainly did this to be consistent with the behavior of assign... I do see a slight benefit in the rejection, since it will hint something is off, so I would prefer to keep it.

plt/plt-token-module/src/key_value_state.rs
limemloh: thanks 😅 , I was realizing this as well as I started on the implementation of the query... I'll adjust it

plt/plt-token-module/tests/token_module_burn.rs
limemloh: right, it could make sense to check the new operations are rejected in the old protocol version.
Otherwise it is pretty much just the auth error message which changed...
Did I miss something?

plt/plt-token-module/tests/token_module_update_metadata.rs
limemloh: I agree, but on that note it makes more sense if these tests were all on the scheduler-level.
The benefit of having them here would be to access the internals.

allanbrondum: True, but it's like having two good things, we would not let one block the other. At least we can do one of them (assert on queries). That will make the test easier to move the scheduler level later, which I think I agree makes sense. We can also write the new test a scheduler level already now if we want?

limemloh: I'll keep these here, use the query and make a [task for moving these to the scheduler level](https://linear.app/concordium/issue/COR-2339/move-token-module-tests-to-be-scheduler-level).

plt/plt-scheduler/tests/plt_create.rs
limemloh: I made the [task to move every test to scheduler level](https://linear.app/concordium/issue/COR-2339/move-token-module-tests-to-be-scheduler-level), I think this task would cover your comment as well, so I'll not do this as part of this PR

plt/plt-token-module/tests/token_module_burn.rs
allanbrondum: But even if is it just the auth error message that is different, it is still two separate code paths for authroization enforcement on P9/P10 and P11?

allanbrondum (approved):

plt/plt-token-module/tests/token_module_burn.rs
limemloh: Right, I'll make sure to hit the code path in a test

PR #1588, opened 2026-03-13
Ar/psr 12 implement left full merkle binary tree in rust

allanbrondum:
## Purpose

Implementation of Rust block state. The state builds on types representing immutable values, with the top level type being `block_state::BlockState`. The only type representing mutability is `block_state::MutableBlockState`.

## Changes

* implement block state queries and operations for `block_state::BlockState`. The only type representing mutability is `block_state::MutableBlockState`
* implement PLT state in `block_state::state::protocol_level_tokens::ProtocolLevelTokens`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlobStore.hs
@@ -54,6 +54,7 @@ module Concordium.GlobalState.Persistent.BlobStore (
     -- * In-memory blob store
     MemBlobStore (..),
     newMemBlobStore,
+    newMemBlobStoreWithBytes,
     destroyMemBlobStore,
     MemBlobStoreT (..),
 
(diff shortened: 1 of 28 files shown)


plt/plt-block-state/src/block_state/blob_reference/hashed_buffered_reference.rs
```diff
@@ -0,0 +1,245 @@
...
+
+                reference
+            }
+            HashedBufferedRefRepr::Cache { reference, .. } => *reference,
+        }
+    }
+}
+
+#[derive(Debug)]
+enum HashedBufferedRefRepr<V> {
```
limemloh: I was hoping we could avoid this abstraction, and track the caching separately, it would get rid of the interior mutability and allow for better control of the caching.
Maybe we have to do this, for the sake of time, but I hope there would be a better approach.

plt/plt-block-state/src/block_state/blob_reference/hashed_buffered_reference.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// The value is in the backing storage.
+    Store { reference: BlobReference },
+    /// The value is in memory and not written to backing storage.
+    Memory { value: V },
+    /// The value is in the backing storage, and also cached in memory.
+    Cache { reference: BlobReference, value: V },
```
limemloh: Maybe we should box `value` here because of the union, otherwise `HashedBufferedRefRepr::Store` will take the same amount of memory as having the value cached.
Alternatively we box it in the type parameter.

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,245 @@
...
+        Self {
+            inner: self.inner.clone(),
+        }
+    }
+}
+
+#[derive(Debug)]
+struct HashedBufferedRefImpl<V> {
+    /// Lazily calculated hash.
+    hash: Option<Hash>,
```
limemloh, line 126: I think ideally these hashes would be stored separately in a cache as well.
Having it here results in `HashedBufferedRefImpl` taking a lot more memory.

PR #1589, opened 2026-03-13
fix: copy plt rust libraries and strip out tokens

squirer:
## Purpose

Rust libraries handling for plt target

## Changes

copy rust plt libraries as part of build static libraries and perform removal of rust specific symbols for library linking

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-node/build.rs
+++ b/concordium-node/build.rs
@@ -884,7 +884,7 @@ fn link_static_libs() -> std::io::Result<()> {
                             );
                         }
                     }
-                    _ => panic!("Unable to link ghc libs"),
+                    _ => panic!("Unable tod link ghc libs"),
                 }
             }
         });
@@ -894,6 +894,8 @@ fn link_static_libs() -> std::io::Result<()> {
         out_dir
     );
     println!("cargo:rustc-link-lib=static=Rcrypto");
+    println!("cargo:rustc-link-lib=static=plt_scheduler");
+
     println!("cargo:rustc-link-lib=static=concordium_smart_contract_engine");
 
     println!("cargo:rustc-link-lib=dylib=gmp");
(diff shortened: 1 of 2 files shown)


scripts/static-libraries/build-static-libraries.sh
```diff
@@ -36,8 +36,11 @@ cp /build/concordium-base/smart-contracts/wasm-chain-integration/target/release/
 find ~/.stack/snapshots/x86_64-linux/ -type f -name "*.a" ! -name "*_p.a" -exec cp {} /target/vanilla/dependencies \;
 find ~/.stack/snapshots/x86_64-linux/ -type f -name "*_p.a" -exec cp {} /target/profiling/dependencies \;
 
+## Rust files - copy the static libraries to the target directory, and strip debug symbols from them to reduce their size, as 
+## they are not needed for the final binaries and can cause linking issues if they contain ruststd symbols
 mkdir -p /target/rust
 cp -r /build/concordium-base/rust-src/target/release/*.a /target/rust/
+cp -r /build/plt/target/release/*.a /target/rust/
```
limemloh: I don't think you need this line

squirer: removed now. originally i was copying this ahead of the first block, but makes more sense to be copy what we need per block, I agree

scripts/static-libraries/build-static-libraries.sh
```diff
@@ -85,10 +91,31 @@ find /target /binaries -type f -exec strip --strip-debug {} \;
...
+            echo "Removing file:"
+            echo $file
+            rm $file;
+        fi
+    done
+
+    set -e
+
+    rm libplt_scheduler.a
+    ar rcs libplt_scheduler.a *.o
```
limemloh, line 117: we need to link to it in the static part of build.rs as well

squirer: Done added.

concordium-node/build.rs
```diff
@@ -884,7 +884,7 @@ fn link_static_libs() -> std::io::Result<()> {
                             );
                         }
                     }
-                    _ => panic!("Unable to link ghc libs"),
+                    _ => panic!("Unable tod link ghc libs"),
```
td202: ?? Revert

```suggestion
                    _ => panic!("Unable to link ghc libs"),
```

td202 (approved):

PR #1595, opened 2026-03-19
Rely on toolchain.toml for Rust version and use windows-mscv instead …

allanbrondum:
## Purpose

Fix windows and macos build: Remove usages of the action `actions-rust-lang/setup-rust-toolchain` since it turns rustc warnings in base into errors via the environment variables it magically sets. We have no reason to use it anymore, since rustup is already installed in the runner images.

## Changes

Additional changes:
* rely on toolchain.toml to specify Rust version instead of passing the Rust version around
* prebuild node Rust library in the windows release before building consensus (building it indirectly via consensus Setup.hs without prebuilding fails for some unknown reason https://github.com/Concordium/concordium-node/pull/1595#discussion_r2986447691)
* move stack.static.yaml up next to stack.yaml in the repo root (both includes base and consensus as sub-packages)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/scripts/distribution/windows/build-all.ps1
+++ b/scripts/distribution/windows/build-all.ps1
@@ -1,9 +1,12 @@
-param ([string] $nodeVersion)
+param ([string] $rustVersion, [string] $nodeVersion)
 
 Write-Output "stack version: $(stack --version)"
 Write-Output "cargo version: $(cargo --version)"
 Write-Output "flatc version: $(flatc --version)"
 Write-Output "protoc version: $(protoc --version)"
+
+# Override the rust toolchain so that consensus rust dependencies use it.
+rustup override set $rustVersion-x86_64-pc-windows-gnu
 
 Write-Output "Building consensus..."
 stack build
@@ -14,7 +17,7 @@ stack exec -- cargo build --manifest-path concordium-node\Cargo.toml --release -
 if ($LASTEXITCODE -ne 0) { throw "Failed building node" }
 
 Write-Output "Building the collector..."
-cargo build --manifest-path collector\Cargo.toml --release --locked
+cargo +$rustVersion-x86_64-pc-windows-msvc build --manifest-path collector\Cargo.toml --release --locked
 if ($LASTEXITCODE -ne 0) { throw "Failed building the collector" }
 
 Write-Output "Building node runner service..."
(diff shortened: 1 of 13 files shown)


scripts/distribution/windows/build-all.ps1
```diff
@@ -1,18 +1,27 @@
...
 Write-Output "cargo version: $(cargo --version)"
 Write-Output "flatc version: $(flatc --version)"
 Write-Output "protoc version: $(protoc --version)"
 
-# Set the default rust toolchain so that consensus rust dependencies use it.
-rustup default $rustVersion-x86_64-pc-windows-gnu
+# Override the rust toolchain so that consensus rust dependencies use it.
+rustup override set $rustVersion-x86_64-pc-windows-gnu
+
+# Sometimes it causes errors to build the Rust library indirectly
```
allanbrondum: If we don't prebuild, we get this error for some reason:
```
concordium-consensus         >    Compiling chrono v0.4.42
concordium-consensus         >    Compiling darling_macro v0.21.3
concordium-consensus         >    Compiling ed25519 v2.2.3
concordium-consensus         >    Compiling crossbeam-epoch v0.9.18
concordium-consensus         >    Compiling hashbrown v0.13.2
concordium-consensus         >    Compiling bs58 v0.5.1
concordium-consensus         >    Compiling concordium-contracts-common-derive v4.1.0 (D:\a\concordium-node\concordium-node\concordium-base\smart-contracts\contracts-common\concordium-contracts-common-derive)
concordium-consensus         > error: failed to add native library D:\a\concordium-node\concordium-node\plt\target\release\deps\rustc0SvHM2\kernel32.dll_imports.lib: The system cannot find the file specified. (os error 2)
concordium-consensus         > 
concordium-consensus         > error: could not compile `chrono` (lib) due to 1 previous error
```

td202: To elaborate on this a little bit: when `Setup.hs` is run (which invokes the `cargo build`), `stack` puts the LLVM version of `dlltool` on the path, which is not compatible with the gnu `dlltool`. It thus does not generate the expected import lib. By the rust upfront, when it is subsequently built in `Setup.hs` it will already be built and so does not get rebuilt.

td202 (approved):

PR #1596, opened 2026-03-20
Implement getTokenAuthorizations in rust token module

limemloh:
## Purpose

Closing [COR-2234](https://linear.app/concordium/issue/COR-2234/concordium-nodeplt-extend-token-module-with-rbac-functionality)

[Depends on PR for base](https://github.com/Concordium/concordium-base/pull/912) 

## Changes

- Extend token kernel interface with iterator for a state prefix.
- Extend token module with get_token_authorization query.
- Implement test suite for RBAC assign and revoke operations, now that the query is ready.
- Update base.

### DIFF
--- a/plt/plt-block-state/src/block_state.rs
+++ b/plt/plt-block-state/src/block_state.rs
@@ -313,13 +313,7 @@ impl<IntState: HasQueryableBlockState, Load: BackingStoreLoad, ExtState: Externa
         token_key_value_state: &Self::TokenKeyValueState,
         prefix: TokenStateKey,
     ) -> impl Iterator<Item = (TokenStateKey, TokenStateValue)> {
-        let mut out = Vec::new();
-        for (key, value) in token_key_value_state.state.iter() {
-            if let Some(rest) = key.strip_prefix(prefix.as_slice()) {
-                out.push((rest.to_vec(), value.clone()));
-            }
-        }
-        out.into_iter()
+        token_key_value_state.iter_prefix(prefix)
     }
 }
 
@@ -393,4 +387,21 @@ struct Token {
 #[derive(Debug, Clone, Default, Serialize)]
 pub struct SimplisticTokenKeyValueState {
     state: BTreeMap<TokenStateKey, TokenStateValue>,
+}
+
+impl SimplisticTokenKeyValueState {
+    fn iter_prefix(
+        &self,
+        prefix: TokenStateKey,
+    ) -> impl Iterator<Item = (TokenStateKey, TokenStateValue)> {
+        // This is just a temporary implementation and will not scale.
+        // However implementation should be trivial once using the actual trie.
+        let mut out = Vec::new();
+        for (key, value) in self.state.iter() {
+            if let Some(rest) = key.strip_prefix(prefix.as_slice()) {
+                out.push((rest.to_vec(), value.clone()));
+            }
+        }
+        out.into_iter()
+    }
 }
--- a/plt/plt-block-state/src/block_state.rs
+++ b/plt/plt-block-state/src/block_state.rs
@@ -308,11 +308,11 @@ impl<IntState: HasQueryableBlockState, Load: BackingStoreLoad, ExtState: Externa
         self.protocol_version
     }
 
-    fn iter_token_state_prefix(
+    fn iter_token_state_prefix<'a>(
         &self,
-        token_key_value_state: &Self::TokenKeyValueState,
+        token_key_value_state: &'a Self::TokenKeyValueState,
         prefix: TokenStateKey,
-    ) -> impl Iterator<Item = (TokenStateKey, TokenStateValue)> {
+    ) -> impl Iterator<Item = (&'a TokenStateKey, &'a TokenStateValue)> {
         token_key_value_state.iter_prefix(prefix)
     }
 }
@@ -393,13 +393,13 @@ impl SimplisticTokenKeyValueState {
     fn iter_prefix(
         &self,
         prefix: TokenStateKey,
-    ) -> impl Iterator<Item = (TokenStateKey, TokenStateValue)> {
+    ) -> impl Iterator<Item = (&TokenStateKey, &TokenStateValue)> {
         // This is just a temporary implementation and will not scale.
         // However implementation should be trivial once using the actual trie.
         let mut out = Vec::new();
         for (key, value) in self.state.iter() {
-            if let Some(rest) = key.strip_prefix(prefix.as_slice()) {
-                out.push((rest.to_vec(), value.clone()));
+            if key.starts_with(prefix.as_slice()) {
+                out.push((key, value));
             }
         }
         out.into_iter()
--- a/plt/plt-block-state/src/block_state.rs
+++ b/plt/plt-block-state/src/block_state.rs
@@ -307,6 +307,20 @@ impl<IntState: HasQueryableBlockState, Load: BackingStoreLoad, ExtState: Externa
     fn protocol_version(&self) -> ProtocolVersion {
         self.protocol_version
     }
+
+    fn iter_token_state_prefix(
+        &self,
+        token_key_value_state: &Self::TokenKeyValueState,
+        prefix: TokenStateKey,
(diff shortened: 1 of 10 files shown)


plt/plt-token-module/src/roles.rs
```diff
@@ -11,6 +11,19 @@ pub const MANDATORY_ROLES: &[TokenAdminRole] = &[
     TokenAdminRole::UpdateMetadata,
 ];
 
+/// List all roles.
+const ALL_ROLES: &[TokenAdminRole] = &[
+    TokenAdminRole::UpdateAdminRoles,
+    TokenAdminRole::Pause,
+    TokenAdminRole::UpdateMetadata,
```
td202:
```suggestion
```
Duplicates

plt/plt-token-module/src/roles.rs
td202: Aside: I would call these `UNIVERSAL_ROLES` rather than `MANDATORY_ROLES`.

plt/plt-token-module/tests/utils/kernel_stub.rs
```diff
@@ -321,6 +322,19 @@ impl TokenKernelQueries for KernelStub {
...
+        &self,
+        prefix: TokenStateKey,
+    ) -> impl Iterator<Item = (TokenStateKey, TokenStateValue)> {
+        let mut out = Vec::new();
+        for (key, value) in self.state.iter() {
+            if let Some(rest) = key.strip_prefix(prefix.as_slice()) {
+                out.push((rest.to_vec(), value.clone()));
+            }
+        }
+        out.into_iter()
```
td202, line 336: This is potentially quite inefficient (e.g. compared to using `range`). That's fine as long as we switch the implementation when we use the `trie`.

plt/plt-token-module/src/key_value_state.rs
```diff
@@ -221,6 +221,51 @@ pub fn revoke_account_roles(
...
+            match role {
+                TokenAdminRole::UpdateAdminRoles => authorizations.update_admin_roles.push(account),
+                TokenAdminRole::Mint => authorizations.mint.push(account),
+                TokenAdminRole::Burn => authorizations.burn.push(account),
+                TokenAdminRole::UpdateAllowList => authorizations.update_allow_list.push(account),
+                TokenAdminRole::UpdateDenyList => authorizations.update_deny_list.push(account),
+                TokenAdminRole::Pause => authorizations.pause.push(account),
+                TokenAdminRole::UpdateMetadata => authorizations.update_metadata.push(account),
+            }
+        }
```
td202, line 280: As per my comment on the base PR, I would exclude roles that are not available given the token configuration.

td202 (approved):

plt/plt-token-module/tests/utils/kernel_stub.rs
limemloh: Yes, we it needs to be done differently in the end, but this is also just the testing stub.

plt/plt-block-state/src/block_state.rs
```diff
@@ -380,3 +388,20 @@ struct Token {
 pub struct SimplisticTokenKeyValueState {
     state: BTreeMap<TokenStateKey, TokenStateValue>,
 }
+
+impl SimplisticTokenKeyValueState {
+    fn iter_prefix(
+        &self,
+        prefix: TokenStateKey,
+    ) -> impl Iterator<Item = (TokenStateKey, TokenStateValue)> {
```
allanbrondum: I don't think it is nice that `TokenStateKey` now has two meanings: A full key or a suffix of a key. Maybe return full key or create a new type for suffixes?

plt/plt-block-state/src/block_state_interface.rs
```diff
@@ -81,20 +81,32 @@ pub trait BlockStateQuery {
...
         token_key_value: &Self::TokenKeyValueState,
         key: &TokenStateKey,
     ) -> Option<TokenStateValue>;
 
+    /// Get iterator over key-value pairs with a shared prefix.
+    ///
+    /// # Arguments
+    ///
+    /// - `token_key_value` The token module state to look up the value in.
+    /// - `prefix` The token state key prefix to iterate over.
```
allanbrondum, line 97: We should explicitly document what the first value in the returned pairs is, a full key or a suffix with the given prefix stripped.

allanbrondum (approved):

plt/plt-token-module/tests/token_module_admin_role.rs
```diff
@@ -0,0 +1,508 @@
...
+
+    // 1st transaction: pause the token
+    let mut execution = stub.execution_with_sender(gov_account);
+    let operations = vec![TokenOperation::Pause(TokenPauseDetails {})];
+    token_module::execute_token_update_transaction(
+        &mut execution,
+        &mut stub,
+        RawCbor::from(cbor::cbor_encode(&operations)),
+    )
+    .expect("execute");
```
soerenbf, line 235: Can also just 
```suggestion
    stub.set_paused(true);
```

plt/plt-token-module/tests/token_module_admin_role.rs
```diff
@@ -0,0 +1,508 @@
...
```
soerenbf, line 467: and here
```suggestion
    stub.set_paused(true);
```

plt/plt-token-module/tests/token_module_admin_role.rs
```diff
@@ -0,0 +1,508 @@
...
+            roles: vec![TokenAdminRole::UpdateAdminRoles],
+            account: CborHolderAccount::from(stub.account_canonical_address(&account2)),
+        },
+    )];
+    token_module::execute_token_update_transaction(
+        &mut execution,
+        &mut stub,
+        RawCbor::from(cbor::cbor_encode(&operations)),
+    )
+    .expect("execute");
```
soerenbf, line 571: Would it make sense to have convenience functionality `stub.set_account_roles`, similar to `stub.set_account_balance`? It might not be worthwhile, as the governance account returned from init will be sufficient for testing purposes 🤷

plt/plt-token-module/tests/token_module_admin_role.rs
limemloh: I think ideally these tests would be not rely on the internals, so using a transaction instead here.
We already have a task to move these up to the scheduler tests, where they can't rely on the internals.

plt/plt-token-module/tests/token_module_admin_role.rs
limemloh: As mentioned above, these will probably be moved to the scheduler tests and should not rely on the internals.
But it could make sense to have these helpers I guess even there.
I'll leave that to the one taking the task of moving these.

allanbrondum: `BlockStateWithExternalStateStubbed::increment_account_balance` does that e.g., it is a helper method that runs a transaction.

plt/plt-token-module/src/key_value_state.rs
limemloh: I have adjusted accordingly

plt/plt-token-module/tests/token_module_admin_role.rs
soerenbf: Alright, maybe I'm missing some context from the broader picture here 🤔

plt/plt-block-state/src/block_state.rs
limemloh: Right, check what the state trie is actually returning, and it seems to be the entire key, so I'll change it to do the same.

plt/plt-token-module/tests/token_module_admin_role.rs
limemloh: I created this [task](https://linear.app/concordium/issue/COR-2339/move-token-module-tests-to-be-scheduler-level) for moving these tests to the scheduler level.
Since it results in fewer tests to maintain overall and is a step towards removing the scheduler/token-module separation.

PR #1603, opened 2026-03-26
Add no-coin-info tests for token creation, transfer, and allow/deny list

limemloh:
## Summary

Extends `TokenHolderTransactions` tests with cases where the `coin_info` field is absent from the `untagged-account-address` (i.e. `CborAccountAddress` with `chaCoinInfo = Nothing` / the "short" form).

## New tests

Three new deterministic test functions are added to `TokenHolderTransactions.hs`, one per operation type. Each runs across all PLT-supporting protocol versions (P9, P10, P11).

### `testNoCoinInfoCreation`
Creates a PLT with the governance account using a short-form address (no coin info). Verifies:
- Token creation is **accepted**
- The queried `tmsGovernanceAccount` **always has coin info set**, because the state stores the account index and reconstructs the address with coin info on query

### `testNoCoinInfoTransfer`
Transfers tokens to a recipient specified with no coin info. Verifies:
- The transfer is **accepted**
- Resulting balances on sender and recipient are correct

### `testNoCoinInfoAllowDenyList`
Adds an account to the allow list and deny list using short-form target addresses. Verifies:
- Operations are **accepted**
- Output events carry the target address **exactly as specified in the operation** (no coin info if the operation had none)
- Reject reasons (for a transfer rejected by the allow-list check) **always report addresses with coin info set**
- The resulting account state correctly reflects list membership

## Key behaviours confirmed

| Scenario | Coin info in output |
|---|---|
| Queried state (governance account) | Always set (reconstructed from account index) |
| Reject reasons | Always set (use `accountTokenHolder`) |
| Allow/deny list output events | Mirrors the operation (absent if operation had none) |

### DIFF
--- a/concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
+++ b/concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
@@ -1385,6 +1385,311 @@ testMintBurn spv mintEnabled burnEnabled = do
             }
         ]
 
+-- | Test that a token transfer is accepted when the recipient address has no coin info.
+--  Verifies that the resulting balances are correct after the transfer.
+testNoCoinInfoTransfer ::
+    forall pv.
+    (IsProtocolVersion pv, PVSupportsPLT pv) =>
+    SProtocolVersion pv ->
+    String ->
+    Spec
+testNoCoinInfoTransfer _ pvString =
+    specify (pvString ++ ": Token transfer with no coin info in recipient address") $ do
+        let govAcct = CBOR.accountTokenHolder dummyAddress
+            -- Recipient address without coin info (the "short" / untagged form).
+            recptShort = CBOR.accountTokenHolderShort dummyAddress2
+            mintAmt = TokenAmount 100 0
+            transferAmt = TokenAmount 40 0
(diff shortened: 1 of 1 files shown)


concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
```diff
@@ -1385,6 +1385,381 @@ testMintBurn spv mintEnabled burnEnabled = do
             }
         ]
 
+-- | Test that token creation is accepted when the governance account address has no coin info.
+--  Verifies that the queried state always reports the governance account with coin info set,
+--  regardless of whether coin info was present in the initialization parameters.
+testNoCoinInfoCreation ::
```
allanbrondum: Maybe this test belongs more in `TokenCreation.hs`

allanbrondum (approved):

concordium-consensus/tests/scheduler/SchedulerTests/TokenHolderTransactions.hs
limemloh: It seems we already have this test in TokenCreation.hs, so removing it instead.

PR #1604, opened 2026-03-27
Add locks to simplistic block state

soerenbf:
Ref COR-2213
Depends on https://github.com/Concordium/concordium-base/pull/919 for some of the types.

## Purpose

Add the locks state model to the simplistic block state.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-block-state/src/block_state.rs
+++ b/plt/plt-block-state/src/block_state.rs
@@ -3,8 +3,8 @@
 use crate::block_state::blob_store::{BackingStoreLoad, BackingStoreStore, DecodeError};
 use crate::block_state::external::{ExternalBlockStateOperations, ExternalBlockStateQuery};
 use crate::block_state::types::{
-    AccountWithCanonicalAddress, TokenAccountState, TokenConfiguration, TokenIndex, TokenStateKey,
-    TokenStateValue,
+    AccountWithCanonicalAddress, LockConfiguration, TokenAccountState, TokenConfiguration,
+    TokenIndex, TokenStateKey, TokenStateValue,
 };
 use crate::block_state_interface::{
     BlockStateOperations, BlockStateQuery, OverflowError, RawTokenAmountDelta,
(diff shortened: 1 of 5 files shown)


plt/plt-block-state/src/block_state.rs
```diff
@@ -366,17 +367,39 @@ impl<Load: BackingStoreLoad, ExtState: ExternalBlockStateOperations> BlockStateO
...
+#[derive(Debug, Clone, Serialize)]
+struct Lock {
+    /// The balances locked per account, mapping each account index to its
+    /// locked token and amount.
+    #[map_size_length = 4]
+    locked_balances: BTreeMap<AccountIndex, TokenAndAmount>,
+    /// The configuration parameters for the lock.
+    configuration: LockConfiguration,
+    /// The key-value state associated with the lock.
+    key_value_state: SimplisticTokenKeyValueState,
```
allanbrondum: I suggest no more key-value states. We should save in a dedicated explicit model. Or if you need a trie/tree, then just use a `BTreeMap` for now.

allanbrondum (commented):
I think my only concern is the introduction of `key_value_state: SimplisticTokenKeyValueState` in the lock state.

plt/plt-scheduler-types/src/types/tokens.rs
```diff
@@ -112,6 +113,15 @@ pub enum TokenHolder {
     Account(AccountAddress),
 }
 
+/// A token amount and the associated token ID as a pair.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Serialize)]
+pub struct TokenAndAmount {
+    /// The token ID corresponding to the `amount`.
+    pub token_id: TokenId,
```
limemloh: Should this use the TokenIndex instead?
I think internally we should avoid storing the token ID more than once in the block state

limemloh (approved):
LGTM for a POC

plt/plt-scheduler-types/src/types/tokens.rs
soerenbf: Yes that makes sense, good catch.

plt/plt-block-state/src/block_state.rs
soerenbf: Might make more sense with a dedicated model, I don't have a complete overview of the state requirements for the lock controllers yet though.

Added an empty locks state model so far, which will be filled in at a later point.

PR #1605, opened 2026-03-27
MetaUpdate transaction

td202:
## Purpose

Support execution of `MetaUpdate` transaction in P11.

## Changes

- Hand off handling of MetaUpdate to Rust scheduler from Haskell scheduler.
- `execute_meta_update_transaction`: execute the transaction operation-wise.
- Refactor `execute_token_update_operation_at_index` out of `execute_token_update_operations` for use in `execute_meta_update_transaction`.

## Checklist

      hard-to-understand areas.


concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -483,6 +485,10 @@ dispatchTransactionBody msg CheckHeaderResult{..} = do
     onlyWithPLT c = case sSupportsPLT (accountVersion @(AccountVersionFor (MPV m))) of
         SFalse -> error "Operation unsupported at this protocol version."
         STrue -> c
+    onlyWithPLTV1 :: ((PltStateVersionFor (MPV m) ~ PLTStateV1) => a) -> a
+    onlyWithPLTV1 c = case sPltStateVersionFor (protocolVersion @(MPV m)) of
+        SPLTStateV1 -> c
```
allanbrondum, line 491: Would it be better to have an explicit configuration of whether the meta update operation is supported. Rather than piggy backing it on `PLTStateV1`? In the future, we might event change `PLTStateV1` to be used on P9 and P10, so it seems like a bad idea to link `PLTStateV1` together with the meta update operation support.

plt/plt-scheduler-types/src/types/reject_reasons.rs
```diff
@@ -16,6 +16,8 @@ use concordium_base::protocol_level_tokens::{RawCbor, TokenId, TokenModuleCborTy
 /// Corresponding Haskell type: `Concordium.Types.Execution.RejectReason`
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum TransactionRejectReason {
+    /// The transaction payload could not be fully deserialized.
+    SerializationFailure,
```
allanbrondum, line 20: We should extend the tests below

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -118,6 +120,119 @@ pub fn execute_token_update_transaction<
...
+/// - `transaction_execution` Context of transaction execution that allows accessing sending account
+///   and charging energy.
+/// - `block_state` Block state that can be queried and updated during execution.
+/// - `payload` The meta-update transaction payload to execute.
+///
+/// # Errors
+///
+/// - [`TransactionExecutionError`] If executing the transaction fails with an unrecoverable error.
+///   Returning this error will terminate the scheduler.
+pub fn execute_meta_update_transaction<
```
allanbrondum, line 143: Notice that the logic in this function is duplicated and without test coverage.

allanbrondum (dismissed):
I think we should consider how we handle the duplicated logic and need for test coverage of the duplicated logic. My suggestion would be to route the token update operations into the meta update logic in the PLT Scheduler.

concordium-consensus/src/Concordium/Scheduler.hs
```diff
@@ -2750,6 +2756,19 @@ handleTokenUpdateHaskellManaged depositContext tokenId tokenOperations =
             (res, events, energyUsed) <- runPLTWithEnergy tokenIndex energy $ TokenModule.executeTokenUpdateTransaction tc parameter
             return ((events <$ res, energyUsed), isLeft res)
 
+-- | Handler for a meta update transaction.
+handleMetaUpdate ::
```
allanbrondum, line 2761: We should have automatic tests of this type of transaction also in some way or another.

concordium-consensus/src/Concordium/Scheduler.hs
td202: `PLTStateV1` is necessary but not sufficient for the meta update operation. In particular, the call to `decodePayload` is what determines if the payload should be supported at the current protocol version (which checks `supportsMetaUpdate`). If we hit the error case here then the node will crash - this is just asserting that the protocol version meets the constraints. We could have built the check in to `decodePayload`, but we'd need to have a dependently-typed version of `Payload` which didn't seem to be quite worth the effort.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
td202: I have added tests at the scheduler level for the new transaction type. The logic for actually performing individual operations is shared between the two transaction types, so there's not a need to test the details of the functionality in depth across both.

concordium-consensus/src/Concordium/Scheduler.hs
td202: Added

soerenbf (approved):
lgtm

PR #1606, opened 2026-04-01
Expose unrecoverable errors from rust scheduler to Haskell 

limemloh:
## Purpose

Closes PSR-39 [link](https://linear.app/concordium/issue/PSR-39/implement-strategy-for-handling-unrecoverable-errors-in-the-rust-code)

To ensure a panic on the rust scheduler implementation does not reach the haskell side, we use catch_unwind when exposed in the FFI to convert panics into a proper error code with panic message, allowing the haskell side to report this message as a proper error.

## Changes

- Introduce our own variant of `catch_unwind` which extracts the error message from panics.
- Added unit tests for ensuring panics with both `str` and `string` messages are extracted.
- Wrap each FFI call in the scheduler with `catch_unwind` and pass the panic message to haskell (note this does not include the blockstate FFI, to be discussed).
- Add unit test with panic in `ffi_execute_transaction`.
- Parsing of the panic status and error message on the haskell side.

### DIFF
--- a/plt/plt-scheduler/src/ffi/scheduler.rs
+++ b/plt/plt-scheduler/src/ffi/scheduler.rs
@@ -2,6 +2,7 @@
 //!
 //! It is only available if the `ffi` feature is enabled.
 
+use crate::ffi::status;
 use crate::scheduler;
 use concordium_base::base::{AccountIndex, Energy, ProtocolVersion};
 use concordium_base::contracts_common::AccountAddress;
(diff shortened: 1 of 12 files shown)


plt/plt-scheduler/src/ffi/scheduler.rs
```diff
@@ -246,69 +241,122 @@ extern "C" fn ffi_execute_chain_update(
...
+        UNIMPLEMENTED_READ_TOKEN_ACCOUNT_BALANCE, UNIMPLEMENTED_TOUCH_TOKEN_ACCOUNT,
+        UNIMPLEMENTED_UPDATE_TOKEN_ACCOUNT_BALANCE,
+    };
+
+    use super::*;
+    use std::ptr;
+
+    /// Test ensuring panics are caught and returned properly when providing invalid arguments to `ffi_execute_transaction`.
+    #[test]
+    fn test_execute_transaction_catches_panic() {
```
allanbrondum, line 326: It is possible to implement a test on the Haskell side also?

allanbrondum (approved):

plt/plt-scheduler/src/ffi/scheduler.rs
limemloh: Right, I did look into this, but did not find an easy way to test it.
I'll take another look into this, but it might mean exposing some unsafe functions in RustPLTScheduler.hs

limemloh: Added a similar test on the haskell side

PR #1607, opened 2026-04-01
Ar/psr 77 hashed cachaeble reference

allanbrondum:
## Purpose

Implement `block_state::blob_reference::hashed_cacheable_reference::HashedCacheableRef`.

## Changes

* implemented traits for caching, hashing and refactored blob store traits a bit
* implemented `hashed_cacheable_reference` module (structure largely matching Haskell side)
* consistently named the storage "blob store" and removed the term "backing store"

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+/// This [LockRef] achieves the following properties
+/// - it is cheap to clone
+/// - it allows for inner mutability
+/// - it is safe to use in a concurrent context
+#[derive(Debug)]
+pub struct LockRef<V> {
+    lock_ref: Arc<RwLock<V>>,
+}
+
+impl<V> Clone for LockRef<V> {
```
limemloh, line 52: Any reason not to derive this implemenation?
I don't see it, but if there is, it would be nice to have it in a comment

plt/plt-block-state/src/block_state/utils.rs
```diff
@@ -0,0 +1,72 @@
+//! Block state utility types and functions.
+
+use std::ops::Deref;
+use std::sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard};
+
+/// Value of type `T` that is either owned or borrowed.
+pub enum OwnedOrBorrowed<'a, T> {
```
limemloh, line 10: Every 2nd week I am wondering why we are not using `Cow` here 😅 
Could you document why this exists

plt/plt-block-state/src/block_state/hash.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn hash(&self, loader: &impl BlobStoreLoad) -> BlockStateResult<Hash>;
+}
+
+impl<T: Serial> Hashable for StoreSerialized<T> {
+    fn hash(&self, _loader: &impl BlobStoreLoad) -> BlockStateResult<Hash> {
+        Ok(hash_of_serialization(&self.0))
+    }
+}
+
+/// Calculate has by digesting the bytes of two hashes.
```
limemloh:
```suggestion
/// Calculate hash by digesting the bytes of two hashes.
```

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,623 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::block_state::blob_store;
+    use crate::block_state::blob_store::StoreSerialized;
+    use crate::block_state::blob_store::test_stub::{BlobStoreStub, UnreachableBlobStore};
+    use assert_matches::assert_matches;
+
+    type TestRef = HashedCacheableRef<StoreSerialized<u64>>;
+
```
limemloh, line 274: I think it would make sense to have a fixture for the byte representation in the blobstore, also to ensure it is compatible with the haskell version.

limemloh (approved):

plt/plt-block-state/src/block_state/blob_store.rs
```diff
@@ -1,45 +1,195 @@
...
+impl<T: Storable> Storable for &mut T {
+    fn store_to_buffer(&self, buffer: impl Buffer, storer: &mut impl BlobStoreStore) {
+        (**self).store_to_buffer(buffer, storer)
+    }
+}
+
+/// Adapter for types implementing [`Serialize`](common::Serialize) that
+/// allows them to be used as block state components.
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
+pub struct StoreSerialized<T>(pub T);
```
limemloh, line 67: Can we just implement the traits everywhere instead and remove this?
I don't think this type reduces the amount of code significantly, and it adds noise to wrap and unwrap everywhere.

plt/plt-block-state/src/block_state/blob_store.rs
```diff
@@ -1,45 +1,195 @@
...
+    storable: impl Storable,
+) -> BlobStoreLocation {
+    let mut buffer = Vec::new();
+    storable.store_to_buffer(&mut buffer, storer);
+    storer.store_raw(buffer)
+}
+
+/// Extension trait for [`common::ParseResult`] that allows mapping error type
+/// to [`DecodeError`].
+pub trait ParseResultExt<T> {
```
limemloh, line 117: Why not implement `From<common::ParseError> for BlockStateError`?

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,623 @@
...
+        Ok(Self::from_inner(inner))
+    }
+}
+
+impl<V: Storable> Storable for HashedCacheableRef<V> {
+    fn store_to_buffer(&self, mut buffer: impl Buffer, storer: &mut impl BlobStoreStore) {
+        let mut inner = self.inner.write();
+        let reference = inner.repr.get_reference_or_store(storer);
+        buffer.put(reference);
```
limemloh, line 239: it hurts a bit knowing we waste storage and by having a blob reference to another blob reference in several places. If we were not forced to write to the buffer, it could be avoided, but seems haskell does the waste as well. 😢

plt/plt-block-state/src/block_state/utils.rs
allanbrondum: Documented it, `Cow` needs `T: Clone`

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: There are higher level test in coming PR for the entire block state: https://github.com/Concordium/concordium-node/pull/1615

plt/plt-block-state/src/block_state/blob_store.rs
allanbrondum: The alternative is a blanked implementation: `impl<T: Serial> Storable for T {..}`. I had that once, but I kind of ended up this this explicit adapter. It is more explicit and less magic. I think it is easier to understand block state types this way. And we don't have an abundance of types that implement `Storable/Loadable` even though they are never in the blob store.

plt/plt-block-state/src/block_state/blob_store.rs
allanbrondum: I did not, because `ParseError` is a type alias (not a newtype) of `anyhow::Error`.

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: We have that some places yes, but I don't really see how you can tell from this line of code. Again, `Storable`/`Loadable` can store a composite part. There are many block state that is a list of different block state pointeres. So I don't really see how this implementation can be a sing of any problems.

limemloh: The extra blob reference could be avoided by changing `Loadable` and `Storable` to not force the buffer, something like:
```rust
fn store_to_blobstore(
        &self,
        storer: &mut impl BlobStoreStore
    ) -> Reference {
        let mut inner = self.inner.write();
        inner.repr.get_reference_or_store(storer);
    }

fn load_from_location(location: Reference, loader: &impl BlobStoreLoad) -> BlockStateResult<Self> {
        let inner = HashedBufferedRefInner {
            hash: None,
            repr: HashedCacheableRefRepr::Store { reference: location },
        };
        Ok(Self::from_inner(inner))
    }
```

PR #1608, opened 2026-04-05
LFMB tree

allanbrondum:
## Purpose

Implement LMFB tree that matches the blob storage and hashing of the Haskell implementation.

## Changes

Implementation is documented in the `lfmb_tree` module. There are test of hashing and storage that shows it matches Haskell side.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -0,0 +1,1220 @@
...
+}
+
+impl<K: LFMBTreeKey, V> Default for LFMBTree<K, V> {
+    fn default() -> Self {
+        Self::empty()
+    }
+}
+
+/// Trait implemented by tree keys, which allows them to be bijectively mapped
+/// to `u46` values.
```
limemloh:
```suggestion
/// to `u64` values.
```

plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -0,0 +1,1220 @@
...
+///            /             \
+///          (1)             (1)
+///         /   \           /   \
+///      (0)     (0)     (0)     (0)
+///      / \     / \     / \     / \
+///    [A] [B] [C] [D] [E] [F] [G] [H]
+///    #0  #1  #2  #3  #4  #5  #6  #7
+/// ```
+#[derive(Debug, Clone)]
+pub struct LFMBTree<K, V> {
```
limemloh: **Nitpicking:** I think the rust convention is to use lower case even for abbreviations:
```suggestion
pub struct LfmbTree<K, V> {
```

plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -0,0 +1,1220 @@
...
+    /// Access the value for the given `key` in the tree
+    /// using the `read` closure. Returns the value returned by the
+    /// closure if there is a value for the key in the tree, or `None` if there
+    /// is no value for the key.
+    ///
+    /// # Arguments
+    ///
+    /// - `loader`: Loader for the blob store the tree is stored in.
+    /// - `key`: The key to access the value for.
+    /// - `read`: Closure that is given the value, either as owned or borrowed,
```
limemloh, line 207: I have forgotten why do we need this `read` closure again, could you maybe elaborate in the doc.

plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -0,0 +1,1220 @@
...
+    fn with_value<T>(
+        &self,
+        loader: &impl BlobStoreLoad,
+        key: SubtreeKey,
+        read: impl FnOnce(OwnedOrBorrowed<'_, V>) -> BlockStateResult<T>,
+    ) -> BlockStateResult<T>
+    where
+        V: Loadable,
+    {
+        match self {
```
limemloh, line 433: Did you try implementing this function using a loop instead of recursion?

plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -0,0 +1,1220 @@
...
+                        HashedCacheableRef::new(new_left),
+                        right_ref.clone(),
+                    )
+                }
+            }
+        })
+    }
+}
+
+struct ValuesIterator<'a, L, F, V> {
```
limemloh, line 623: Some documentation on the fields would be nice here

plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -0,0 +1,1220 @@
...
+                j += 1;
+                assert_eq!(values.len(), (i - j) as usize);
+            }
+            assert_eq!(values.len(), 0);
+            assert_eq!(values.next().transpose().unwrap(), None);
+            assert_eq!(values.len(), 0);
+        }
+    }
+
+    /// Test [`LFMBTree::with_value`]
```
limemloh:
```suggestion
    /// Test [`LFMBTree::update_value`]
```

limemloh (approved):

plt/plt-block-state/src/block_state/lfmb_tree.rs
allanbrondum: It is because of the `RwLock`, you cannot returned the value as borrowed from this function. You could potentially return a projected/mapped guard (but that is unstable in std lib https://doc.rust-lang.org/std/sync/struct.RwLockReadGuard.html#method.map), but I think that will make it just more complicated, especially since there are. multiple layers of `RwLock`s.

plt/plt-block-state/src/block_state/lfmb_tree.rs
allanbrondum: No, but I guess it would be fairly simple to do.

PR #1609, opened 2026-04-05
Ar/psr 11 port the plt block state to rust

allanbrondum:
## Purpose

Implementation of Rust block state (this state replaces the previous "dummy" state, except for the smart contract trie; there is separate task to replace the dummy key-value store with the trie). The state builds on types representing immutable values, with the top level type being `block_state::BlockState`. The only type representing mutability is `block_state::MutableBlockState`.

## Changes

* implement block state queries and operations for `block_state::BlockState`. The only type representing mutability is `block_state::MutableBlockState`
* implement PLT state in `block_state::state::protocol_level_tokens::ProtocolLevelTokens` (using persistent map `im::HashMap` for the token id to token index map)
* added `loader` to `Loadable::load_from_buffer` to support loading token id to token index map
* the changes in other modules than `block_state` and `block_state::state::protocol_level_tokens` are mostly consequences of renaming and other minor cleanup.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/state/protocol_level_tokens.rs
```diff
@@ -0,0 +1,325 @@
...
+
+impl Loadable for ProtocolLevelTokens {
+    fn load_from_buffer(
+        mut buffer: impl Read,
+        loader: &impl BlobStoreLoad,
+    ) -> BlockStateResult<Self> {
+        let tokens: LFMBTree<TokenIndex, Token> = Loadable::load_from_buffer(&mut buffer, loader)?;
+        // To construct the full token id to token index map, we need to read the LFMBTree from
+        // the blob store. This is not ideal. If the state is to be cached after loading, we would
+        // rather wait until it is cached in memory before constructing the map.
```
allanbrondum, line 202: Although not ideal, this is exactly as the Haskell implementation does.

plt/plt-block-state/src/block_state/blob_store.rs
```diff
@@ -32,7 +32,13 @@ pub trait Loadable: Sized {
     /// If the value is composed of [blob references](super::blob_reference), these references should not
     /// have their values loaded into memory as a result of the [`Self::load_from_buffer`] operation.
     /// As such, [`Self::load_from_buffer`] is a "shallow" operation.
-    fn load_from_buffer(buffer: impl Read) -> Result<Self, BlockStateError>;
+    ///
+    /// The given `loader` should generally not be used. If it is needed, it is generally a warning
+    /// sign that the state might not have the right model.
+    fn load_from_buffer(
+        buffer: impl Read,
+        loader: &impl BlobStoreLoad,
```
allanbrondum, line 48: Had to add `loader` in order to support creating the token id to token index map when loading `ProtocolLevelTokens`

plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -252,17 +252,19 @@ impl<K: LFMBTreeKey, V> LFMBTree<K, V> {
     pub fn values<'a, L: BlobStoreLoad, F, T>(
         &self,
         loader: &'a L,
-        read: F,
+        mut read: F,
     ) -> impl ExactSizeIterator<Item = BlockStateResult<T>> + use<'a, K, V, L, F, T>
     where
         V: Loadable,
-        F: FnMut(OwnedOrBorrowed<'_, V>) -> BlockStateResult<T>,
+        F: FnMut(K, OwnedOrBorrowed<'_, V>) -> BlockStateResult<T>,
```
allanbrondum, line 259: Added key to values iterator `read` closure

plt/plt-block-state/src/block_state/types.rs
```diff
@@ -1,42 +1,8 @@
 //! Types used specifically in the block state.
 
-use concordium_base::common::Serialize;
```
allanbrondum, line 3: Moved into submodule

plt/plt-block-state/src/block_state/blob_store.rs
limemloh: I think this trait is getting a bit off now. 

Before it was just capturing the byte deserialization and kind of duplicating the purpose of `concordium_base::common::Deserial` (so already off if you ask me). 
Now it still expects something outside to load the first blobref, but also has access to the loader and could actually load it... 

I think a more clear trait boundary would be that it handles loading a type from a reference and then `common::Deserial` can be used within implementations for the decoding.

```rust
trait Loadable {
  fn load_from_store(
    loader: &impl BlobStoreLoad,
    location: BlobStoreLocation,
)  -> Result<Self, BlockStateError>;
}
```

limemloh: Thinking about it, my suggestion might not be trivial to make compatible with the haskell approach

allanbrondum: I you look at Haskell and wasm-chain-integration, they both do a combination via implementation inheritance, e.g. `Loadable` in wasm-chain-integration:
```
pub trait Loadable: Sized {
    fn load<S: std::io::Read, F: BackingStoreLoad>(
        loader: &mut F,
        source: &mut S,
    ) -> LoadResult<Self>;

    fn load_from_location<F: BackingStoreLoad>(
        loader: &mut F,
        location: Reference,
    ) -> LoadResult<Self> {
        let mut source = std::io::Cursor::new(loader.load_raw(location)?);
        Self::load(loader, &mut source)
    }
}
```

allanbrondum: I wanted to avoid that, because I think it is one of the things that makes both the Haskell code and the wasm-chain-integration confusing to read. Hence I chose that `Loadable` and `Storable` always relates to a buffer, since it is how most values are loaded and stored (most values are not directly stored at a blob store location, but as part of a compound value). And the functions `blob_store::load_from_store` and `blob_store::store_to_store`, then deal with loading values stored at a given blob store location (which will be values inside blob references). I think it gives a clear separation of responsibilities.

limemloh: I disagree with there being separation of responsibilities with this definition :sweat_smile: Loadable is a now handling partial decoding and partial loading from blobstore location, it means we can spread the loadable definition across more types, but this is just splitting up the code, not separating responsibilities.
The values which are not directly stored at a blob store location, should just implement `Deserial` and not be `Loadable`, because they, by definition, cannot be loaded from a blob store location.

allanbrondum: TBH I think you need to work a bit more with the code, it is easy to have ideas, but we need to implement something that works and is consistent :) For many reasons, I don't think using `Deserial` is a good idea (just the reasons that you cannot do the "dual" with `Serial` - since you need the storer - rules it you in my opinion).

allanbrondum: When I say it has a clearer separation is because as I implemented it, the `Loadable` is implemented for types that can be stored in the blob store, and that is all the trait does. In wasm-chain-integration (and to some extend in Haskell, just in a different way) the trait has the additional function (in addition to what you call "partial" loading), of being able to load from a specific blob store location. I is cleaner to only have one function/responsibility for the trait. And the implementer of `Loadable` should not really care if the value is stored as the direct target of a blob reference, or as part of a compound value.

PR #1612, opened 2026-04-08
Ar/psr 84 use smart contract trie for key value store

allanbrondum:
## Purpose

Change the token key-value state to be implemented by the smart contract trie.

## Changes

* implemented adapter for the smart contract trie in `block_state::smart_contract_trie`
* changed `state::protocol_level_tokens` to use the smart contract trie adapter

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/smart_contract_trie.rs
```diff
@@ -0,0 +1,446 @@
...
+use crate::block_state::hash::Hashable;
+use crate::block_state_interface::{BlockStateFailure, BlockStateResult};
+use concordium_base::common::Buffer;
+use concordium_base::hashes::Hash;
+use concordium_smart_contract_engine::v1::trie;
+use std::io::Read;
+use std::sync::{Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
+
+/// Immutable (persistent) trie. The internal structure may be changed via interior mutability,
+/// but the entries in the trie never changes. This is the frozen/persistent dual to [`MutableState`].
```
td202, line 20:
```suggestion
/// but the entries in the trie never change. This is the frozen/persistent dual to [`MutableState`].
```

plt/plt-block-state/src/block_state/smart_contract_trie.rs
```diff
@@ -0,0 +1,446 @@
...
+    fn lock_write(&self) -> RwLockWriteGuard<'_, trie::PersistentState> {
+        self.0.write().expect("PersistentState lock poisoned")
+    }
+
+    fn lock_read(&self) -> RwLockReadGuard<'_, trie::PersistentState> {
+        self.0.read().expect("PersistentState lock poisoned")
+    }
+}
+
+struct PrefixIterator<'a, L> {
```
td202, line 74: Should there be a `FusedIterator` implementation for this?

td202 (approved):

limemloh (approved):

plt/plt-block-state/src/block_state/smart_contract_trie.rs
allanbrondum: I don't think it is essential to implement it. Also when using return type position `impl Iterator`, you can only specify one trait. And then `ExactSizeIterator` is probably more useful.

PR #1616, opened 2026-04-10
Disable lto for plt workspace

limemloh:
## Purpose

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/plt/Cargo.toml
+++ b/plt/Cargo.toml
@@ -12,7 +12,6 @@ members = [
 
 # Optimize a good as possible with fat linking and a single code generation unit
 [profile.release]
-lto = "fat"
 codegen-units = 1
 
 [workspace.dependencies]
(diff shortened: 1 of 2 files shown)


td202 (approved):

plt/Cargo.toml
```diff
@@ -12,7 +12,6 @@ members = [
 
 # Optimize a good as possible with fat linking and a single code generation unit
 [profile.release]
-lto = "fat"
```
allanbrondum, line 15: Shouldn't we keep the link time optimization, if that is not causing the problem?

limemloh: I am a bit back and forth on this one... it feels a bit risky since we are stripping object files with certain symbols in the static build and to my understanding this might inline across object files, meaning some of rust-std might be inlined. Not sure if this is actually an issue in practice.
I think this makes more sense on the top-level binary, i.e put it in `concordium-node/Cargo.toml`.

limemloh: Been thinking about this a bit more and now my conclusion is that we would hit linker issues if this causes issue in the future, so we might as well keep it for now.

PR #1619, opened 2026-04-13
Sz/locks/cor 2214 lock controller interface

soerenbf:
Closes COR-2214

## Purpose

Adds a common interface to be implemented by the different lock controllers.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state_interface.rs
```diff
@@ -39,6 +44,10 @@ pub trait BlockStateQuery {
     /// The token is guaranteed to exist on chain, when holding an instance of this type.
     type Token;
 
+    /// Opaque type that represents a lock on chain.
+    /// The lock is guaranteed to exist on chain, when holding an instance of this type.
```
td202, line 48: I am not sure this is true. Unlike Tokens, Locks can be destroyed.

plt/Cargo.toml
```diff
@@ -17,6 +18,7 @@ codegen-units = 1
 
 [workspace.dependencies]
 plt-block-state = {path = "plt-block-state"}
+plt-lock-module = {path = "plt-lock-module"}
```
allanbrondum, line 21: I suggest moving this into `plt-scheduler` and not add a new crate. The crate `plt-token-module` will also soon be removed as part of removing the token kernel trait.

plt/plt-lock-module/src/lock_controller.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn approve_operation<BSQ: BlockStateQuery>(
+        &self,
+        bsq: &BSQ,
+        sender: &BSQ::Account,
+        lock: &BSQ::Lock,
+        operation: &LockOperation,
+    ) -> Result<(), LockControllerRejectReason>;
+
+    /// Query controller configuration and state as CBOR.
+    fn query_info<BSQ: BlockStateQuery>(&self, bsq: &BSQ, lock: &BSQ::Lock) -> RawCbor;
```
allanbrondum, line 32: Since we now use CBOR just as encoding format, and not as a mechanism to encapsulate dynamic/polymorphic behavior, I think we should just return a concrete type here. The returned value can be an enum, if we want to express that is can be multiple types.

td202 (approved):

plt/plt-lock-module/src/lock_controller.rs
```diff
@@ -0,0 +1,63 @@
...
+/// defined in `concordium_base` once it exists.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum LockOperation {}
+
+/// Lock-controller specific rejection.
+///
+/// TODO: Replace this placeholder with the corresponding lock reject reason type
+/// defined in `concordium_base` once it exists.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct LockControllerRejectReason;
```
soerenbf: As stated, these are placeholders. It could potentially make sense for the controller to define it's own operation type to make it easier to support both token/meta update transactions and use the assurance of the opaque types from the block state associated types, i.e.

```rust
enum LockOperationDetails<BSQ: BlockStateQuery> {
    Fund { lock: BSQ::Lock, ... },
    Send { lock: BSQ::Lock, source: BSQ::Account, ... },
    Return { lock: BSQ::Lock, source: BSQ::Account, ... },
    Cancel { lock: BSQ::Lock, ... }
}
```

plt/plt-lock-module/src/lock_controller.rs
```diff
@@ -0,0 +1,63 @@
...
+
+/// Runtime interface implemented by protocol-level locks.
+pub trait LockController {
+    /// Approve or reject a lock operation.
+    ///
+    /// * `bsq`: the block state to query on
+    /// * `sender`: the transaction sender reference
+    /// * `lock`: the lock reference. This is used instead of the lock ID from the operation.
+    /// * `operation`: the lock operation to approve/reject.
+    fn approve_operation<BSQ: BlockStateQuery>(
```
td202, line 29: I would suggest a more neutral name such as `validate_operation`.

plt/plt-block-state/src/block_state_interface.rs
soerenbf: hmm, that's a good point.. Will this be an issue though? is everything not synchronous, meaning that the lock is guaranteed to exist at the time you get the value and until you destroy it yourself?

plt/plt-lock-module/src/lock_controller.rs
soerenbf: yeah good point. Will revise.

plt/Cargo.toml
soerenbf: Cool, will do.

plt/plt-block-state/src/block_state_interface.rs
td202: Sure, but holding an instance of the type is not enough to guarantee its existence.

soerenbf: Removed the comment.

PR #1620, opened 2026-04-14
Ar/psr 11 oncelock

allanbrondum:
## Purpose

Replace the usages of `RwLock` with `OnceLock`, that is a simpler and more lightweight synchronization primitive. Reading the value in a `OnceLock` e.g. has almost no overhead. And you can borrow the value (no guard needed like for `RwLock`).

## Changes

* Changed `HashedCacheableRef` to use `OnceLock`
* Changed `HashedCacheableRef::with_value` to `HashedCacheableRef::value` that borrows the value
  from `HashedCacheableRef` is it is memory (no read closure needed).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
+++ b/plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
@@ -256,6 +256,23 @@ mod tests {
 
     type TestRef = HashedCacheableRef<StoreSerialized<u64>>;
 
+    fn assert_in_memory_repr<V>(hcr: &HashedCacheableRef<V>) -> &V {
+        assert!(hcr.inner.blob_location.get().is_none(), "in blob store");
+        hcr.inner.value.get().expect("not in memory")
+    }
+
+    fn assert_stored_repr<V>(hcr: &HashedCacheableRef<V>) -> BlobStoreLocation {
+        assert!(hcr.inner.value.get().is_none(), "in memory");
+        *hcr.inner.blob_location.get().expect("not in blob store")
+    }
+
+    fn assert_cached_repr<V>(hcr: &HashedCacheableRef<V>) -> (&V, BlobStoreLocation) {
+        (
+            hcr.inner.value.get().expect("not in memory"),
+            *hcr.inner.blob_location.get().expect("not in blob store"),
+        )
+    }
+
     /// Test full lifecycle of a value:
     ///
     /// * create as new in memory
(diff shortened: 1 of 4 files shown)


plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -141,127 +138,124 @@ impl<V> HashedCacheableRefRepr<V> {
...
-                // in order to borrow the value just written.
-                let HashedCacheableRefRepr::Cache { value, .. } = self else {
-                    unreachable!("not HashedBufferedRefRepr::Cache though it was just written")
-                };
-                value
+            HashedCacheableRefRepr::Store {
+                blob_location,
+                value_lock,
+            } => {
+                match value_lock.get() {
```
limemloh, line 171: Did you try using [`get_or_init`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.get_or_init) here as well ?
I guess we lose the log to stderr then

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -272,9 +266,37 @@ mod tests {
     use crate::block_state::blob_store::StoreSerialized;
     use crate::block_state::blob_store::test_stub::{BlobStoreStub, UnreachableBlobStore};
     use assert_matches::assert_matches;
+    use std::fmt::Debug;
 
     type TestRef = HashedCacheableRef<StoreSerialized<u64>>;
 
+    fn assert_in_memory_repr<V: Debug>(hcr: &HashedCacheableRef<V>) -> &V {
```
limemloh, line 281: Maybe make these macros, otherwise I think the line number in the error message is going to be here and not within the relevant test.

limemloh (approved):

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: That true, would probably be better with a macro, but you should have the stacktrace still, so leaving it for another time.

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: @limemloh The actual difference is actually that loading can return an error and https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.get_mut_or_try_init is unstable. Else we could just rely on the critical region `get_mut_or_try_init` creates instead of logging an error if two threads cached the value. Will add a comment.

PR #1622, opened 2026-04-15
Revise locks state to match the feature proposal.

soerenbf:


plt/plt-token-module/src/key_value_state.rs
```diff
@@ -110,6 +144,18 @@ fn account_roles_state_key(account_index: AccountIndex) -> TokenStateKey {
...
+    let mut locked_balance_key = Vec::with_capacity(
+        ACCOUNT_LOCKED_BALANCES_STATE_PREFIX.len()
+            + size_of::<AccountIndex>()
+            + size_of::<LockId>(),
+    );
+    locked_balance_key.extend_from_slice(&ACCOUNT_LOCKED_BALANCES_STATE_PREFIX);
+    account_index.serial(&mut locked_balance_key);
+    lock_id.serial(&mut locked_balance_key);
+    locked_balance_key
+}
```
td202: This should be under the account state, not with a separate index.

plt/plt-token-module/src/key_value_state.rs
```diff
@@ -52,6 +59,18 @@ pub trait KernelOperationsExt: TokenKernelOperations {
...
+    fn set_account_locked_balance_state(
+        &mut self,
+        account: AccountIndex,
+        lock: LockId,
+        amount: RawTokenAmount,
+    ) {
+        self.set_token_state_value(
+            account_locked_balance_state_key(account, lock),
+            common::to_bytes(&amount).into(),
+        );
```
td202: If the amount is 0, the key should be removed (if present).

plt/plt-token-module/src/key_value_state.rs
soerenbf: It's still account state just under a specific prefix which only contains locked balances.. is that an issue? It follows the same structure as we applied to RBAC

allanbrondum (approved):

PR #1623, opened 2026-04-17
Merge plt-token-module and plt-scheduler-interface into plt-scheduler

limemloh:
## Purpose

Part of COR-2347 ([link](https://linear.app/concordium/issue/COR-2347/remove-the-token-kernel-abstraction))

## Changes

- Merge crates `plt-token-module` and `plt-scheduler-interface` into `plt-scheduler`. 
- Remove `TransactionExecution` trait, since there are only one implementation.


plt/plt-scheduler/src/scheduler.rs
```diff
@@ -1,61 +1,16 @@
...
 use plt_block_state::block_state_interface::BlockStateOperations;
-use plt_scheduler_interface::transaction_execution_interface::{
-    OutOfEnergyError, TransactionExecution,
-};
 use plt_scheduler_types::types::execution::{ChainUpdateOutcome, TransactionExecutionSummary};
 
 mod plt_scheduler;
 
-/// Tracks the energy remaining and some context during the execution.
-struct TransactionExecutionImpl<Account> {
```
limemloh, line 17: This moved to a new file `transaction_execution.rs`

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -3,6 +3,9 @@
 
 use crate::scheduler::{ChainUpdateExecutionError, TransactionExecutionError};
 use crate::token_kernel::TokenKernelOperationsImpl;
+use crate::token_module::module::{TokenInitializationError, TokenUpdateError};
+use crate::token_module::{TOKEN_MODULE_REF, module};
```
soerenbf: does the formatter/clippy not prevent "duplicate" use definitions? 🤔 seems like there are a lot of these in the changes proposed here

plt/plt-scheduler/src/token_module/module/mod.rs
soerenbf, line 1: Any reason why the functionality exposed here is not simply all available at `crate::token_module` instead of some of it `crate::token_module::module`? as far as i can identify, the distinction seems to be that `module` encapsulates functions for the scheduler to use 🤷 I would suggest we move it or find a more descriptive name (e.g. `token_module::exec`?)

allanbrondum (approved):

soerenbf (approved):
Looks good in general. Only have a few cosmetic concerns! 😃

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
limemloh: No, not since we removed the `rustfmt.toml` from all repos

td202 (approved):

plt/plt-scheduler/src/token_module/module/mod.rs
limemloh: Yes, I noticed that as well, but did not act on it.
Now I flatten the module, so these functions are available at `crate::token_module`

PR #1624, opened 2026-04-17
Feature/rust block state

allanbrondum:
## Purpose

Rust block state feature branch

## Changes

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/blob_store.rs
```diff
@@ -1,45 +1,210 @@
...
+    fn store_to_buffer(&self, buffer: impl Buffer, storer: &mut impl BlobStoreStore) {
+        (**self).store_to_buffer(buffer, storer)
+    }
+}
+
+impl<T: Storable> Storable for &mut T {
+    fn store_to_buffer(&self, buffer: impl Buffer, storer: &mut impl BlobStoreStore) {
+        (**self).store_to_buffer(buffer, storer)
+    }
+}
```
limemloh, line 74: ~~I am surprised that we even need these~~
If we apply [this change](https://github.com/Concordium/concordium-node/pull/1624/changes#r3122295300) then we won't need these

plt/plt-block-state/src/block_state/blob_store.rs
```diff
@@ -1,45 +1,210 @@
...
+    Ok(value)
+}
+
+/// Store the value in the blob store, and return the location of the stored value.
+/// Notice that when storing the value, it will recursively store all values pointed to by
+/// [blob references](super::blob_reference) it may be composed of, if these values are not already
+/// stored in the blob store.
+pub fn store_to_store(
+    storer: &mut impl BlobStoreStore,
+    storable: impl Storable,
```
limemloh, line 123:
```suggestion
    storable: &impl Storable,
```

plt/plt-block-state/src/block_state.rs
```diff
@@ -150,110 +146,126 @@ impl PltBlockState {
...
+            token_key_value_state
+                .insert_value(&self.blob_store_load, &key.0, value.0)
+                .unwrap();
         } else {
-            token_key_value_state.state.remove(key);
+            token_key_value_state
+                .delete_value(&self.blob_store_load, &key.0)
+                .unwrap();
         }
     }
```
limemloh, line 270: Maybe for another PR, but these are just mappings to the trie interface. Let's ditch the opaque type for the trie and let the module operate directly on this.
The block state would then only have getter and setter for the trie and we don't need the block state everywhere we do any operations on the trie like in `token_module/key_value_state.rs`.

limemloh (approved):

plt/plt-block-state/src/block_state/blob_store.rs
allanbrondum: I think it is pretty common for serilization etc to have these blanket implementations. It makes things a bit more ergonomic.

plt/plt-block-state/src/block_state.rs
allanbrondum: Yeah, generally I don't like this trie. I think how we solve it, depends on which direction we want to go long term.

allanbrondum: But it will have to be a different PR

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,545 @@
...
+                // When OnceLock::get_mut_or_try_init is stable, we can use that instead of the
+                // get/set pattern used here. get_mut_or_try_init imposes a critical region such
+                // that two threads will not load the value for caching. Right now we just log
+                // if it happens.
+                match value_lock.get() {
+                    None => {
+                        let value: V = blob_store::load_from_store(loader, *blob_location)?;
+                        value_lock
+                            .set(value)
+                            .inspect_err(|_| eprintln!("HashedCacheableRef: Value loaded by two threads at the same time"))
```
td202, line 176: We shouldn't be using stderr for logging. If we want to log, we should log properly. However, I think maybe we should just not as this isn't really a problem.

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,545 @@
...
+                            .ok();
+                        value_lock.get().expect("HashedCacheableRefRepr::get_or_cache_value: value not present though just set in lock")
```
td202, line 178: Why not:
```suggestion
                        value_lock.get_or_init(|| value)
```

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,545 @@
...
+                }
+                Some(value) => OwnedOrBorrowed::Borrowed(value),
+            },
+            HashedCacheableRefRepr::Memory { value, .. } => OwnedOrBorrowed::Borrowed(value),
+        })
+    }
+
+    /// Cache the referenced value and return it. If the value is already in memory, a reference
+    /// to it is simply returned. If it is not in memory, it is loaded from the blob store and
+    /// cached in [`HashedCacheableRefRepr::Cache`] first.
```
td202, line 157: This description seems outdated as `HashedCacheableRefRepr::Cache` doesn't exist.

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
```diff
@@ -0,0 +1,545 @@
...
+            None => {
+                let value = self.inner.repr.get_or_load_value(loader)?;
+                let hash = value.hash(loader)?;
+                self.inner
+                    .hash
+                    .set(hash)
+                    .inspect_err(|_| {
+                        eprintln!(
+                            "HashedCacheableRef: Hash calculated by two threads at the same time"
+                        )
```
td202, line 261: Again, I don't think we should be logging like this.

plt/plt-block-state/src/block_state/state/protocol_level_tokens.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    pub fn plt_list(
+        &self,
+        loader: &impl BlobStoreLoad,
+    ) -> impl ExactSizeIterator<Item = BlockStateResult<TokenId>> {
+        self.tokens.values(loader, |_token_index, token| {
+            Ok(match token.configuration.value(loader)? {
+                OwnedOrBorrowed::Owned(v) => v.0.token_id,
+                OwnedOrBorrowed::Borrowed(r) => r.0.token_id.clone(),
+            })
```
td202, line 50:
```suggestion
            Ok(token.configuration.value(loader)?.into_owned())
```

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: Just wanted to log any race conditions that should not happen. But it is not big deal, it is not so important if we log or not.

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: I don't think it is so important if we log or not. I'm just logging it because it should not happen. And using stderr until we setup proper logging in the Rust scheduler.

plt/plt-block-state/src/block_state/state/protocol_level_tokens.rs
allanbrondum: There is a projection on `token_id`, hence it is written like this to avoid cloning other fields than `token_id`

plt/plt-block-state/src/block_state/blob_reference/hashed_cacheable_reference.rs
allanbrondum: Eventually no, but for now, this is a simple solution.

PR #1627, opened 2026-04-19
Ar/psr 11 oncelock remove read closure from lfmb tree 2

allanbrondum:
## Purpose

Using the `OnceLock` in `HashedCacheableRef` mean that we can also borrow values when looking them up in the `LfmbTree` and get rid of the read closure.

## Changes

Removed read closure from

* `LfmbTree::lookup_value` 
* `LfmbTree::values`

They both use `LfmbTree::with_referenced_values` in their implementation. Notice that there is an internally enforced invariant described on `with_referenced_values`, that describes, how we can transform `OwnedOrBorrowed<'_, Subtree>` to `OwnedOrBorrowed` leaf value and branch subtrees, without breaking lifetimes.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-block-state/src/block_state/lfmb_tree.rs
+++ b/plt/plt-block-state/src/block_state/lfmb_tree.rs
@@ -28,7 +28,7 @@ use std::{iter, vec};
 /// first.
 ///
 /// The operations supported for creating new trees are:
-/// * Create empty tree with [`LfmbTree::empty`]: Returns an new empty tree.
+/// * Create empty tree with [`LfmbTree::empty`]: Returns a new empty tree.
 /// * Insert new value with [`LfmbTree::insert_value`]: Inserts a new value, assigning the sequentially next unused key,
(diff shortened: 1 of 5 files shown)


plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -153,12 +153,21 @@ use std::{iter, vec};
...
 ///    #0  #1  #2  #3  #4  #5  #6  #7
 /// ```
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct LfmbTree<K, V> {
     inner: LfmbTreeInner<V>,
     _key_type: PhantomData<K>,
 }
 
+impl<K, V> Clone for LfmbTree<K, V> {
```
allanbrondum, line 162: Rewrote to not require `K` and `V` to be `Clone`

PR #1628, opened 2026-04-20
Migration of block state values

allanbrondum:
## Purpose

Protocol migration of the Rust block state.

## Changes

* trait `block_state::migration::Migrate` that defines and describes how to migrate a block state value
* implementation of the trait on all the block state types

We may need to add protocol version or migration target type (`type Migrate::ToType`) at some point, but keeping it as simple as possible for now.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/migration.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Trait implemented by block state values to support migration when protocol version increments.
+/// The migration must,
+///
+/// * recursively store all [blob references](super::blob_reference) into the new blob store
+///   of we migrate to (the blob store of the new protocol version)
+/// * apply any changes to the representation of the block state value (data model migration)
+///
+/// Since each protocol version has its own blob store, migration is always needed at
+/// protocol update, even if the block state value has no data model changes.
+pub trait Migrate {
```
td202, line 18: I am not sure that it's a good idea to have a universal `Migrate` trait like this. At the moment, for the Rust code it's mostly just copying things for P10->P11 migration, so it works. But in general, migrations can alter data, possibly using context provided as part of the protocol update, or even doing more complex things, including changing types. I would rather that types have their own migration functions that are not abstracted by a trait, and can, for instance, be specific to the version migration taking place. For instance `migrate_trivial` for where the data and representation is unchanged, or `migrate_p10_to_p11` that involves changes that are specific to that protocol update.

allanbrondum: Right now I think having the trait is easier, because it is used as trait bound in migration code for `HashedCacheableRef` and `LfmbTree` (which would otherwise take a closure). We can remove the trait again if it because a problem (or modify the trait). Until we have the block state model in place, I would rather leave the path open and see what works.

PR #1636, opened 2026-04-23
Ar/psr 11 oncelock remove read closure from lfmb tree 3

allanbrondum:
## Purpose

Using the `OnceLock` in `HashedCacheableRef` mean that we can also borrow values (or return owned) when looking them up in the `LfmbTree` and get rid of the read closure.

## Changes

Removed read closure from

* `LfmbTree::lookup_value` 
* `LfmbTree::values` (implementation at lifetime bounds are also simpler and more natural now)

They both use `OwnedOrBorrowed<HashedCacheableRef>::bind_value` in their implementation. Notice that there is an internally enforced invariant described on `lookup_value` and `bind_value`, that describes why/how loading referenced values in the monadic structure of `OwnedOrBorrowed` works.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state/lfmb_tree.rs
```diff
@@ -153,12 +153,21 @@ use std::{iter, vec};
...
 ///    #0  #1  #2  #3  #4  #5  #6  #7
 /// ```
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct LfmbTree<K, V> {
     inner: LfmbTreeInner<V>,
     _key_type: PhantomData<K>,
 }
 
+impl<K, V> Clone for LfmbTree<K, V> {
```
allanbrondum, line 162: Implement manually such that we don't require K and V to be Clone

limemloh: Ideally this would be a comment in the code as well

limemloh (approved):

PR #1637, opened 2026-04-24
Add lock queries to grpc servivce

soerenbf:
Closes COR-2309

Dependent on https://github.com/Concordium/concordium-base/pull/927

## Purpose

Adds grpc service queries for lock list and lock info for a given lock id.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
@@ -240,7 +240,7 @@ getLockInfoV2 ::
     -- | Block hash.
     Ptr Word8 ->
     -- | Lock ID (24 bytes: three big-endian Word64 fields).
-    Ptr Word8 ->
+    Ptr SerializedLockId ->
     -- | Out pointer for writing the block hash that was used.
     Ptr Word8 ->
     Ptr ReceiverVec ->
@@ -251,7 +251,7 @@ getLockInfoV2 cptr blockType blockHashPtr lockIdPtr outHash outVec copierCbk = d
     Ext.ConsensusRunner mvr <- deRefStablePtr cptr
     let copier = callCopyToVecCallback copierCbk
     bhi <- decodeBlockHashInput blockType blockHashPtr
-    lockId <- peek (castPtr lockIdPtr) :: IO SerializedLockId
+    lockId <- peek (castPtr lockIdPtr)
     res <- runMVR (Q.getLockInfo bhi lockId) mvr
     case res of
         Q.BQRBlock _ (Left QLEUnknownLock) -> do
@@ -1419,7 +1419,7 @@ foreign export ccall
         -- | Block hash.
         Ptr Word8 ->
         -- | Lock ID (24 bytes: three big-endian Word64 fields).
-        Ptr Word8 ->
+        Ptr SerializedLockId ->
         -- | Out pointer for writing the block hash that was used.
         Ptr Word8 ->
         Ptr ReceiverVec ->
--- a/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
@@ -20,7 +20,6 @@ import Data.Functor
 import Data.Int
 import qualified Data.ProtoLens as Proto
 import qualified Data.ProtoLens.Combinators as Proto
-import qualified Data.Serialize as S
 import qualified Data.Vector as Vec
 import Data.Word
 import Foreign
@@ -43,9 +42,12 @@ import Concordium.Crypto.SHA256 (Hash (Hash))
 import Concordium.External.Helpers
 import Concordium.GlobalState.Parameters (CryptographicParameters)
 import Concordium.ID.Parameters (withGlobalContext)
-import Concordium.Scheduler.ProtocolLevelTokens.Queries (QueryLockError (..), QueryTokenModuleError (..))
+import Concordium.Scheduler.ProtocolLevelTokens.Queries (
+    QueryLockError (..),
+    QueryTokenModuleError (..),
+    SerializedLockId,
+ )
 import qualified Concordium.Types.InvokeContract as InvokeContract
-import qualified Concordium.Types.Locks as Locks
 import qualified Concordium.Wasm as Wasm
 
 -- | An opaque representation of a Rust vector. This is used by callbacks to copy
@@ -230,15 +232,6 @@ getLockListV2 ::
     IO Int64
 getLockListV2 = blockStreamHelper Q.getLockList
 
--- | Decode a 'Locks.LockId' from a foreign pointer. Assumes 24 bytes are available
--- (three big-endian Word64 fields, matching the @Serialize LockId@ instance).
-decodeLockId :: Ptr Word8 -> IO Locks.LockId
-decodeLockId lockIdPtr = do
-    bs <- BS.unsafePackCStringLen (castPtr lockIdPtr, 24)
-    case S.decode bs of
-        Left err -> error $ "Precondition violation in FFI call (decodeLockId): " ++ err
-        Right lockId -> return lockId
-
 -- | Foreign-exported FFI entry point for the unary `GetLockInfo` gRPC v2 endpoint.
 getLockInfoV2 ::
     StablePtr Ext.ConsensusRunner ->
@@ -258,7 +251,7 @@ getLockInfoV2 cptr blockType blockHashPtr lockIdPtr outHash outVec copierCbk = d
     Ext.ConsensusRunner mvr <- deRefStablePtr cptr
     let copier = callCopyToVecCallback copierCbk
     bhi <- decodeBlockHashInput blockType blockHashPtr
-    lockId <- decodeLockId lockIdPtr
+    lockId <- peek (castPtr lockIdPtr) :: IO SerializedLockId
     res <- runMVR (Q.getLockInfo bhi lockId) mvr
     case res of
         Q.BQRBlock _ (Left QLEUnknownLock) -> do
--- a/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
@@ -43,7 +43,7 @@ import Concordium.Crypto.SHA256 (Hash (Hash))
 import Concordium.External.Helpers
 import Concordium.GlobalState.Parameters (CryptographicParameters)
 import Concordium.ID.Parameters (withGlobalContext)
-import Concordium.Scheduler.ProtocolLevelTokens.Queries (QueryLockModuleError (..), QueryTokenModuleError (..))
+import Concordium.Scheduler.ProtocolLevelTokens.Queries (QueryLockError (..), QueryTokenModuleError (..))
 import qualified Concordium.Types.InvokeContract as InvokeContract
 import qualified Concordium.Types.Locks as Locks
 import qualified Concordium.Wasm as Wasm
@@ -261,16 +261,9 @@ getLockInfoV2 cptr blockType blockHashPtr lockIdPtr outHash outVec copierCbk = d
     lockId <- decodeLockId lockIdPtr
     res <- runMVR (Q.getLockInfo bhi lockId) mvr
     case res of
-        Q.BQRBlock _ (Left QLMEUnknownLock) -> do
+        Q.BQRBlock _ (Left QLEUnknownLock) -> do
             copyHashTo outHash res
             return $ queryResultCode QRNotFound
-        Q.BQRBlock _ (Left e@QLMEInternal{}) -> do
-            mvLog mvr Logger.External Logger.LLError $
-                "Internal error processing GetLockInfo: " ++ show e
-            return $ queryResultCode QRInternalError
-        Q.BQRBlock _ (Left QLMEUnavailable) -> do
-            copyHashTo outHash res
-            return $ queryResultCode QRUnavailable
         Q.BQRBlock _ (Right r) ->
             returnMessageWithBlock (copier outVec) outHash (res $> r)
         Q.BQRNoBlock ->
--- a/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
+++ b/concordium-consensus/src-lib/Concordium/External/GRPC2.hs
@@ -20,6 +20,7 @@ import Data.Functor
 import Data.Int
 import qualified Data.ProtoLens as Proto
 import qualified Data.ProtoLens.Combinators as Proto
+import qualified Data.Serialize as S
 import qualified Data.Vector as Vec
 import Data.Word
 import Foreign
@@ -42,8 +43,9 @@ import Concordium.Crypto.SHA256 (Hash (Hash))
 import Concordium.External.Helpers
 import Concordium.GlobalState.Parameters (CryptographicParameters)
 import Concordium.ID.Parameters (withGlobalContext)
-import Concordium.Scheduler.ProtocolLevelTokens.Queries (QueryTokenModuleError (..))
+import Concordium.Scheduler.ProtocolLevelTokens.Queries (QueryLockModuleError (..), QueryTokenModuleError (..))
 import qualified Concordium.Types.InvokeContract as InvokeContract
+import qualified Concordium.Types.Locks as Locks
 import qualified Concordium.Wasm as Wasm
 
 -- | An opaque representation of a Rust vector. This is used by callbacks to copy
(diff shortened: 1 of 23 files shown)


plt/plt-scheduler/src/queries.rs
```diff
@@ -123,3 +131,125 @@ pub fn query_token_authorizations(
         details,
     })
 }
+
+/// Get the [`LockId`]s of all protocol-level locks registered on the chain at the
+/// end of the block.
+pub fn query_lock_list(block_state: &impl BlockStateQuery) -> Vec<LockId> {
+    block_state.lock_list().collect()
+}
```
soerenbf, line 138: Not sure if this is the correct solution here, maybe we need to implement some way to stream the lock ids in batches for scalability.. alternatively, we can leave it as and handle scaling issues if/when they surface 🤷

plt/plt-block-state/src/block_state.rs
```diff
@@ -425,6 +425,21 @@ impl<IntState: HasBlockState, Load: BlobStoreLoad, ExtState: ExternalBlockStateQ
...
+        // returned iterator preserves the BTreeMap ordering and reports the exact
+        // number of locks via `ExactSizeIterator`.
+        self.internal_block_state
+            .block_state()
+            .locks
+            .locks
+            .0
+            .keys()
+            .cloned()
+            .collect::<Vec<LockId>>()
```
limemloh: Do we really need to collect here? if so, then maybe just return the `Vec` directly and change the signature, because we collect it once again when calling it in `plt-scheduler/queries.rs`

plt/plt-block-state/src/block_state.rs
```diff
@@ -529,18 +544,45 @@ impl<Load: BlobStoreLoad, ExtState: ExternalBlockStateOperations> BlockStateOper
...
+                    lock_id.clone(),
+                    Lock {
+                        locked_balances: Default::default(),
+                        configuration: configuration.clone(),
+                    },
+                );
+                debug_assert!(prev.is_none(), "create_lock called for an existing lock id");
+                Ok(state)
+            })
+            .unwrap();
```
limemloh, line 556: Long term we want to do something about these unwraps here, so maybe link to the same issue as the rest of the methods on the blockstate
```
// todo propagate block state error as part of https://linear.app/concordium/issue/COR-2346/push-blockstateerror-to-scheduler-code
```

plt/plt-scheduler/src/queries.rs
limemloh: I don't really see a clear use case for this query other than debugging.
Maybe just create a task to figure this out and leave it for now.

limemloh (approved):

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
```diff
@@ -213,6 +215,60 @@ getTokenAuthorizationsV2 cptr blockType blockHashPtr tokenIdPtr tokenIdLen outHa
...
+getLockListV2 = blockStreamHelper Q.getLockList
+
+-- | Decode a 'Locks.LockId' from a foreign pointer. Assumes 24 bytes are available
+-- (three big-endian Word64 fields, matching the @Serialize LockId@ instance).
+decodeLockId :: Ptr Word8 -> IO Locks.LockId
+decodeLockId lockIdPtr = do
+    bs <- BS.unsafePackCStringLen (castPtr lockIdPtr, 24)
+    case S.decode bs of
+        Left err -> error $ "Precondition violation in FFI call (decodeLockId): " ++ err
+        Right lockId -> return lockId
```
td202: I'm not saying we should do something different, but there's something a little bit perverse about this. We get the serialized LockId from Rust, deserialize it, serialize it and pass it to Rust.

I think there's also something potentially quite dangerous about the pattern here (though I know we use it in a few other places). In particular, can we guarantee that there won't be a use-after-free problem with `lockIdPtr`. Specifically `lockId` (and also `err`) could be a thunk that retains a reference to `lockIdPtr`, and if they are evaluated after we've lost ownership of the `lockIdPtr` then we could have a use-after-free problem. Consider the following:

```haskell
getWord64be' :: Get Word64
getWord64be' = do
    s <- getBytes 8
    return $
        (fromIntegral (s `BS.unsafeIndex` 0) `shift` 56)
            .|. (fromIntegral (s `BS.unsafeIndex` 1) `shift` 48)
            .|. (fromIntegral (s `BS.unsafeIndex` 2) `shift` 40)
            .|. (fromIntegral (s `BS.unsafeIndex` 3) `shift` 32)
            .|. (fromIntegral (s `BS.unsafeIndex` 4) `shift` 24)
            .|. (fromIntegral (s `BS.unsafeIndex` 5) `shift` 16)
            .|. (fromIntegral (s `BS.unsafeIndex` 6) `shift` 8)
            .|. (fromIntegral (s `BS.unsafeIndex` 7))

getLockId' :: Get LockId
getLockId' = do
    liAccountIndex <- getWord64be'
    liSequenceNumber <- getWord64be'
    liCreationOrder <- getWord64be'
    return LockId{..}

insaneTest :: Assertion
insaneTest = allocaBytes arrSize $ \ptr -> do
    mapM_ (\o -> pokeByteOff ptr o (0 :: Word8)) [0 .. arrSize - 1]
    bs <- BS.unsafePackCStringLen (castPtr ptr, arrSize)
    assertEqual "Initial bs" bs "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
    lockId <- case runGet getLockId' bs of
        Left err -> error $ "Failed to decode " ++ err
        Right lockId -> return $ lockId
    mapM_ (\o -> pokeByteOff ptr o (1 :: Word8)) [0 .. arrSize - 1]
    assertEqual "Updated bs" bs "\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01"
    assertEqual "LockId" lockId (LockId 0 0 0)
  where
    arrSize = 24
```

If you run this code, the last assertion fails because the `lockId` is a thunk that references the byte string until it is actually evaluated at the `assertEqual`, by which point the string has been changed. However, there is a difference to the actual implementation that saves us in practice, which is that `getWord64be` has `return $! ...`, which forces evaluation.

I'm not sure what I'm actually recommending here, but one thing could be that we don't actually deserialize the `LockId`, but use for instance a `FixedByteString` to just keep the serialized LockId.

plt/plt-scheduler/src/locks/lock_controller.rs
```diff
@@ -1,12 +1,14 @@
 //! Runtime interface for protocol-level lock controllers.
 
 use concordium_base::{
-    protocol_level_locks::LockId,
+    protocol_level_locks::{LockController as CborLockController, LockId},
```
td202: Please no! I'd rather we use the qualified name than renaming like this.

soerenbf: Sure, I'll change it 😁

plt/plt-block-state/src/block_state.rs
soerenbf: Sure, In practice there is no error case for this and the following function though. The `update_block_state_` fn returns a result, so I guess it's best to just align all functions to this long-term 🙂

plt/plt-scheduler/src/queries.rs
soerenbf: Well.. it's part of the GRPC service API specification but I agree that it's a weird use case to iterate all locks. @td202 what's your take on this?

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
soerenbf: Sure, we're not doing anything but crossing another FFI boundary so we can do that.. It makes the lock query signature harder to reason about and different from the rest of the functions in that module, which is why I chose this implementation in the first place. But I guess this is also new territory in the sense that there is no haskell implementation of this that does _not_ rely on the rust plt scheduler.

If we want to change this we might as well take it to the next level and just pass the pointer and not do any form of interpretation in the haskell layer, as we're not doing anything with the bytes but creating a new pointer? 🤷

plt/plt-scheduler/src/queries.rs
soerenbf: We keep it for now as it's useful for diagnostics. We might need to optimize it at some point with a proper streaming implementation. Will leave a note in the code.

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
soerenbf: Discussed with @td202 and we agreed to have the haskell layer keep it's own reference to a serialized lock id and treat that as opaque, leaving the rust layer to only interpret the bytes.

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
```diff
@@ -213,6 +217,51 @@ getTokenAuthorizationsV2 cptr blockType blockHashPtr tokenIdPtr tokenIdLen outHa
...
+    Ptr Word8 ->
+    Ptr ReceiverVec ->
+    -- | Callback to output data.
+    FunPtr CopyToVecCallback ->
+    IO Int64
+getLockInfoV2 cptr blockType blockHashPtr lockIdPtr outHash outVec copierCbk = do
+    Ext.ConsensusRunner mvr <- deRefStablePtr cptr
+    let copier = callCopyToVecCallback copierCbk
+    bhi <- decodeBlockHashInput blockType blockHashPtr
+    lockId <- peek (castPtr lockIdPtr) :: IO SerializedLockId
```
td202: No need to cast.

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
```diff
@@ -213,6 +217,51 @@ getTokenAuthorizationsV2 cptr blockType blockHashPtr tokenIdPtr tokenIdLen outHa
...
+
+-- | Foreign-exported FFI entry point for the unary `GetLockInfo` gRPC v2 endpoint.
+getLockInfoV2 ::
+    StablePtr Ext.ConsensusRunner ->
+    -- | Block type.
+    Word8 ->
+    -- | Block hash.
+    Ptr Word8 ->
+    -- | Lock ID (24 bytes: three big-endian Word64 fields).
+    Ptr Word8 ->
```
td202:
```suggestion
    Ptr SerializedLockId ->
```

concordium-consensus/src-lib/Concordium/External/GRPC2.hs
```diff
@@ -1362,6 +1411,35 @@ foreign export ccall
...
 
+foreign export ccall
+    getLockInfoV2 ::
+        StablePtr Ext.ConsensusRunner ->
+        -- | Block type.
+        Word8 ->
+        -- | Block hash.
+        Ptr Word8 ->
+        -- | Lock ID (24 bytes: three big-endian Word64 fields).
+        Ptr Word8 ->
```
td202:
```suggestion
        Ptr SerializedLockId ->
```

concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
```diff
@@ -520,3 +527,196 @@ unliftBlockStateQueryCallbacks bs = BS.withUnliftBSQ $ \unlift ->
...
+        FFI.Ptr FFI.CSize ->
+        IO Word.Word8
+
+lockIdSize :: Int
+lockIdSize = 24
+
+data LockIdSize
+
+instance FBS.FixedLength LockIdSize where
+    fixedLength _ = lockIdSize
```
td202:
```suggestion
-- | Placeholder type representing the size of a serialized 'LockId' (24 bytes).
data LockIdSize

instance FBS.FixedLength LockIdSize where
    fixedLength _ = 24
```

concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
```diff
@@ -520,3 +527,196 @@ unliftBlockStateQueryCallbacks bs = BS.withUnliftBSQ $ \unlift ->
...
+
+-- | A serialized lock id as passed across the FFI boundary.
```
td202:
```suggestion
-- | A serialized Lock ID as passed across the FFI boundary.
```

concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
```diff
@@ -520,3 +527,196 @@ unliftBlockStateQueryCallbacks bs = BS.withUnliftBSQ $ \unlift ->
...
+newtype SerializedLockId = SerializedLockId (FBS.FixedByteString LockIdSize)
+    deriving (FFI.Storable)
+
+-- | Get the CBOR-encoded `lock-info` payload for a given lock id, for a protocol version where the
```
td202:
```suggestion
-- | Get the CBOR-encoded `lock-info` payload for a given Lock ID, for a protocol version where the
```

concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
```diff
@@ -520,3 +527,196 @@ unliftBlockStateQueryCallbacks bs = BS.withUnliftBSQ $ \unlift ->
...
+                        -- after this call returns. The CBOR bytes are stored opaquely in
+                        -- 'LockQueries.LockInfo'.
+                        returnData <-
+                            BS.unsafePackCStringFinalizer
+                                returnDataPtr
+                                (fromIntegral returnDataLen)
+                                (Memory.rs_free_array_len_2 returnDataPtr (fromIntegral returnDataLen))
+                        case Status.parseStatusCode statusCode of
+                            Just Status.FSCSuccess ->
+                                return $ Just $ LockQueries.LockInfo (BS.copy returnData)
```
td202: If you want to make a copy, then it would be better to use `packCStringLen` instead:

```suggestion
                        -- Copy the FFI output buffer into a strict 'ByteString' that owns its
                        -- storage, so the buffer can be released by the finalizer immediately
                        -- after this call returns. The CBOR bytes are stored opaquely in
                        -- 'LockQueries.LockInfo'.
                        returnData <-
                            BS.packCStringLen
                                (returnDataPtr, (fromIntegral returnDataLen))
                        Memory.rs_free_array_len_2 returnDataPtr (fromIntegral returnDataLen)
                        case Status.parseStatusCode statusCode of
                            Just Status.FSCSuccess ->
                                return $ Just $ LockQueries.LockInfo returnData
```
Admittedly, this does mean copying also in the panic path - though that's probably only an issue if we're out of memory.

Alternatively, we can just not do the copy and keep the Rust bytestring around as needed:

```suggestion
                        returnData <-
                            BS.unsafePackCStringFinalizer
                                returnDataPtr
                                (fromIntegral returnDataLen)
                                (Memory.rs_free_array_len_2 returnDataPtr (fromIntegral returnDataLen))
                        case Status.parseStatusCode statusCode of
                            Just Status.FSCSuccess ->
                                return $ Just $ LockQueries.LockInfo returnData
```

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -584,6 +584,51 @@ extern "C" {
         copier: CopyToVecCallback,
     ) -> i64;
 
+    /// Stream the list of all PLT lock ids that exist in the given block.
```
td202:
```suggestion
    /// Stream the list of all PLT Lock IDs that exist in the given block.
```

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -584,6 +584,51 @@ extern "C" {
...
+    ///
+    /// Individual lock ids are enqueued using the provided callback.
```
td202:
```suggestion
    /// Individual Lock IDs are enqueued using the provided callback.
```

plt/plt-scheduler/src/ffi/queries.rs
```diff
@@ -359,3 +360,168 @@ extern "C" fn ffi_query_token_account_infos(
...
+/// - Argument `return_data_len_out` must be a non-null and valid pointer for writing.
+#[unsafe(no_mangle)]
+extern "C" fn ffi_query_lock_info(
+    load_callback: LoadCallback,
+    read_token_account_balance_callback: ReadTokenAccountBalanceCallback,
+    get_account_index_by_address_callback: GetAccountIndexByAddressCallback,
+    get_account_address_by_index_callback: GetCanonicalAddressByAccountIndexCallback,
+    get_token_account_states_callback: GetTokenAccountStatesCallback,
+    block_state: *const BlockState,
+    lock_id: *const u8,
```
td202: Suggestion: I think we can refine the type as
```suggestion
    lock_id: *const [u8; 24],
```

plt/plt-scheduler/src/ffi/queries.rs
```diff
@@ -359,3 +360,168 @@ extern "C" fn ffi_query_token_account_infos(
...
+            get_token_account_states_ptr: get_token_account_states_callback,
+        };
+        let internal_block_state = unsafe { &*block_state };
+        let block_state = ExecutionTimeBlockState {
+            internal_block_state,
+            blob_store_load: load_callback,
+            external_block_state: external_callbacks,
+        };
+        // The Haskell side serializes a `LockId` as three big-endian `u64`s, exactly 24 bytes.
+        let lock_id_bytes = unsafe { std::slice::from_raw_parts(lock_id, 24) };
```
td202: With the change of type:
```suggestion
        let lock_id_bytes = unsafe { lock_id.as_ref_unchecked() };
```

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -584,6 +584,51 @@ extern "C" {
...
+    /// * `lock_id` - Pointer to 24 bytes containing the serialized `LockId` (three
+    ///   big-endian `u64` fields: `account_index`, `sequence_number`, `creation_order`).
+    /// * `out_hash` - Location to write the block hash used in the query.
+    /// * `out` - Location to write the proto-encoded `plt.LockInfo` bytes.
+    /// * `copier` - Callback for writing the output.
+    pub fn getLockInfoV2(
+        consensus: *mut consensus_runner,
+        block_id_type: u8,
+        block_id: *const u8,
+        lock_id: *const u8,
```
td202:
```suggestion
        lock_id: *const [u8; 24],
```

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -2648,6 +2693,91 @@ impl ConsensusContainer {
...
+        let bhi = crate::grpc2::types::block_hash_input_to_ffi(block_hash).require()?;
+        let (block_id_type, block_hash) = bhi.to_ptr();
+
+        // Serialize the LockId as three big-endian u64 fields, exactly 24 bytes.
+        // This matches the Haskell `Serialize LockId` instance and the FFI input contract
+        // documented on `getLockInfoV2`.
+        let mut lock_id_bytes = [0u8; 24];
+        lock_id_bytes[0..8].copy_from_slice(&lock_id.account_index.to_be_bytes());
+        lock_id_bytes[8..16].copy_from_slice(&lock_id.sequence_number.to_be_bytes());
+        lock_id_bytes[16..24].copy_from_slice(&lock_id.creation_order.to_be_bytes());
```
td202: We should probably use the serialization defined by `LockId` rather than having an ad hoc reimplementation here:

```suggestion
        let lock_id : [u8; 24] = concordium_base::common::to_bytes(&lock_id).try_into()?;
```

concordium-node/src/consensus_ffi/ffi.rs
```diff
@@ -2648,6 +2693,91 @@ impl ConsensusContainer {
...
+        let consensus = self.consensus.load(Ordering::SeqCst);
+        let mut out_data: Vec<u8> = Vec::new();
+        let mut out_hash = [0u8; 32];
+
+        let response: ConsensusQueryResponse = unsafe {
+            getLockInfoV2(
+                consensus,
+                block_id_type,
+                block_hash.as_ptr(),
+                lock_id_bytes.as_ptr(),
```
td202: If the type is changed:
```suggestion
                &lock_id_bytes,
```

concordium-node/src/consensus_ffi/ffi.rs
td202: Actually, probably need to convert the GRPC `LockId` type to the internal `LockId` type.

td202 (approved):

PR #1638, opened 2026-04-27
Add locks data to account cbor

soerenbf:
Closes COR-2316

Depends on https://github.com/Concordium/concordium-base/pull/932

## Purpose

Extend the account info query to include locked balances

## Checklist

      hard-to-understand areas.


plt/plt-scheduler/src/token_module/module/queries.rs
```diff
@@ -68,15 +69,19 @@ fn query_token_module_state_impl<BSQ: BlockStateQuery>(
 pub fn query_token_module_account_state<BSQ: BlockStateQuery>(
     context: &TokenQueryContext<'_, BSQ>,
     account: AccountIndex,
-) -> RawCbor {
-    let state = query_token_module_account_state_impl(context, account);
-    RawCbor::from(cbor::cbor_encode(&state))
+    balance: RawTokenAmount,
+    decimals: u8,
```
td202, line 86: Suggestion: document the arguments

plt/plt-scheduler/tests/utils/block_state_external_stubbed.rs
```diff
@@ -83,6 +83,8 @@ pub struct Account<Token> {
     index: AccountIndex,
     /// The canonical account address of the account.
     address: AccountAddress,
+    /// The next transaction sequence number to use for test transactions sent by this account.
+    next_transaction_sequence_number: Nonce,
```
td202: I don't understand why this is needed. The sequence number on an account is not accessible. We only pass in the current sequence number from the scheduler for the purposes of ensuring that lock IDs are assigned appropriately, meaning that for most transactions the sequence number used is not relevant.

Also, I think this struct has been removed, and it might be undesirable to add the sequence number in the entity model.

plt/plt-scheduler/src/token_module/key_value_state.rs
```diff
@@ -461,6 +461,31 @@ pub fn get_locked_balance_for(
...
+/// account.
+pub fn get_locked_balances_for_account(
+    context: &impl ReadTokenKeyValueState,
+    account: AccountIndex,
+) -> Result<Vec<(LockId, RawTokenAmount)>, TokenStateInvariantError> {
+    let prefix = account_state_key(account, ACCOUNT_STATE_KEY_QUANTA);
+    context
+        .iter_token_state_prefix(&prefix)
+        .map(|(key, value)| {
+            let lock = common::from_bytes_complete(&key.0[prefix.0.len()..]).map_err(|err| {
```
td202: From Copilot: the slice `key.0[prefix.0.len()..]` could panic. Something like this is safer:
```suggestion
            let lock_bytes = key.0.strip_prefix::<[u8]>(prefix.0.as_ref()).ok_or_else(|| {
                TokenStateInvariantError(
                    "Iterator over account quanta state produced invalid key".to_string(),
                )
            })?;
            let lock = common::from_bytes_complete(&lock_bytes).map_err(|err| {
```

plt/plt-scheduler/tests/utils/block_state_external_stubbed.rs
soerenbf: Don't remember why I thought it was a good idea.. rolled it back.

plt/plt-scheduler/tests/utils/entity_traits/scheduler_impl.rs
```diff
@@ -114,6 +114,7 @@ impl SchedulerOperations for BlockStateP9 {
         };
 
         queries::query_token_account_infos(&exec_block_state, account)
+            .expect("query_token_account_infos must succeed")
```
limemloh: I think we should introduce an error instead of panicking here, maybe just forward the error

plt/plt-scheduler/tests/utils/entity_traits/scheduler_impl.rs
```diff
@@ -260,6 +261,7 @@ impl SchedulerOperations for BlockStateP11 {
...
```
limemloh: Same here

limemloh (approved):

plt/plt-scheduler/tests/utils/entity_traits/scheduler_impl.rs
soerenbf: sure

plt/plt-scheduler/src/token_module/key_value_state.rs
```diff
@@ -462,6 +462,46 @@ pub fn get_locked_balance_for(
...
+/// - `account`: Account whose lock-controlled balances should be returned.
+///
+/// # Errors
+///
+/// Returns an error if a stored lock identifier or locked amount cannot be
+/// decoded from token-module state.
+pub fn get_locked_balances_for_account(
+    context: &impl ReadTokenKeyValueState,
+    account: AccountIndex,
+) -> Result<Vec<(LockId, RawTokenAmount)>, TokenStateInvariantError> {
```
td202: Perhaps this could return an iterator instead of a `Vec`. Since you immediately iterate over it anyway, it's potentially wasteful to create the intermediate `Vec`.

td202 (approved):

plt/plt-scheduler/src/token_module/key_value_state.rs
soerenbf: Good call, will adapt :)

PR #1640, opened 2026-04-29
Create lock

td202:
## Purpose

Closes [COR-2302](https://linear.app/concordium/issue/COR-2302/locks-lockcreate-implementation)

## Changes

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/block_state.rs
```diff
@@ -530,8 +442,20 @@ impl<Load: BlobStoreLoad, ExtState: ExternalBlockStateOperations> BlockStateOper
     }
 
     // TODO: lock creation implemented as part of COR-2302
```
soerenbf: can remove the TODO now 🙂

plt/plt-block-state/src/block_state/types/protocol_level_locks.rs
```diff
@@ -1,18 +1,109 @@
...
+    pub fn expiry(&self) -> TransactionTime {
+        self.expiry
+    }
+
+    /// Get the lock controller configuration.
+    pub fn controller(&self) -> &LockControllerConfig {
+        &self.controller
+    }
+
+    /// Get the lock configuration as a `LockConfig` with resolved account addresses.
```
soerenbf: would it make sense to mention here that the purpose is to convert into something we can represent as CBOR?

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -221,11 +231,139 @@ pub fn execute_meta_update_transaction<BSO: BlockStateOperations>(
...
+                        controller.keep_alive,
+                        controller.memo,
+                    ) {
+                        Ok(controller) => controller,
+                        Err(err) => {
+                            return Ok(Some(err));
+                        }
+                    }
+                }
+            };
```
soerenbf: Maybe it would make sense to make this part of the `LockController` trait?
```suggestion
            let controller = config.controller.from_cbor_controller(&block_state);
```

In https://github.com/Concordium/concordium-node/pull/1637 we're adding a `to_cbor_controller` used when reading the config through a query.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -221,11 +231,139 @@ pub fn execute_meta_update_transaction<BSO: BlockStateOperations>(
...
+    block_state: &mut BSO,
+    _index: usize,
+    lock_operation: LockOperation,
+    events: &mut Vec<BlockItemEvent>,
+) -> Result<Option<TransactionRejectReason>, TransactionExecutionError> {
+    match lock_operation {
+        LockOperation::Fund(_meta_lock_fund_details) => todo!(),
+        LockOperation::Send(_meta_lock_send_details) => todo!(),
+        LockOperation::Return(_meta_lock_return_details) => todo!(),
+        LockOperation::Create(meta_lock_create_details) => {
```
soerenbf, line 260: We need to check whether the lock already exists at this point? The only way this gets checked currently is through the `assert` inside create_lock, which states that it should be checked prior to calling that function.

plt/plt-block-state/src/block_state/types/protocol_level_locks.rs
```diff
@@ -1,18 +1,109 @@
...
+                            })
+                        })
+                        .collect::<Result<_, _>>()?,
+                    tokens: controller.tokens.clone(),
+                    keep_alive: controller.keep_alive,
+                    memo: controller.memo.clone(),
+                };
+                protocol_level_locks::LockController::SimpleV0(controller)
+            }
+        };
```
soerenbf: #1637 adds this to the `LockController` trait.

soerenbf (approved):
As also mentioned in some of your inline comments/todos, there is some overlap with #1637 that we need to be aware of when merging.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
td202: The lock cannot already exist, or else we're replaying a transaction nonce.

soerenbf: oh.. right 🙈

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
td202: OK. I've refactored to make this work.

plt/plt-scheduler/src/locks/lock_controller.rs
```diff
@@ -93,11 +107,29 @@ impl LockController for LockControllerConfig {
     fn to_cbor_controller<BSQ: BlockStateQuery>(
         &self,
         bsq: &BSQ,
-    ) -> Result<concordium_base::protocol_level_locks::LockController, QueryLockError> {
+    ) -> Result<concordium_base::protocol_level_locks::LockController, AccountNotFoundByIndexError>
```
soerenbf, line 110: Might make sense to also use the associated type here? Won't really make a difference.. thought it'd be nice to convert to/from the same types 🤷
```suggestion
    ) -> Result<Self::ControllerConfig, AccountNotFoundByIndexError>
```

soerenbf (approved):
Looks good!

plt/plt-scheduler/src/locks/lock_controller.rs
td202: I'm not going to do it in this PR. You can change it if you want.

PR #1641, opened 2026-04-29
Update test block state hashes for P11.

td202:
## Purpose

Fix the tests for P11 by updating the block state hashes. The hashes are updated for P11 as the state now includes PLT locks.


soerenbf (approved):
❤️

PR #1643, opened 2026-05-07
Lock cancel

td202:
## Purpose

## Changes

Closes [COR-2304](https://linear.app/concordium/issue/COR-2304/locks-lockcancel-implementation)
Depends on https://github.com/Concordium/concordium-base/pull/937

Support the `lockCancel` meta-update operation.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -142,6 +142,15 @@ pub struct LockControllerSimpleV0 {
...
 
+impl LockControllerSimpleV0 {
+    /// Check if an account has a specified role.
+    pub fn has_role(&self, account: AccountIndex, role: LockControllerSimpleV0Capability) -> bool {
+        self.grants
+            .iter()
+            .any(|grant| grant.account == account && grant.roles.contains(&role))
+    }
+}
+
```
allanbrondum, line 153: I think we should not add too much logic in the persistence layer. In order not to mix up business logic and persistence logic. This function seems like something that would be more appropriate as a free function in the scheduler code?

plt/plt-scheduler/src/ffi/queries.rs
```diff
@@ -566,25 +566,13 @@ extern "C" fn ffi_query_lock_list(
...
-                let exec_block_state = ExecutionTimeBlockStateP10 {
-                    block_state,
-                    context,
-                };
-                queries::query_lock_list(&exec_block_state)
+            PersistentBlockState::P10(_persistent) => {
+                // Locks are not supported at P10.
+                Vec::new()
             }
             PersistentBlockState::P11(persistent) => {
```
allanbrondum, line 589: I think this logic to differentiate behaviour on protocol versions belongs elsewhere than in the FFI layer. I would prefer keeping it as is (dispatching to `queries::query_lock_list`), I intentionally wrote the code like this to also call `queries` on P9 and P10. And then we implement the variation by protocol in the function(s) that are called in `queries` (when we introduce the protocol specific block state types in the scheduler lock code). That is how I've implemented it for tokens. And that also makes the varying behaviour per protocol testable.

plt/plt-block-state/src/block_state_interface.rs
```diff
@@ -90,6 +96,11 @@ pub trait BlockStateQuery {
     /// The token is guaranteed to exist on chain, when holding an instance of this type.
     type Token;
 
+    /// Opaque type that represents a protocol-level lock on chain.
+    /// An instance of this type *does not* guarantee that the lock exists on chain, as it
+    /// may have been deleted.
+    type Lock: HasLockId;
```
allanbrondum: Instead of an associated type, it is simpler just to use the type LockP11 directly? (will have less conflicts with my changes, and the block_state.rs and block_state_interface.rs is temporary code that is bound to be deleted anyway.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -183,29 +185,17 @@ pub fn execute_meta_update_transaction<BSO: BlockStateOperations>(
...
+                        &operation,
+                    )
+                });
+
                 match token_update_result {
-                    Ok(()) => {
-                        if token_module_state_dirty {
-                            block_state.set_token_key_value_state(&token, token_module_state);
-                        }
-                    }
```
allanbrondum, line 208: Can we revert the refactor of this part of the code, since I'm rewriting this part of the code (the tokens part of the scheduler)? It seems it does not actually change the behaviour?

plt/plt-scheduler/tests/lock_cancel.rs
```diff
@@ -0,0 +1,205 @@
...
+        payload: MetaUpdatePayload {
+            operations: RawCbor::from(cbor::cbor_encode(&vec![lock_cancel(lock_id.clone(), None)])),
+        },
+    };
+    let summary = block_state
+        .execute_transaction(&mut context, transaction_context, account_index_2, payload)
+        .unwrap();
+    assert_matches!(summary.outcome, TransactionOutcome::Success(events) => {
+        assert_eq!(events.len(), 1);
+        assert_matches!(&events[0], BlockItemEvent::LockDestroyed(LockDestroyEvent{lock_id: event_lock_id}) => {
```
allanbrondum, line 86: Should there also be some tests that emits the new type of transfer event?

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
allanbrondum: (the function can e.g. be in the entity model, or just as a free function in the model where it is used)

td202: I put it in the persistence layer because the implementation is tied to how the lock controller state is stored. In particular, I expect that we will change how grants are stored to more efficiently implement this function.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -302,9 +316,50 @@ fn execute_lock_operation<BSO: BlockStateOperations>(
 
             block_state.create_lock(lock_id.clone(), configuration);
 
-            Ok(None)
+            Ok(())
+        }
+        LockOperation::Cancel(meta_lock_cancel_details) => {
+            // TODO: charge
+            let lock = block_state
```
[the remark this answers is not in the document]
td202: This is to be handled in COR-2306.

plt/plt-scheduler/src/token_module/module/queries.rs
```diff
@@ -93,6 +93,8 @@ fn query_token_module_account_state_impl<BSQ: BlockStateQuery>(
     TokenModuleAccountState {
         allow_list,
         deny_list,
+        locks: vec![],   // FIXME: COR-2316
+        available: None, // FIXME: COR-2316
     }
```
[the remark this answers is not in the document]
td202: Out of scope for this PR.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
td202: I disagree that this is business logic. The fact that we have to iterate over the grants to determine if an account has a particular role is an artefact of how the controller state is represented, and should change with the persistence logic.

plt/plt-scheduler/src/ffi/queries.rs
td202: OK, I will revert it. The reason I changed this was because `query_lock_list` would panic on P9/P10.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
td202: OK.

plt/plt-scheduler/tests/lock_cancel.rs
td202: Added

allanbrondum:
LGTM, but have not looked at much into the lock stuff.

plt/plt-scheduler/tests/lock_cancel.rs
```diff
@@ -0,0 +1,333 @@
...
+        energy_limit: Energy::from(u64::MAX),
+        sender_account_address: context.external.account_canonical_address(account_index_2),
+        transaction_sequence_number: 1.into(),
+        block_timestamp: 0.into(),
+    };
+    let payload = Payload::MetaUpdate {
+        payload: MetaUpdatePayload {
+            operations: RawCbor::from(cbor::cbor_encode(&vec![lock_cancel(lock_id.clone(), None)])),
+        },
+    };
```
soerenbf, line 309: I think it's worth adding a memo here to verify that this is included in the resulting event(s) as well. Assuming that is the intended purpose of the memo; the design docs don't explicitly state what happens to the memo 🤔

plt/plt-scheduler/tests/lock_cancel.rs
soerenbf, line 1: Maybe a test asserting the error returned when trying to cancel a lock that does not exist?

soerenbf (approved):

plt/plt-block-state/src/block_state_interface.rs
td202: I instinctively don't like this, but I've made the change. If the intention is to delete this trait, then that's fine.

plt/plt-scheduler/tests/lock_cancel.rs
td202: Good call.

PR #1644, opened 2026-05-07
Ar/cor 2397 implement block state entity model

allanbrondum:
## Purpose

Implement statically typed and protocol specific block state types.

## Changes

Persistent types are in the module `persistent`. The types that the scheduler sees are in `entity`.

The module `block_state` defines temporary implementations of block state query and operations traits in order to make the changes in the scheduler minimal in this PR. Next PR will remove those two traits and push the new entity types into the scheduler.

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/utils.rs
```diff
@@ -1,40 +1,72 @@
...
-pub enum OwnedOrBorrowed<'a, T> {
+pub enum Cow<'a, T> {
     /// Value is owned
     Owned(T),
     /// Value is borrowed
     Borrowed(&'a T),
 }
 
-impl<'a, T> OwnedOrBorrowed<'a, T> {
+impl<'a, T> Cow<'a, T> {
```
allanbrondum, line 21: Renamed to `Cow`, since this is essentially just our own variant of std `std::borrow::Cow`

plt/plt-block-state/src/persistent/block_state/p10.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::block_state::hash::Hashable;
+use crate::block_state_interface::{BlockStateFailure, BlockStateResult};
+use crate::persistent::block_state::p9::PersistentBlockStateP9;
+use concordium_base::common::Buffer;
+use concordium_base::hashes::Hash;
+use std::io::Read;
+
+/// P10 block state.
+#[derive(Debug, Clone, Default)]
+pub struct PersistentBlockStateP10 {
```
limemloh: Maybe just make it a type-alias?

plt/plt-block-state/src/entity/block_state.rs
```diff
@@ -0,0 +1,27 @@
...
+use crate::entity::{EntityContext, EntityContextTypes};
+use concordium_base::base::AccountIndex;
+use concordium_base::contracts_common::AccountAddress;
+
+pub mod p10;
+pub mod p11;
+pub mod p9;
+
+/// Trait that defines block state operations related to accounts.
+pub trait Accounts {
```
limemloh, line 14: Every operation here require the `EntityContext` can't we just call these function directly on the context and remove this trait?
It seems they all have the same implementation anyway

I see the `Account` wrapper on return types, but this could be moved to the entity context as well.

plt/plt-block-state/src/persistent/block_state.rs
```diff
@@ -0,0 +1,117 @@
...
+pub mod p11;
+pub mod p9;
+
+/// Persistent block that that may represent any protocol version know to the Rust scheduler.
+#[derive(Debug, Clone)]
+pub enum PersistentBlockState {
+    P9(PersistentBlockStateP9),
+    P10(PersistentBlockStateP10),
+    P11(PersistentBlockStateP11),
+}
```
limemloh: I am not a fan of this union, since it results in the largest block state dictating the size of every block state.

ATTOW `PersistentBlockStateP11` takes up 8 bytes, but `PersistentBlockStateP10` and `PersistentBlockStateP9` each takes up 64 bytes, and this might get worse as we port more to rust.

plt/plt-block-state/src/entity.rs
```diff
@@ -0,0 +1,89 @@
...
+        context: &EntityContext<C>,
+        blob_ref: BlobStoreLocation,
+    ) -> BlockStateP9 {
+        let persistent_block_state: PersistentBlockStateP9 =
+            blob_store::load_from_store(&context.loader, blob_ref).expect("load block state");
+        BlockStateP9 {
+            persistent: persistent_block_state,
+        }
+    }
+}
```
limemloh: Not sure these are used, but I would expect:

```suggestion
    pub fn load_block_state_p10<C: EntityContextTypes>(
        context: &EntityContext<C>,
        blob_ref: BlobStoreLocation,
    ) -> BlockStateP10 {
        let persistent_block_state: PersistentBlockStateP10 =
            blob_store::load_from_store(&context.loader, blob_ref).expect("load block state");
        BlockStateP10 {
            persistent: persistent_block_state,
        }
    }

    pub fn load_block_state_p11<C: EntityContextTypes>(
        context: &EntityContext<C>,
        blob_ref: BlobStoreLocation,
    ) -> BlockStateP11 {
        let persistent_block_state: PersistentBlockStateP11 =
            blob_store::load_from_store(&context.loader, blob_ref).expect("load block state");
        BlockStateP11 {
            persistent: persistent_block_state,
        }
    }
}
```

plt/plt-scheduler-types/src/types/protocol_version.rs
```diff
@@ -0,0 +1,37 @@
+use thiserror::Error;
+
+/// Protocol version relevant for the Rust scheduler.
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug)]
+pub enum ProtocolVersion {
```
limemloh, line 5: Can we give this a more descriptive name? I think name clashing is a bit confusing

plt/plt-block-state/Cargo.toml
```diff
@@ -23,6 +23,7 @@ im.workspace = true
 libc = { workspace = true, optional = true }
 sha2.workspace = true
 thiserror.workspace = true
+util = "0.1.3"
```
limemloh: Is this a mistake? Never heard of this crate and where are we even using it?

plt/plt-block-state/src/persistent/lfmb_tree.rs
```diff
@@ -769,25 +758,21 @@ impl<'a, 'b, L: BlobStoreLoad, V: Loadable> Iterator for ValuesIterator<'a, 'b,
...
+        SubtreeCowProjection::Leaf(val_ref) => {
+            val_ref.bind_value(loader, "leaf in lfmb_tree::Subtree")?
+        }
+        SubtreeCowProjection::Node(_, left_ref, right_ref) => {
+            let right = right_ref.bind_value(loader, "left branch in lfmb_tree::Subtree")?;
             node_stack.push(right);
-            let left = left_ref
-                .bind_value(loader)?
-                .expect("Owned right branch reference returned borrowed value");
+            let left = left_ref.bind_value(loader, "right branch in lfmb_tree::Subtree")?;
```
limemloh, line 775:
```suggestion
            let right = right_ref.bind_value(loader, "right branch in lfmb_tree::Subtree")?;
node_stack.push(right);
            let left = left_ref.bind_value(loader, "left branch in lfmb_tree::Subtree")?;
```

plt/plt-scheduler/src/ffi/scheduler.rs
```diff
@@ -19,8 +26,21 @@ use plt_block_state::ffi::block_state_callbacks::{
     ReadTokenAccountBalanceCallback, TouchTokenAccountCallback, UpdateTokenAccountBalanceCallback,
 };
 use plt_block_state::ffi::memory;
+use plt_block_state::persistent::block_state::PersistentBlockState;
+use plt_block_state::persistent::block_state::p10::PersistentBlockStateP10;
 use plt_scheduler_types::types::execution::{ChainUpdateOutcome, TransactionOutcome};
 
+/// Context with no external block state (will panic if accessed).
+#[derive(Debug)]
+pub struct FfiQueryBlockStateTypes;
```
limemloh: This is named ...Query... but maybe it should use ...Operation... instead?

plt/plt-block-state/tests/block_state_p11.rs
```diff
@@ -0,0 +1,185 @@
...
+    // Read values
+    let mut token = block_state.token_by_index(&context, token_index).unwrap();
+    let mut expected_roles = Roles::none();
+    expected_roles.assign(TokenAdminRole::Mint);
+    expected_roles.assign(TokenAdminRole::Burn);
+    assert_eq!(
+        token.get_account_roles(&context, account_index1).unwrap(),
+        expected_roles
+    );
+    // todo ar token authoriations
```
limemloh: Will this be addressed later?

limemloh (approved):

plt/plt-scheduler/src/ffi/scheduler.rs
allanbrondum: Yes, copy error, will rename

plt/plt-block-state/tests/block_state_p11.rs
allanbrondum: Yes, I'll address it when adjusting the scheduler tests. The reason I could not test it was that I need the external block state (accounts). Will change such that the block state tests also has this stubbed, like the scheduler.

plt/plt-block-state/Cargo.toml
allanbrondum: Yes, must have been added by autocompletion of "util::..."

plt/plt-block-state/src/persistent/block_state.rs
allanbrondum: Though I think any block state will be a composition of `HashedCacheableRef`s which are `Arc`s? So size difference wont be that big? Anyways, we can always `Box` variants if size difference becomes too big?

plt/plt-block-state/src/entity/block_state.rs
allanbrondum: My idea was that we might have different/more traits. `Accounts` just being one example of what a block state can implement. There could also be `AccountsV2` e.g. in the future. But we can change it if something else makes more sense.

plt/plt-block-state/src/persistent/block_state.rs
limemloh: yes, but it means we need to be aware of this, but I guess clippy will warn us since we don't have any generics

plt/plt-block-state/src/utils.rs
```diff
@@ -0,0 +1,72 @@
...
+use concordium_base::common::cbor;
+use concordium_base::common::cbor::{
+    CborDeserialize, CborSerializationResult, SerializationOptions, UnknownMapKeys,
+};
+use std::ops::Deref;
+
+/// Value of type `T` that is either owned or borrowed.
+///
+/// We use our own type instead of [`std::borrow::Cow`], since we don't want to require
+/// `T` to implement `Clone` which `Cow` does.
```
td202, line 12: I'm not sure where we really need to be able to have a `Cow` without `T` being `Clone` (or at least `ToOwned`). Could you please elaborate why this is necessary more specifically?

allanbrondum: An annoying thing is that you need to declare T: Clone in any context where you use the std `Cow`. Also on functions where you never call `clone`. That is not so nice and a bit non-standard I think (despite Rust std doing it).

PR #1648, opened 2026-05-22
Ar/cor 2398 push block state entity model into the scheduler

allanbrondum:
## Purpose

Use the new entity block state model in the tokens part of the scheduler. Leave locks untouched where possible.

## Changes

* Removed use of `BlockStateQuery` and `BlockStateOperations` from token scheduler and queries. Also removed `TokenOperationContext`. All queries and transaction and chain update execution is now has a variant per protocol. 
* The token module is though polymorphic using the TokenPX enum
* Introduced `TokenP9Base` which `TokenP9` and `TokenP11` is composed of instead of `TokenP11` having `TokenP9` inside of it. It means you only have the type `TokenP9` if you are actually on P9.
* Move accounts lookup functions on the `Accounts` trait to be implemented by `EntityContext`

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/entity/protocol_level_tokens/p11.rs
```diff
@@ -136,98 +133,43 @@ impl TokenP11 {
...
-            update_allow_list: self
-                .token_p9
-                .has_allow_list(context)
-                .then_some(update_allow_list),
-            update_deny_list: self
-                .token_p9
-                .has_deny_list(context)
-                .then_some(update_deny_list),
-            pause: Some(pause),
-            update_metadata: Some(update_metadata),
```
allanbrondum, line 219: Moved to `plt_scheduler::protocol_level_tokens::p11`

plt/plt-block-state/src/block_state_interface.rs
```diff
@@ -9,65 +15,7 @@ use concordium_base::protocol_level_locks::LockId;
...
-    /// An invariant that must be true is broken. The invariant can either be in the
-    /// stored block state or a runtime logical invariant related to the in-memory block state.
-    #[error("State invariant broken: {0}")]
-    Invariant(String),
-    /// When looking up a value with in an owned
-    /// [blob reference](super::block_state::blob_reference::hashed_cacheable_reference::HashedCacheableRef),
-    /// a borrowed value was returned. This should generally never happen in they way we maintain
-    /// blob references.
-    #[error("Borrowed value found inside of owned value: {0}")]
-    CowJoin(&'static str),
```
allanbrondum, line 67: Moved all these types to more appropriate locations, since the entire file is bound for deletion

plt/plt-scheduler/src/ffi/queries.rs
```diff
@@ -97,40 +98,33 @@ extern "C" fn ffi_query_plt_list(
...
+        let token_ids_res = match unsafe { &*block_state } {
             PersistentBlockState::P9(persistent) => {
                 let block_state = BlockStateP9 {
                     persistent: persistent.clone(),
                 };
-                let exec_block_state = ExecutionTimeBlockStateP9 {
-                    block_state,
-                    context,
-                };
-                queries::query_plt_list(&exec_block_state)
```
limemloh, line 109: These top-level queries I would probably move to a `BlockState` trait, since this functionality is actually something we need for every block state there is, even the old ones, since we have to decide what the behavior is for them as well.
Can be done in a future iteration

plt/plt-scheduler/src/protocol_level_tokens/token_module/initialize.rs
```diff
@@ -25,89 +23,105 @@ pub enum TokenInitializationError {
...
-impl From<TokenMintError> for TokenInitializationError {
-    fn from(err: TokenMintError) -> Self {
-        match err {
-            TokenMintError::StateInvariantViolation(err) => Self::StateInvariantViolation(err),
-            TokenMintError::MintWouldOverflow(err) => Self::MintAmountNotRepresentable(err),
-        }
-    }
-}
+/// List roles which are unaffected by which features are enabled.
+const UNIVERSAL_ROLES: &[TokenAdminRole] = &[
```
limemloh, line 31: I am afraid we forget to update this, when it does not live next to the list of all roles

limemloh (approved):

plt/plt-scheduler/src/ffi/queries.rs
allanbrondum: There is the trait `SchedulerOperations` in the `plt-scheduler` integration tests. I did not move it to the production code, because there was not really a use case for it. In the tests it is used to make tests generic over the protocol version.

plt/plt-scheduler/src/protocol_level_tokens/token_module/initialize.rs
allanbrondum: Mentioned it in the comments for ALL_RULES

PR #1649, opened 2026-05-27
Lock transfer

td202:
Closes COR-2305, Closes COR-2395

## Purpose

Implements the lock transfer operations. This also changes the behaviour of token transfers to validate the transfer amount against the "available" token balance of the sender account, which for protocol versions lower than 11 will never differ from the total token balance.

## Checklist

      hard-to-understand areas.


plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -279,17 +285,262 @@ where
...
+                    transaction_execution.sender_account_address(),
+                )
+                .into());
+            }
+
+            let token = block_state
+                .token_by_id(&meta_lock_fund_details.token)
+                .map_err(|TokenNotFoundByIdError(token_id)| {
+                    TransactionRejectReason::NonExistentTokenId(token_id)
+                })?;
```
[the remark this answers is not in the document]
td202: Should now be handled.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -279,17 +285,262 @@ where
...
+            let source = block_state
+                .account_by_address(&source_address)
+                .map_err(|_| TransactionRejectReason::InvalidAccountReference(source_address))?;
+
+            let recipient_address = meta_lock_send_details.recipient.address;
+            let recipient = block_state
+                .account_by_address(&recipient_address)
+                .map_err(|_| TransactionRejectReason::InvalidAccountReference(recipient_address))?;
+            let destination = Some((&recipient, recipient_address));
+
```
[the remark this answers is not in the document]
td202: Should be fixed.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -279,17 +285,262 @@ where
...
+            let memo: Option<transactions::Memo> = meta_lock_return_details
+                .memo
+                .clone()
+                .map(transactions::Memo::from);
+
+            let source_address = meta_lock_return_details.source.address;
+            let source = block_state
+                .account_by_address(&source_address)
+                .map_err(|_| TransactionRejectReason::InvalidAccountReference(source_address))?;
+
```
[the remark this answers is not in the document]
td202: The token restriction in the simple controller only affects funding, not send or return. Otherwise, the logic should now be implemented in the controller's validation.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -279,17 +285,262 @@ where
...
+                    })
+            })?;
+            if is_removed_lock_holder {
+                // The lock controller state needs to be updated to reflect
+                // that the lock no longer controls any of this token for the account.
+                block_state.remove_lock_balance_ref(
+                    lock.lock_id(),
+                    transaction_execution.sender_account(),
+                    &token,
+                );
```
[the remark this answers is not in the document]
td202: Should be fixed.

plt/plt-scheduler/src/token_module/key_value_state.rs
```diff
@@ -462,6 +462,31 @@ pub fn get_locked_balance_for(
...
+        .iter_token_state_prefix(&prefix)
+        .map(|(key, value)| {
+            let lock = common::from_bytes_complete(&key.0[prefix.0.len()..]).map_err(|err| {
+                TokenStateInvariantError(format!("Stored lock id cannot be decoded: {}", err))
+            })?;
+            let amount = decode_locked_balance(value)?;
+            Ok((lock, amount))
+        })
+        .collect()
+}
```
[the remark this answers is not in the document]
soerenbf: This is already covered tests in both the scheduler and block state.

plt/plt-block-state/src/block_state_interface.rs
```diff
@@ -329,4 +329,26 @@ pub trait BlockStateOperations: BlockStateQuery {
...
+    /// - `lock` The lock to update.
+    /// - `account` The account whose locked balance is no longer tracked.
+    /// - `token` The token whose locked balance is no longer tracked.
+    ///
+    /// The caller must ensure the following conditions are true, and failing to do so results in
+    /// undefined behavior.
+    ///
+    /// - The `lock` MUST already exist in the block state, i.e.
+    ///   `s.lock_by_id(lock_id).expect("lock exists")`.
+    fn remove_lock_balance_ref(
```
limemloh, line 348: We are getting rid of this interface, so we should not extend it.
I think we already not using the interface, so this is probably just 'legacy'

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -1,41 +1,366 @@
-//! Scheduler implementation for protocol-level token updates. This module implements execution
-//! of transactions related to protocol-level tokens.
+//! Scheduler implementation for protocol-level lock operations.
```
limemloh: The description is not really more accurate than the previous one

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -1,41 +1,366 @@
...
     match lock_operation {
-        LockOperation::Fund(_meta_lock_fund_details) => todo!(),
-        LockOperation::Send(_meta_lock_send_details) => todo!(),
-        LockOperation::Return(_meta_lock_return_details) => todo!(),
+        LockOperation::Fund(meta_lock_fund_details) => {
+            // TODO: (COR-2306) charge.
+            let bsq = ExecutionTimeBlockStateP11 {
+                block_state: block_state.clone(),
+                context: context.clone(),
+            };
```
limemloh: I thought we were getting rid of this? 
I think this code still depends on block state query and operations.

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -1,41 +1,366 @@
...
+                token_id: token_configuration.token_id,
+                from: TokenHolder::Account(transaction_execution.sender_account_address()),
+                to: TokenHolder::Account(transaction_execution.sender_account_address()),
+                amount: TokenAmount::from_raw(raw_amount.0, token_configuration.decimals),
+                memo,
+                from_lock: None,
+                to_lock: Some(meta_lock_fund_details.lock),
+            }));
+        }
+        LockOperation::Send(meta_lock_send_details) => {
```
limemloh: It seems Send and Return are very similar, did you try to abstract it out into a function? or what are your thoughts

limemloh (approved):
Have some comments, but nothing which can't be dealt with in another PR

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
soerenbf: Yeah I thought the same, mainly built on top of what was there. Will try to factor the common paths out 😊

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
soerenbf: I don't know, I have a hard time keeping track of what is in/out when it comes to all the block state changes.. I'll have a look.

plt/plt-block-state/src/block_state_interface.rs
soerenbf: Yeah.. I think it's easier to take all of this when refactoring.. a lot of existing code is still using these traits

plt/plt-scheduler/src/scheduler/plt_scheduler.rs
soerenbf: the lock controller as it is is based on the block state query trait.. I think it'll be easier to refactor all of it along with the query code to get rid of what we want to get rid of once and for all.

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
```diff
@@ -391,63 +391,140 @@ fn check_authorized<C: EntityContextTypes>(
...
+///
+/// # Errors
+///
+/// Returns a [`TransferConstraintError`] describing the first constraint that
+/// is violated.
+pub(crate) fn check_transfer_constraints<C: EntityContextTypes>(
+    context: &EntityContext<C>,
+    token: &TokenP9Base,
+    sender: Result<Account, AccountNotFoundByAddressError>,
+    receiver: Result<Account, AccountNotFoundByAddressError>,
```
limemloh: This function is a bit unusual, why would this function handle the results and not the caller? we even pass back the results again, so this function is just mapping the errors, which is then mapped again at every callsite.

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
```diff
@@ -391,63 +391,140 @@ fn check_authorized<C: EntityContextTypes>(
...
-    transaction_execution: &mut TransactionExecution,
-    context: &mut EntityContext<C>,
-    events: &mut impl Extend<BlockItemEvent>,
-    token: &mut TokenP9Base,
-    transfer_operation: &TokenTransfer,
-) -> Result<(), TokenUpdateErrorInternal> {
-    let token_configuration = token.token_configuration(context)?;
+/// Resolved and validated sender and receiver accounts, returned by
+/// [`check_transfer_constraints`] on success.
+pub(crate) struct ValidatedAccounts {
```
limemloh: I don't think we need this

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
```diff
@@ -391,63 +391,140 @@ fn check_authorized<C: EntityContextTypes>(
...
+    /// The token is currently paused; no balance-affecting operations are
+    /// allowed.
+    Paused,
+    /// The sender account is not permitted to send (not in the allow list, or
+    /// in the deny list). The `reason` string is a human-readable explanation.
+    SenderNotAllowed { reason: &'static str },
+    /// The recipient account is not permitted to receive (not in the allow
+    /// list, or in the deny list). The `reason` string is a human-readable
+    /// explanation.
+    RecipientNotAllowed { reason: &'static str },
```
limemloh: It seems both of these are being mapped to the same notpermitted error in both places this is used. Do we really need this error type? or can we just use something we have like `TokenUpdateErrorInternal`

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
```diff
@@ -391,63 +391,140 @@ fn check_authorized<C: EntityContextTypes>(
...
+    /// One of the account addresses could not be resolved in the block state.
+    AccountNotFound(AccountNotFoundByAddressError),
+}
 
-    // operation execution
-    check_not_paused(context, token)?;
+/// Validate that a token transfer between `sender` and `receiver` is permitted
+/// by the token module's current state.
+///
+/// Checks, in order:
```
limemloh: For reject reasons, the ordering is not important because this is not included in the block hash.

soerenbf: yeah.. this was the whole reason why this check function was so acrobatically structured 😆 I'll revise, thanks for pointing this out!

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
soerenbf: I tried another approach that uses the internal error instead.

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
soerenbf: I agree, I thought the order things were checked mattered 🫠 which is the reason for this mess. It's gone with the next commit

limemloh (approved):

PR #1651, opened 2026-05-28
Fix migration reads from incorrect blob store.

td202:
## Purpose

This fixes two (relatively minor) issues in the block state migration code where a function with a `MonadBlobStore` constraint is being executed against the incorrect `BlobStore`. While these are potentially catastrophic issues, in practice these are not actually seen.

## Changes

- `accountCanonicalAddress` moved inside `lift`. Although this in principle could be an issue, the implementation of `accountCanonicalAddress` does not actually read from the `BlobStore`. (Of course, this could potentially change in the future.)
- `bakersFromEpochBlocks` moved inside `lift`. This only applies to the P3->P4 migration and is likely never triggered due to the caching behaviour ensuring that the epoch blocks are loaded into memory, and therefore will not be read from the (incorrect) `BlobStore`.


soerenbf (approved):
Looks good

limemloh (approved):

PR #1652, opened 2026-06-03
Use LfmbTree for locks, introducing LockIndex and an index from LockId

limemloh:
## Purpose                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                          
   Implements the real persistent state model for protocol-level locks, replacing the dummy in-memory `BTreeMap`-backed implementation. Closes [COR-2295](https://linear.app/concordium/issue/COR-2295/locks-real-state-implementation).                                  
                                                                                                                                                                                                                                                                          
   ## Changes                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                          
   - **Introduced `LockIndex`**: a sequential internal key type for the lock `LfmbTree`, consistent with how `TokenIndex` works for tokens. It is intentionally internal and must never appear in the public API, events, or state hashing.                               
                                                                                                                                                                                                                                                                          
   - **Replaced `BTreeMap<LockId, PersistentLockP11>` with `LfmbTree<LockIndex, Option<PersistentLockP11>>`**: the tree uses `None` entries as tombstones for deleted (cancelled) locks. A companion `im::HashMap<LockId, LockIndex>` serves as a derived in-memory       
 reverse index, rebuilt on every load from the blob store by scanning the tree and reading each live lock's configuration.                                                                                                                                                
                                                                                                                                                                                                                                                                          
   - **Embedded `LockId` inside `LockConfiguration`**: the `create_lock` call no longer takes a separate `lock_id` argument — the ID is extracted from the configuration. This avoids dual-passing and makes the configuration self-contained.                            
                                                                                                                                                                                                                                                                          
   - **Changed `delete_lock` to return `bool`**: the return type changed from `Option<LockP11>` (the deleted lock) to `bool` (whether a lock was deleted). All callers already held a `LockP11` reference and only needed to know whether deletion occurred.              
                                                                                                                                                                                                                                                                          
   - **`PersistentLockP11::configuration` stored as `HashedCacheableRef`**: the lock configuration is now stored as a blob-store reference rather than inlined, consistent with how token configuration is stored.                                                        
                                                                                                                                                                                                                                                                          
   - **Fixed `lock_list` ordering and documented the lack of guarantee**: the switch from `BTreeMap` to `im::HashMap` for the internal index means `lock_list` no longer returns lock IDs in sorted order. The trait doc now explicitly states no ordering guarantee and  
 warns callers not to drive state-mutating scheduler operations from the unordered iterator without first sorting, to preserve consensus safety.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


plt/plt-scheduler/src/scheduler/plt_scheduler.rs
```diff
@@ -474,8 +481,8 @@ fn remove_lock_balance_ref<C: EntityContextTypes>(
     context: &EntityContext<C>,
     block_state: &mut BlockStateP11,
     events: &mut Vec<BlockItemEvent>,
+    lock_keeps_alive: bool,
     mut lock: plt_block_state::entity::protocol_level_locks::p11::LockP11,
-    lock_configuration: &LockConfiguration,
     account_index: AccountIndex,
```
limemloh, line 465: I removed cloning the lock_configuration meaning the borrow checker was unhappy about the lock being mutable here while holding a reference to the internal configuration. So I had to remove the configuration from this function signature.

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Consensus/Blocks.hs
soerenbf, line 1: Just a verification check: this means a new devnet with the release of this block state representation, right?

plt/plt-block-state/src/entity/protocol_level_locks/p11.rs
```diff
@@ -1,30 +1,40 @@
...
+        configuration: HashedCacheableRef::new(StoreSerialized(configuration)),
     };
-    let existing = persistent_locks.locks.0.insert(lock_id.clone(), persistent);
+    let (lock_index, updated_locks) = persistent_locks
+        .locks
+        .insert_value(&context.loader, Some(persistent))?;
+    persistent_locks.locks = updated_locks;
+    let existing = persistent_locks
+        .lock_id_map
+        .insert(lock_id.clone(), lock_index);
```
soerenbf, line 37: Would it make sense to do a check whether the lock ID exists in the lock id map prior to inserting into the lfmbtree?
```suggestion
    if persistent_locks.lock_id_map.contains(&lock_id) { return Err(...) };
    
    let (lock_index, updated_locks) = persistent_locks
        .locks
        .insert_value(&context.loader, Some(persistent))?;
    persistent_locks.locks = updated_locks;
    let _ = persistent_locks
        .lock_id_map
        .insert(lock_id.clone(), lock_index);
```
It probably makes no difference in practice, but might be good to do anyway for maintainability? 🤷

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -49,20 +89,102 @@ impl Hashable for PersistentLocksP11 {
...
+        Ok(hash)
+    }
+}
+
 /// The block state for a single protocol-level lock.
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone)]
 pub struct PersistentLockP11 {
     /// Contains references to the tokens with balances locked within this lock
     pub locked_balances: BTreeSet<(AccountIndex, TokenIndex)>,
```
soerenbf: Maybe a note that this means we write all these every time any of them change

soerenbf (approved):

concordium-consensus/tests/consensus/ConcordiumTests/KonsensusV1/Consensus/Blocks.hs
limemloh: yes, the state has changed

plt/plt-block-state/src/entity/protocol_level_locks/p11.rs
limemloh: state changes should already be rolled back in this case, so we should be good and this way we avoid the extra lookup.

plt/plt-block-state/tests/block_state_p11.rs
allanbrondum, line 1: When we add new operations to the block state, like deleting a lock, remember to add tests of the operations on the block state here at integration test level. I think we should have tests of deleting locks, and then looking up the lock after that.

allanbrondum: Also the test a unit test level in `test_store_and_load_locks` which focuses on storage and hashing, should also be updated to include deleted locks.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
allanbrondum: For consistency, consider:
```suggestion
    pub locked_balances: StoreSerialized<BTreeSet<(AccountIndex, TokenIndex)>>,
```

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
allanbrondum, line 1: There is a vec of `TokenId`s in `LockControllerSimpleV0` that should probably be `TokenIndex`s

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -49,20 +89,102 @@ impl Hashable for PersistentLocksP11 {
     }
 }
 
+impl Loadable for Option<PersistentLockP11> {
```
allanbrondum: Maybe it will be more readable if we promote this to be generic blanket implementations. E.g.
`impl<T: Loadable> Loadable for Option<T>` and the same for the other traits. I think it could be one of the cases where code actually becomes more readable by being more general.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -1,33 +1,73 @@
...
+
 /// Block state for protocol level locks on P11 and later protocols that uses the same representation.
 #[derive(Debug, Clone, Default)]
 pub struct PersistentLocksP11 {
-    pub(crate) locks: StoreSerialized<BTreeMap<LockId, PersistentLockP11>>,
+    /// Persistent map of lock index to locks.
+    ///
+    /// Here `None` represents a lock which has been deleted and acts as tombstone, preventing new
+    /// locks from using the same index.
+    pub(crate) locks: LfmbTree<LockIndex, Option<PersistentLockP11>>,
```
allanbrondum, line 43: I'm not sure that `LmfbTree` is a good data structure, for objects that are expected to have a limited lifespan, like locks? The tree will grow indefinitely. And migration, and even loading, because of the `LockId` map, will be have linear complexity in the size of the tree. An alternative is using the smart contract trie. Or we can use `LmfbTree` for now with the expectation of writing a new trie in Rust, dedicated for the block state, and then migrate to that.

allanbrondum (approved):

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
limemloh: yes, I agree, this won't last long term, and we need to have something better, but it will last for some time until we have a plan for the block state.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
limemloh: I am a bit hesitant to do this, because I am not convinced there is a general implementation of the traits for `Option`. But I am not able to come up with a counterargument, so I'll just do it.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
limemloh: This would put the TokenIndex into the block hash which is not ideal... but maybe inevitable.

limemloh: I go back and forth on this one, so made a task for the final decision [here](https://linear.app/concordium/issue/COR-2408/node-decide-whether-to-use-tokenindex-or-tokenid-within-the-lock) and won't do anything in this PR

PR #1635, opened 2026-04-23
P10->P11 migration into Rust block state

allanbrondum:
## Purpose

Implement migration of the Rust block state from P10 to P11 on the Haskell side.

## Changes

The P10 PLT state is loaded by Rust and then migrated to P11 by Rust also.

## Checklist

- [x] My code follows the style of this project.
- [x] The code compiles without warnings.
- [x] I have performed a self-review of the changes.
- [x] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG.

PR #1654, opened 2026-06-12
Ar/psr 91 implement migration of rust block state and block state 2

allanbrondum:
## Purpose

Implement Rust block state migration. This includes moving from blob store to blob store. But also protocol specific migration that adds authorization roles to the governance account in the P10 to P11 migration.

## Changes

* implemented block state high level migration entry point in `entity::block_state::migration`
* implemented plain blob store movement of persistent block state via the trait `BlobStoreMovable`
* P10 to P11 PLT migration is in `entity::protocol_level_tokens::migration::p10_to_p11`

## Checklist

      hard-to-understand areas.


plt/plt-scheduler/src/protocol_level_tokens/token_module/initialize.rs
```diff
@@ -27,13 +28,6 @@ pub enum TokenInitializationError {
     MintAmountNotRepresentable(#[from] MintWouldOverflowError),
 }
 
-/// List roles which are unaffected by which features are enabled.
-const UNIVERSAL_ROLES: &[TokenAdminRole] = &[
-    TokenAdminRole::UpdateAdminRoles,
-    TokenAdminRole::Pause,
-    TokenAdminRole::UpdateMetadata,
-];
```
allanbrondum, line 35: Moved to block state crate such that it can be reused from migration.

soerenbf (approved):
Looks good to me. Would be good to comment that this is succeeded by #1635 to make it easier to discover the PR chain 😉

plt/plt-block-state/src/entity/block_state/migration.rs
```diff
@@ -0,0 +1,119 @@
...
+
+pub mod p10_to_p11;
+pub mod p9_to_p10;
+
+/// Migrate the PLT block state from protocol version to another. The persistent block state
+/// is first "lifted" into an entity block state and then migrated.
+///
+/// # Arguments
+///
+/// - `from_store` Blob store loader for the blob store we are migrating from.
```
limemloh, line 18: Nitpick: The list is missing `block_state`, although it is kind of obvious 🤷 
Just an observation

plt/plt-block-state/src/block_state.rs
```diff
@@ -23,19 +23,31 @@ use concordium_base::contracts_common::AccountAddress;
...
 /// [`BlockStateQuery`] and [`BlockStateOperations`].
-#[derive(Debug)]
 pub struct ExecutionTimeBlockStateP9<C: EntityContextTypes> {
     /// The library block state implementation.
     pub block_state: BlockStateP9,
     /// External function for reading from the blob store.
     pub context: EntityContext<C>,
 }
 
+impl<C: EntityContextTypes> std::fmt::Debug for ExecutionTimeBlockStateP9<C>
```
limemloh, line 39: Consider adding comment to why the manual implementation is needed.

plt/plt-scheduler/src/protocol_level_tokens/token_module/initialize.rs
limemloh: Would future additions to this list break the migration? Say we add a new role in P12 and the P10 -> P11 migration starts adding the new P12 role breaking consensus for nodes catching up.

limemloh (approved):

plt/plt-block-state/src/block_state.rs
allanbrondum: Type will be deleted very soon

plt/plt-scheduler/src/protocol_level_tokens/token_module/initialize.rs
allanbrondum: That is true, I'll change the list to be hardcoded in the migration

PR #1655, opened 2026-06-18
Handle "any" lock recipient variant in node

soerenbf:
Closes COR-2427

## Purpose

Adds handling for the "any" recipients variant in a lock configuration. This implements a temporary non-breaking solution in the block state to allow upgrading the current devnet with support.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
+++ b/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
@@ -75,24 +75,29 @@ pub enum LockRecipients {
 }
 
 impl LockRecipients {
-    /// Get an iterator over the limited recipient accounts.
-    fn recipients_iter(&self) -> impl Iterator<Item = &AccountIndex> {
-        match self {
-            Self::Any => &[],
-            Self::Limited(recipients) => recipients.as_slice(),
-        }
-        .iter()
+    /// Normalize the recipient representation for storage in block state.
+    pub fn limited(mut recipients: Vec<AccountIndex>) -> Self {
+        recipients.sort();
+        Self::Limited(recipients)
     }
 
     /// Check whether this representation allows any recipient.
-    fn is_any(&self) -> bool {
+    pub fn is_any(&self) -> bool {
         matches!(self, Self::Any)
+    }
+
+    /// Check if the given account is a recipient.
+    pub fn is_recipient(&self, account: &AccountIndex) -> bool {
+        match self {
+            Self::Any => true,
+            Self::Limited(recipients) => recipients.binary_search(account).is_ok(),
+        }
     }
 }
 
 impl From<Vec<AccountIndex>> for LockRecipients {
     fn from(recipients: Vec<AccountIndex>) -> Self {
-        Self::Limited(recipients)
+        Self::limited(recipients)
     }
 }
 
@@ -124,7 +129,7 @@ impl Deserial for LockRecipients {
         Ok(if recipients.as_slice() == [ANY_RECIPIENT_SENTINEL] {
             Self::Any
         } else {
-            Self::Limited(recipients)
+            Self::limited(recipients)
         })
     }
 }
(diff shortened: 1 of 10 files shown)


plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -58,48 +58,138 @@ pub struct PersistentLockP11 {
...
+        self.recipients.recipients_iter()
+    }
+
+    /// Check whether this configuration represents an any-recipient lock.
+    pub fn has_any_recipient(&self) -> bool {
+        self.recipients.is_any()
     }
 
     /// Check if the given account is a recipient.
     pub fn is_recipient(&self, account: &AccountIndex) -> bool {
```
allanbrondum: To me it seems that this function and other function that just forwards to `recipients` belongs on the `LockRecipients` type. This includes enforcing the invariant that recipient list is sorted. That invariant should be enforced in the constructor for `LockRecipients`. And then you can make all the fields in `LockConfiguration` pub and remove its constructor function.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -58,48 +58,138 @@ pub struct PersistentLockP11 {
...
             expiry,
             controller,
         }
     }
 
     /// Get an iterator over the recipient accounts.
     pub fn recipients_iter(&self) -> impl Iterator<Item = &AccountIndex> {
-        self.recipients.iter()
+        self.recipients.recipients_iter()
+    }
```
allanbrondum: Suggestion: Remove function, it does not seem type safe.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -58,48 +58,138 @@ pub struct PersistentLockP11 {
...
+
+impl LockRecipients {
+    /// Get an iterator over the limited recipient accounts.
+    fn recipients_iter(&self) -> impl Iterator<Item = &AccountIndex> {
+        match self {
+            Self::Any => &[],
+            Self::Limited(recipients) => recipients.as_slice(),
+        }
+        .iter()
+    }
```
allanbrondum:
```suggestion
```

plt/plt-scheduler/src/locks/lock_configuration.rs
```diff
@@ -6,27 +6,36 @@ use crate::{
...
+/// Get the recipients for a lock configuration, resolving [`AccountIndex`]es
+/// to [`CborHolderAccount`]s unless the block-state sentinel represents an
+/// any-recipient lock.
 fn get_recipients<BSQ: BlockStateQuery>(
     bsq: &BSQ,
     configuration: &LockConfiguration,
-) -> Result<Vec<CborHolderAccount>, AccountNotFoundByIndexError> {
-    configuration
+) -> Result<LockRecipients, AccountNotFoundByIndexError> {
+    if configuration.has_any_recipient() {
```
allanbrondum: Seems that it would be better to match on the variants of the enum in `configuration.recipients`?

allanbrondum (approved):

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -58,48 +58,138 @@ pub struct PersistentLockP11 {
...
+    fn is_any(&self) -> bool {
+        matches!(self, Self::Any)
+    }
+}
+
+impl From<Vec<AccountIndex>> for LockRecipients {
+    fn from(recipients: Vec<AccountIndex>) -> Self {
+        Self::Limited(recipients)
+    }
+}
```
allanbrondum, line 129: Remove since it mean you can create an unsorted list of recipients.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -58,48 +58,138 @@ pub struct PersistentLockP11 {
...
+    pub fn has_any_recipient(&self) -> bool {
+        self.recipients.is_any()
     }
 
     /// Check if the given account is a recipient.
     pub fn is_recipient(&self, account: &AccountIndex) -> bool {
-        self.recipients.binary_search(account).is_ok()
+        match &self.recipients {
+            LockRecipients::Any => true,
+            LockRecipients::Limited(recipients) => recipients.binary_search(account).is_ok(),
```
allanbrondum: Notice that the recipient list is no longer sorted. I suggest adding that invariant to the `LockRecipients` type, together with `has_any_recipient` and other functions accessing `recipients`.

allanbrondum: Also, we should have an automatic test catching this?

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -58,48 +58,138 @@ pub struct PersistentLockP11 {
...
+const ANY_RECIPIENT_SENTINEL: AccountIndex = AccountIndex { index: u64::MAX };
+
+/// Accounts that can receive funds from this lock in block state.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum LockRecipients {
+    /// Any eligible account can receive funds from this lock.
+    Any,
+    /// Only the listed accounts can receive funds from this lock.
+    Limited(Vec<AccountIndex>),
+}
```
allanbrondum, line 108: Since enum tuples are always `pub`, create a newtype `LockRecipientsList` to make it non-pub.

```suggestion
/// Accounts that can receive funds from this lock in block state.
#[derive(Debug, Clone, Eq, PartialEq)]
pub enum LockRecipients {
    /// Any eligible account can receive funds from this lock.
    Any,
    /// Only the listed accounts can receive funds from this lock.
    Limited(LockRecipientsList),
}

#[derive(Debug, Clone, Eq, PartialEq)]
pub struct LockRecipientsList {
    list: Vec<AccountIndex>,
}

impl LockRecipientsList {
    /// Construct a lock configuration.
    pub fn new(recipients: Vec<AccountIndex>) -> Self {
        todo!()
    }

    /// Get an iterator over the recipient accounts.
    pub fn recipients_iter(&self) -> impl Iterator<Item = AccountIndex> {
        todo!()
    }

    /// Check if the given account is a recipient.
    pub fn is_recipient(&self, account: &AccountIndex) -> bool {
        todo!()
    }
}
```

allanbrondum (changes requested):

plt/plt-scheduler/src/locks/lock_configuration.rs
soerenbf: yeah I agree

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
soerenbf: Yeah.. I feel like as the lock config is immutable after creation, it makes sense to have the type describe this.. but then looking at the lock controller implementation, this is not the case there.. will adapt to what you propose as it is the simpler solution.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
soerenbf: can add a test, I removed the sorting as a mistake. will add it back in, good catch :)

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
soerenbf: Good suggestion, have applied it.

allanbrondum (approved):
Looks nice!

PR #1661, opened 2026-06-26
Ar/cor 2436 refactor locks scheduler to use new block state types

allanbrondum:
## Purpose

Remove `BlockStateQueries` and `BlockStateOperations` traits in favour of using the (new) block state types directly.

## Changes

* remove `BlockStateQueries` and `BlockStateOperations` and refactored their use in locks queries and operations in the scheduler
* introduced the generic wrapper `WithBlockStateResult` which allows to use the error `BlockStateFailure` together with protocol level errors like `TransactionRejectReason` and `QueryLockError` in a "flattened" way without nesting results

## Checklist

      hard-to-understand areas.


plt/plt-scheduler/src/scheduler.rs
```diff
@@ -20,24 +19,6 @@ pub enum TransactionExecutionError {
     BlockStateFailure(#[from] BlockStateFailure),
 }
 
-/// [`BlockStateFailure`] and [`TransactionRejectReason`] flattened into one error
-/// for convenience.
-#[derive(Debug, thiserror::Error)]
-pub enum TransactionFailure {
```
allanbrondum, line 26: Replaced by the generic `WithBlockStateResult`

plt/plt-scheduler/src/queries.rs
```diff
@@ -24,23 +40,25 @@ pub enum QueryLockError {
     /// The requested lock does not exist in the block state.
     #[error("Lock does not exist")]
     LockDoesNotExist,
-    /// On-chain block state violates an invariant assumed by the
-    /// query assembly path. These are unrecoverable for clients and surface as `Panic`
-    /// over FFI.
-    #[error("State invariant violation while assembling lock info: {0}")]
-    StateInvariantViolation(String),
```
allanbrondum, line 31: Replaced with use of `BlockStateFailure`/`WithBlockStateResult`

limemloh (approved):

plt/plt-block-state/src/failure.rs
```diff
@@ -26,3 +28,39 @@ pub enum BlockStateFailure {
...
+/// error in the inner.
+pub fn nest<E, T>(result: WithBlockStateResult<T, E>) -> BlockStateResult<Result<T, E>> {
+    match result {
+        Ok(val) => Ok(Ok(val)),
+        Err(WithBlockStateFailure::Error(err)) => Ok(Err(err)),
+        Err(WithBlockStateFailure::BlockStateFailure(err)) => Err(err),
+    }
+}
+
+impl HigherLevelProtocolError for TransactionRejectReason {}
```
soerenbf: This feels a little over-engineered to me. Would it be simpler to use nested results directly, e.g. `BlockStateResult<Result<T, E>>`, throughout these paths? That makes the two error layers explicit in the type signature.

If we do keep this abstraction, I think it probably belongs in the scheduler crate rather than `plt-block-state`. At least for now, block state shouldn’t need to know about scheduler-level/protocol errors such as `TransactionRejectReason`

soerenbf (approved):
In general looks good to me, have one thing I'd change though.

plt/plt-block-state/src/failure.rs
allanbrondum: I'm not against using `BlockStateResult<Result<T, E>>` where it makes sense, but using `WithBlockStateFailure<T, E>` as a function return type make it possible to apply `?` to values of both type `Result<_, E>` and `Result<_, BlockStateFailure>` within the given function. That ergonomicity is the reason I introduced the type.

allanbrondum: Moved it to scheduler crate, agree that it is better placed there

PR #1665, opened 2026-07-01
Remove controller trait

allanbrondum:
## Purpose

Remove lock controller trait

## Changes

Removed the controller trait, effectively leaving one branch in the code. Persistent state and CBOR API still has an enum with a single variant.

## Checklist

      hard-to-understand areas.


soerenbf (approved):
Looks good!

plt/plt-scheduler/src/protocol_level_locks/p11.rs
```diff
@@ -220,6 +222,7 @@ fn execute_lock_fund<C: EntityContextTypes>(
         raw_amount,
         memo,
     )? {
+        // todo ar
```
soerenbf: What are these todos related to? could we have a more descriptive comment and also something describing when this will be fixed, potentially a linear issue if that exists 🙏

allanbrondum: I wanted to see if I could refactor the error handling to make it less verbose/noisy. But leaving it for another PR. Removed the todos since they were only there for that reason.

PR #1671, opened 2026-07-08
Add token chain parameters + authorizations

soerenbf:
Ref COR-2335
Depends on https://github.com/Concordium/concordium-base/pull/945

## Purpose

Adds structure for rust managed chain parameters + max lock duration chain parameter. Additionally, the authorizations structure to be used for the update.

`CHANGELOG.md` updated as part of subsequent PR: https://github.com/Concordium/concordium-node/pull/1672

## Changes

Adds a separate type for persistent state storage of chain parameters. This is due to changing the chain parameters from being a flat serialization to additionally have a pointer to token parameters managed by rust.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Parameters.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Parameters.hs
@@ -11,7 +11,7 @@
 -- This type is the persistent node representation used by the update state. It
 -- is distinct from the @concordium-base@ public/wire @ChainParameters'@ view.
 -- The aggregate public/wire type is only used at conversion boundaries; the
--- persistent storage model has its own record fields and a P11-and-onwards
+-- persistent storage model has its own record fields and a
 -- Rust-managed external chain-parameters pointer.
 module Concordium.GlobalState.Persistent.BlockState.Parameters (
     PersistentChainParameters,
@@ -23,7 +23,6 @@ module Concordium.GlobalState.Persistent.BlockState.Parameters (
 ) where
 
 import Control.Monad.IO.Class
-import Data.Bool.Singletons
 import qualified Data.ByteString as BS
 import qualified Data.Serialize as S
 import Data.Singletons
@@ -37,7 +36,7 @@ import Concordium.Types.HashableTo
 import Concordium.Types.Parameters
 
 -- | Persistent node-owned chain parameters.
-data PersistentChainParameters' cpv auv = PersistentChainParameters
+data PersistentChainParameters' (pv :: ProtocolVersion) cpv auv = PersistentChainParameters
     { -- | Consensus parameters.
       pcpConsensusParameters :: !(ConsensusParameters cpv),
       -- | Exchange rates.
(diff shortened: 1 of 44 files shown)


plt/plt-block-state/src/persistent/chain_parameters/p11.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// Node-owned persistent external chain parameters for P11.
+///
+/// This mirrors the parts of the public chain-parameter view whose authoritative
+/// state is managed outside the ordinary Haskell chain-parameter record.
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
+pub struct PersistentChainParametersP11 {
+    /// Maximum relative duration for protocol-level token locks, in milliseconds.
+    pub max_lock_duration: u64,
+}
```
allanbrondum, line 18: Do we wan't this chain parameter to be persisted by Rust. Seems like a lot of work?

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Parameters.hs
```diff
@@ -0,0 +1,259 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+-- | Node-owned persistent chain parameters.
+--
+-- This type is the persistent node representation used by the update state. It
+-- is distinct from the @concordium-base@ public/wire @ChainParameters'@ view.
+-- The aggregate public/wire type is only used at conversion boundaries; the
+-- persistent storage model has its own record fields and a P11-and-onwards
+-- Rust-managed external chain-parameters pointer.
```
limemloh: Likewise here, I think it is better not to say `P11 and onwards` since this might be older version as well at some point.
```suggestion
-- persistent storage model has its own record fields and a
-- Rust-managed external chain-parameters pointer.
```

plt/plt-block-state/src/ffi/external_chain_parameters.rs
```diff
@@ -0,0 +1,232 @@
...
+    });
+    if let Some(message) = panic_message {
+        eprintln!("{}", message);
+        status::FfiStatusCode::Panic
+    } else {
+        status::FfiStatusCode::Success
+    }
+}
+
+/// Allocate new P11 external chain parameters with an initial maximum lock duration.
```
limemloh:
```suggestion
/// Allocate new external chain parameters with an initial maximum lock duration.
```

limemloh (approved):

plt/plt-block-state/src/persistent/chain_parameters/p11.rs
soerenbf: Discussed elsewhere, but this has been decided to be the way forward.

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
```diff
@@ -661,13 +662,13 @@ makeBasicPendingUpdates PendingUpdates{..} = withCPVConstraints (chainParameters
     return UQ.PendingUpdates{..}
 
 -- | Current state of updatable parameters and update queues.
-data Updates' (cpv :: ChainParametersVersion) (auv :: AuthorizationsVersion) = Updates
+data Updates' (pv :: ProtocolVersion) (cpv :: ChainParametersVersion) (auv :: AuthorizationsVersion) = Updates
```
soerenbf, line 665: @limemloh I feel like this makes the `Updates'` redundant as cpv and auv in the type are redundant.. I did not want to remove it as maybe I'm missing something and the change would also be bigger that what I'd like..

The reason why it needs to carry the protocol version information is that the external chain params ptr is parameterised by it to be able to load the rust chain parameters corresponding to the protocol version. We could also pass an authorizations version but given the fact that the authorizations for all token parameter updates is the same, any future token parameter would need another way to differentiate anyway.

Might be a little confusing.. let's discuss if needed.

limemloh (approved):

concordium-consensus/src/Concordium/GlobalState/Persistent/BlockState/Updates.hs
limemloh: I think we should remove the other type parameters, since it can be derived, but also such that we don't have to handle the impossible cases of mismatching versions.
But I imagine it affects a lot of places, so maybe in a separate PR for reviewability

soerenbf: Great, will do this further down the PR stack and also consolidate the identical types

PR #1673, opened 2026-07-13
Ar/cor 2441 refactor code module structure for locks scheduler error handling

allanbrondum:
## Purpose

Simplify/clean up error handling in the scheduler code

## Changes

* pushed `TransactionRejectReason` down into token module update function signatures and removed the `TokenUpdateError` "error hierarchy" (internally in the module, the error `TokenUpdateErrorInternal` is still used, we can consider also removing that later, but it is internal only now)
* moved creation of token module reject reasons to `token_module::reject`
* used `ResultWithBlockStateFailure` more places to make error handling more fluent

## Checklist

      hard-to-understand areas.


plt/plt-scheduler/src/protocol_level_locks/p11.rs
```diff
@@ -565,82 +551,3 @@ fn lock_configuration_keeps_alive(configuration: &LockConfiguration) -> bool {
         LockControllerConfig::SimpleV0(controller) => controller.keep_alive,
     }
 }
-
-fn parse_raw_amount(
```
allanbrondum, line 569: Using util in `token_module::token_amount` insetead

plt/plt-scheduler/src/protocol_level_locks/p11.rs
```diff
@@ -565,82 +551,3 @@ fn lock_configuration_keeps_alive(configuration: &LockConfiguration) -> bool {
...
-                reason_type: reason_type.to_type_discriminator(),
-                details: Some(details),
-            })
-        }
-    }
-}
-
-fn token_balance_insufficient_reject_reason(
-    operation_index: usize,
-    token_configuration: &TokenConfiguration,
```
allanbrondum, line 627: Using `token_module::reject` now

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
```diff
@@ -105,125 +90,76 @@ pub fn execute_token_update_operation_at_index<C: EntityContextTypes>(
     mut token: TokenPXRefMut<'_>,
     index: usize,
     operation: &TokenOperation,
-) -> BlockStateResult<Result<(), TokenUpdateError>> {
-    let int_err = match execute_token_update_operation_internal(
+) -> ResultWithBlockStateFailure<(), TransactionRejectReason> {
```
allanbrondum, line 93: Changed to use `TransactionRejectReason` as error type

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
```diff
@@ -105,125 +90,76 @@ pub fn execute_token_update_operation_at_index<C: EntityContextTypes>(
...
-/// ```ignore
-/// let external = token_update_error_internal_to_external(
-///     &token_configuration,
-///     0,
-///     "transfer",
-///     TokenUpdateErrorInternal::Paused,
-/// )?;
-/// ```
-pub(crate) fn token_update_error_internal_to_external(
+fn token_update_error_internal_to_external(
```
limemloh, line 117: Maybe not for now, but I think we should remove these internal errors and just have one big scheduler error. In the end we map all of it to one result anyway.

plt/plt-scheduler/src/failure.rs
limemloh, line 9: I think it would be more elegant to just have a single error representation at the top which can nest transaction reject reason in a single variant and probably have variants for blockstate and other scheduler errors.
```rust
pub enum SchedulerError {
    TransactionRejection(TransactionRejectReason),
    BlockStateFailure(BlockStateFailure),
    UnexpectedPayload,
    ...
}
```
Then the top level call to the scheduler can map out whether it is a rejected or something unexpected and provide the transaction outcome.
It means we don't have to these nested result and uses the usual rust conversion for most of it.

I think we should merge your PR though

limemloh (approved):

plt/plt-scheduler/src/failure.rs
allanbrondum: I agree that we can probably replace `WithBlockStateFailure<TransactionRejectReason>` with `SchedulerError`, I think the PR moves the code in a direction, that actually makes this a fair amount easier than before the PR. There are also some lower level errors wrapped in `WithBlockStateFailure`, so we may want to keep using `WithBlockStateFailure` some places.

plt/plt-scheduler/src/protocol_level_tokens/token_module/update.rs
allanbrondum: I agree that we can remove `TokenUpdateErrorInternal`, after this PR, is should be a change affecting `update` module only.

PR #1675, opened 2026-07-17
Use PacketDestination as argument and remove call indirections

limemloh:
## Purpose
- Simplify the network messaging layer by passing `PacketDestination` directly instead of routing through direct/broadcast wrapper helpers.
- Reduce call indirection around consensus and connectivity message sending.

No semantic changes

## Changes
- Replace separate direct/broadcast send helpers with `send_message_over_network`.
- Pass `PacketDestination` explicitly from consensus and connection call sites.
- Simplify consensus send logic and logging around message destinations.
- Update connection tests to use the new messaging API.

### DIFF
--- a/concordium-node/src/p2p/connectivity.rs
+++ b/concordium-node/src/p2p/connectivity.rs
@@ -722,42 +722,13 @@ fn is_valid_broadcast_target(
         && conn.remote_end_networks.contains(&network_id)
 }
 
-/// Send a direct packet with `msg` contents to the specified peer.
 #[inline]
-pub fn send_direct_message(
+pub fn send_message_over_network(
     node: &P2PNode,
-    target_id: RemotePeerId,
-    network_id: NetworkId,
-    msg: Arc<[u8]>,
-) -> usize {
-    send_message_over_network(node, Some(target_id), vec![], network_id, msg)
-}
-
-/// Send a broadcast packet with `msg` contents to the specified peer.
-#[inline]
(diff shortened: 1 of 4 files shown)


concordium-node/src/p2p/connectivity.rs
```diff
@@ -722,42 +722,13 @@ fn is_valid_broadcast_target(
...
-    msg: Arc<[u8]>,
-) -> usize {
-    send_message_over_network(node, None, dont_relay_to, network_id, msg)
-}
-
-#[inline]
-fn send_message_over_network(
-    node: &P2PNode,
-    target_id: Option<RemotePeerId>,
-    dont_relay_to: Vec<RemotePeerId>,
```
limemloh, line 751: These two arguments are essentially a flat `PacketDestination`, so all changes stem from using the actual type

concordium-node/src/plugins/consensus.rs
```diff
@@ -430,28 +428,23 @@ fn send_msg_to_consensus(
 
 fn send_consensus_msg_to_net(
     node: &P2PNode,
-    dont_relay_to: Vec<RemotePeerId>,
-    target_id: Option<RemotePeerId>,
-    (payload, msg_desc): (Arc<[u8]>, PacketType),
```
limemloh, line 435: Also snuck in unwrapping of this pair in the argument

allanbrondum (approved):

PR #1678, opened 2026-08-04
Max lock duration execution

soerenbf:
Ref COR-2335

Depends on:

- https://github.com/Concordium/concordium-grpc-api/pull/123
- https://github.com/Concordium/concordium-base/pull/950
- https://github.com/Concordium/concordium-base/pull/953

## Purpose

Enforce max lock duration during lock creation

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -93,13 +93,13 @@ import Concordium.GlobalState.ContractStateFFIHelpers (LoadCallback)
 import qualified Concordium.GlobalState.ContractStateV1 as StateV1
 import Concordium.GlobalState.CooldownQueue (Cooldowns)
 import qualified Concordium.GlobalState.Persistent.Account.ProtocolLevelTokens as GSAccount
+import qualified Concordium.GlobalState.Persistent.BlockState.ExternalChainParameters as ECP
 import Concordium.GlobalState.Persistent.BlockState.ProtocolLevelTokens (
     PLTConfiguration,
     TokenIndex,
     TokenStateKey,
     TokenStateValue,
  )
-import qualified Concordium.GlobalState.Persistent.BlockState.ExternalChainParameters as ECP
 import Concordium.GlobalState.Persistent.BlockState.ProtocolLevelTokens.RustPLTBlockState as RustBS
 import Concordium.GlobalState.Persistent.LMDB (FixedSizeSerialization)
 import Concordium.GlobalState.TransactionTable (TransactionTable)
--- a/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
+++ b/concordium-consensus/src/Concordium/GlobalState/BlockState.hs
@@ -1772,12 +1772,12 @@ class
         (PVSupportsRustManagedPLT (MPV m)) =>
         UpdatableBlockState m -> m (RustBS.ForeignPLTBlockStatePtr (MPV m))
 
-    -- | Get the node-owned Rust external chain-parameters pointer, if the
-    -- current protocol version has external chain parameters.
+    -- | Get the node-owned Rust external chain-parameters pointer.
     --
     -- This is a low-level interface needed for foreign function interface access.
     bsoGetExternalChainParameters ::
-        UpdatableBlockState m -> m (Maybe ECP.ForeignExternalChainParametersPtr)
+        (PVSupportsRustManagedECP (MPV m)) =>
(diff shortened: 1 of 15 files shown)


concordium-consensus/src/Concordium/GlobalState/BlockState.hs
```diff
@@ -1771,6 +1772,13 @@ class
         (PVSupportsRustManagedPLT (MPV m)) =>
         UpdatableBlockState m -> m (RustBS.ForeignPLTBlockStatePtr (MPV m))
 
+    -- | Get the node-owned Rust external chain-parameters pointer, if the
+    -- current protocol version has external chain parameters.
+    --
+    -- This is a low-level interface needed for foreign function interface access.
+    bsoGetExternalChainParameters ::
```
limemloh, line 1778: Just want to point out, this is slightly different approach than we usually do, whether it is present becomes a runtime check now and usually this is kept compile-time checked (like `bsoGetRustPLTBlockState` have the constraint `(PVSupportsRustManagedPLT (MPV m))`).
I am not saying we should change it, since it is a valid choice as it simplifies things on the type-level, so just adding this to ensure it was deliberate choice.

plt/plt-scheduler/src/protocol_level_locks/p11.rs
```diff
@@ -560,6 +580,15 @@ fn remove_lock_balance_ref<C: EntityContextTypes>(
     Ok(())
 }
 
+fn expiry_exceeds_max_duration(
+    expiry_seconds: u64,
+    block_timestamp_millis: u64,
+    max_lock_duration: u64,
+) -> bool {
+    u128::from(expiry_seconds) * 1000
```
limemloh: Nitpicking here, but instead of doing arithmetic in `u128` could we just keep it in u64 by computing the duration between expiry and block time, then compare that to max duration.
Also, not sure if this justifies a separate function

plt/plt-scheduler/src/protocol_level_locks/p11.rs
```diff
@@ -137,6 +137,7 @@ pub fn execute_lock_operation<C: EntityContextTypes>(
     context: &mut EntityContext<C>,
     transaction_execution: &mut TransactionExecution,
     block_state: &mut BlockStateP11,
+    max_lock_duration: u64,
```
limemloh: Maybe we can use a type wrapper here, we already have `SlotDuration` in `concordium-base` which matches.

plt/plt-scheduler/tests/lock_create.rs
```diff
@@ -17,13 +17,56 @@ use concordium_base::{
...
+use plt_block_state::entity::accounts::Account;
 use plt_block_state::entity::entity_test_stub;
+use plt_block_state::persistent::chain_parameters::p11::PersistentChainParametersP11;
 use plt_scheduler::TOKEN_MODULE_REF;
 use plt_scheduler_types::types::events::{BlockItemEvent, LockCreateEvent};
 use std::collections::HashMap;
 
 mod utils;
 
+macro_rules! execute_lock_create_with_duration {
```
limemloh: Why using a macro here? it seems this could just be a function

limemloh (approved):

plt/plt-scheduler/tests/lock_create.rs
soerenbf: Did we not have a discussion in another PR where we agreed to avoid assertions in helper functions? I cannot remember which PR, but that is the reason why I structured it like this instead of a helper function 😅 Happy to do it as a helper function instead

soerenbf: There used to be assertions inside this, but they're gone so you're right, there's no reason this is not just a function. Will revise.

plt/plt-scheduler/src/protocol_level_locks/p11.rs
soerenbf: Sure, that also seems to match the corresponding `Duration` type in haskell.

concordium-consensus/src/Concordium/GlobalState/BlockState.hs
soerenbf: It was not, thanks for pointing this out. I'll see what I can do to align 😊

soerenbf: Added a corresponding type level constraint, depends on the changes implemented in https://github.com/Concordium/concordium-base/pull/953.

PR #1682, opened 2026-08-25
Fix p10 to p11 plt migration

soerenbf:
## Purpose

Fix an error in the PLT state migration caused by a double dereference of the haskell managed PLT blob store reference

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
+++ b/concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
@@ -11,6 +11,7 @@ module Concordium.Scheduler.ProtocolLevelTokens.RustPLTScheduler.Queries (
     SerializedLockId,
     queryPLTList,
     queryTokenInfo,
+    queryTokenInfoInBlobStoreMonad,
     queryTokenAccountInfos,
     queryTokenAuthorizations,
     queryLockList,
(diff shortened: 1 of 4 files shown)


concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/RustPLTScheduler/Queries.hs
```diff
@@ -164,75 +165,79 @@ queryTokenInfo ::
 queryTokenInfo bs tokenId = do
     queryCallbacks <- unliftBlockStateQueryCallbacks bs
     pltState <- BS.getRustPLTBlockState bs
-    BS.liftBlobStore $ queryTokenInfoInBlobStoreMonad pltState queryCallbacks
-  where
-    -- Get token info for the given token in the given block state. The function is a wrapper around an FFI call
-    -- to the Rust PLT Scheduler library.
-    queryTokenInfoInBlobStoreMonad ::
```
allanbrondum, line 171: I think it makes it a bit hard to review the diff when this function is e.g. moved to be a top level function. Is the refactor needed?

soerenbf: It's refactored as it's used in the unit test proving correctness of the migrated state. It was either this or adding a re-declaration of the FFI function for querying the state. The doc-comment for the function should explain why it's exposed?

allanbrondum: Ok, it just want clear what had changed, if anything

allanbrondum (approved):

PR #1684, opened 2026-08-26
Dedupe lock grant roles when stored

soerenbf:
## Purpose

Ensure lock grant roles don't contain duplicates

## Checklist

      hard-to-understand areas.


plt/plt-scheduler/src/protocol_level_locks/lock_controller.rs
```diff
@@ -118,7 +134,7 @@ pub fn from_cbor_controller<C: EntityContextTypes>(
 
             Ok(LockControllerSimpleV0Grant {
                 account: account.account_index(),
-                roles: grant.roles,
+                roles: canonicalize_roles(grant.roles),
             })
```
allanbrondum: Suggestion: make fields private and canonicalize in a `LockControllerSimpleV0Grant::new` function

plt/plt-scheduler/tests/lock_create.rs
```diff
@@ -77,7 +77,7 @@ fn execute_lock_create_with_duration(
 }
 
 #[test]
-fn test_create_simple_lock() {
```
allanbrondum, line 80: Shouldn't the original test still be there or at least preserve its name that describes its intend to test creating a lock and not just testing canonicalization of roles?

allanbrondum (approved):

PR #1686, opened 2026-09-03
Refactor persistent locks state to trie

soerenbf:
Disclaimer: co-authored by AI, heavily reviewed and revised.

The implementation is superseded by #1704. The PR will be closed when this is merged.

## Purpose

Refactor locks persistence to trie to support efficient historical lookup

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
+++ b/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
@@ -8,7 +8,7 @@ use crate::persistent::protocol_level_tokens::p9::TokenIndex;
 use crate::persistent::smart_contract_trie::PersistentState;
 use concordium_base::base::AccountIndex;
 use concordium_base::common::types::TransactionTime;
-use concordium_base::common::{Buffer, Serialize};
+use concordium_base::common::{Buffer, Serialize, from_bytes_complete, to_bytes};
 use concordium_base::hashes::Hash;
 use concordium_base::protocol_level_locks::{LockControllerSimpleV0Capability, LockId};
 use concordium_base::protocol_level_tokens::{CborMemo, RawCbor, TokenId};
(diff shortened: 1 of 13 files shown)


plt/plt-block-state/src/persistent/smart_contract_trie.rs
soerenbf, line 1: Don't know whether it's interesting or not, but would be nice to be able to define specific key/value types contained in the trie structure with a constraint that both implement `Serialize` and `Deserialize`.. is that not feasible, or how come we dont? I realize we've only used it to store arbitrary bytes up until this point.

allanbrondum (commented):
I think there several changes, and it is not fully clear to me, what is changed why. If we just thaw/freeze the trie when creating and updating a lock, I think we can do which much fewer changes. I'm not sure if that is a performance issue. But I would maybe start with that smaller change. And then if we think there is a performance issue, I would choose the arhitectural change of the types (like having a total mutable locks entity state, but there may be other options) by hand.

soerenbf:
> I think there several changes, and it is not fully clear to me, what is changed why.

These are the layers as I see them:

1. change the data structure used for persistency to avoid loading all locks to query the details of one
2. support mutating state across the scope of a transaction instead of in the scope of individual lock operations (block state interface change, `LocksP11`)

> If we just thaw/freeze the trie when creating and updating a lock, I think we can do which much fewer changes. I'm not sure if that is a performance issue. But I would maybe start with that smaller change. And then if we think there is a performance issue, I would choose the architectural change of the types (like having a total mutable locks entity state, but there may be other options) by hand.

I deliberately went for this solution, weighing the choice between thaw/freeze every operation and this 😅. It's a bigger change and I'm fine with going the other route if that's what we want.

The reason why I went this route in the first place is because I think it matches the nature of the transaction it executes nicely, i.e. execute a sequence of operations as a batch (which is also what we do for token updates, one could make the same argument here?). I might be able to make the change smaller by removing the `LocksP11` type and just store the mutable locks state directly in the block state on a private field.. I thought it was nicer to create a dedicated wrapper type to capture state mutation for this specific entity.

soerenbf:
Benchmarks revealed that thaw/freeze is not that expensive and not worth the increased complexity. Will refactor to the simple proposed solution.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -97,7 +67,7 @@ pub struct PersistentLockP11 {
     /// Note the entire collection will be written to disk every time this struct is written to disk.
     pub locked_balances: StoreSerialized<BTreeSet<(AccountIndex, TokenIndex)>>,
     /// The configuration parameters for the lock.
-    pub configuration: HashedCacheableRef<StoreSerialized<LockConfiguration>>,
```
allanbrondum, line 100: Why is the reference removed? The idea is to have the value that does not change behind a reference, such that they are not rewritten to the block state when the mutable values change.

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -1,73 +1,43 @@
...
-    ///
-    /// Here `None` represents a lock which has been deleted and acts as tombstone, preventing new
-    /// locks from using the same index.
-    pub(crate) locks: LfmbTree<LockIndex, Option<PersistentLockP11>>,
-    /// Index for mapping Lock ID to the internal lock index used above.
-    ///
-    /// Deleted locks are absent from this index.
-    pub(crate) lock_id_map: im::HashMap<LockId, LockIndex>,
+    /// Persistent trie of serialized [`LockId`] keys and [`PersistentLockP11`] values.
+    pub(crate) locks: PersistentState,
```
allanbrondum, line 20: I would kind of think that the values should be references to `PersistentLockP11` and not `PersistentLockP11` values. Well, don't know how difficult that would be. Generally, I think this is just another reason we should write a new trie dedicated for use in the Rust block state that is generic over keys and values.

soerenbf: Took some digging, but this is already part of the SC trie implementation, i.e. values are stored either inline or as a hashed reference depending on the size of the serialized value. Anything > 64 bytes is written as a reference. Follow this function: https://github.com/Concordium/concordium-base/blob/eb31fa0102c849febc0f409c6b9bfa446c181182/smart-contracts/wasm-chain-integration/src/v1/trie/low_level.rs#L1389

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
soerenbf: Cannot use the `HashedCachableRef` for anything inside the trie as it uses the raw bytes representation of any value. So if we were to use references here, those would have the be serialized and would thus be included in the hash which would make it non-deterministic, as now the hash depends on the disk location (as far as I understand😅).

I'm afraid we'll have to use a dedicated data structure if we want to optimise the storage of the values.

soerenbf: It also occurred to me that `StoreSerialized` does nothing here as well, so will also remove that.

allanbrondum: Yes, you are right, the hashing will be an issue if we store references. Another option would be to replace the in memory id to index map with a persisted trie, and still keep the left-full index tree. Maybe we should discuss the different options when Emil is back?

soerenbf: I'll try to conduct some form of analysis on the storage cost for both approaches 🙂

soerenbf: The reference should be reintroduced when #1696 is compete

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
soerenbf: This will be refactored to `Trie<LockId, PersistentLockP11>` once #1696 is merged

PR #1689, opened 2026-09-08
Fix macos release

soerenbf:

PR #1692, opened 2026-09-09
Pin macos runner image (node version 10.0.10)

soerenbf:
Ref COR-2633

## Purpose

See https://linear.app/concordium/issue/COR-2633/macos-node-package-fails-to-start-due-to-duplicate-lc-rpath-entries for context.


.github/workflows/release.yaml
soerenbf, line 1: Unrelated changes are pulled in from https://github.com/Concordium/concordium-node/pull/1689. These are needed to pin the lmdb version as the latest version (1.0.1) contains breaking changes.

Ivan-Mahda (approved):

PR #1693, opened 2026-09-09
Pin macos runner image (node version 10.0.11)

soerenbf:
Ref COR-2633

## Purpose

See https://linear.app/concordium/issue/COR-2633/macos-node-package-fails-to-start-due-to-duplicate-lc-rpath-entries for context.


.github/workflows/release.yaml
soerenbf, line 1: Don't know what's up with the diff here, but most of the changes are already in `release/10.0.11`. This should be identical to #1691

Ivan-Mahda (approved):

PR #1696, opened 2026-09-11
Block state trie

allanbrondum:
## Purpose

Implement dedicated trie for the Rust block state.

## Changes

## Checklist

      hard-to-understand areas.


plt/plt-block-state/proptest-regressions/persistent/trie.txt
soerenbf, line 1: Good idea

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -0,0 +1,1676 @@
...
+        &self,
+        loader: &impl BlobStoreLoad,
+        path: &[u8],
+        value: V,
+    ) -> BlockStateResult<(Node<INLINE_KEY_LENGTH, V>, bool)>
+    where
+        V: Loadable + Clone,
+    {
+        Ok(if let Some(&path_byte) = path.first() {
+            if let Some(edge) = self.children.get(path_byte) {
```
soerenbf, line 509: would be good to invert these to do early returns instead. the else cases get lost in the mountain of code here, and would also reduce nesting 🙂

soerenbf (approved):
Found a few things that I fixed here:

https://github.com/Concordium/concordium-node/pull/1699

Inverted a few conditionals and asserted a property that could fail due to state corruption

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -0,0 +1,1676 @@
...
+/// * Delete an entry with [`Trie::delete_entry`]: Returns the new trie with the entry removed.
+///
+/// ## Interior mutability
+///
+/// The internal representation in the tree may change during the lifetime via interior mutability.
+/// This happens if values are cached, stored or hashes are lazily calculated.
+///
+/// ## Data structure and invariants
+///
+/// The data structure i a compact trie. The stems are maximal, which means that each
```
limemloh:
```suggestion
/// The data structure is a compact trie. The stems are maximal, which means that each
```

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -0,0 +1,1676 @@
...
+/// node either has a value, or at lest two children (except for the root node)
+/// ```
+#[derive(Debug)]
+pub struct Trie<const INLINE_KEY_LENGTH: usize, K, V> {
```
limemloh: Do we need `INLINE_KEY_LENGTH` to be a generic? or could it be a const?

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -0,0 +1,1676 @@
...
+
+        None
+    }
+}
+
+/// Trie node
+#[derive(Debug)]
+struct Node<const INLINE_KEY_LENGTH: usize, V> {
+    children: ChildEdges<INLINE_KEY_LENGTH, V>,
+    stem: TinyVec<[u8; INLINE_KEY_LENGTH]>,
```
limemloh: Did you consider nibbles instead of bytes (4-bit chunks) like we do in the smart contract state trie?

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -0,0 +1,1676 @@
...
+}
+
+// todo ar change hash implementation, make hash take a digest instead of returning hash
+
+impl<const INLINE_KEY_LENGTH: usize, K, V: Hashable + Loadable> Hashable
+    for Trie<INLINE_KEY_LENGTH, K, V>
+{
+    fn hash(&self, loader: &impl BlobStoreLoad) -> BlockStateResult<Hash> {
+        Ok(hash::hash_of_hashes(
+            StoreSerialized(self.size).hash(loader)?,
```
limemloh: Seems a bit overkill to do a full hash for the size, consider just serializing it

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -0,0 +1,1676 @@
...
+        path: &[u8],
+        value: V,
+    ) -> BlockStateResult<(Node<INLINE_KEY_LENGTH, V>, bool)>
+    where
+        V: Loadable + Clone,
+    {
+        Ok(if let Some(&path_byte) = path.first() {
+            if let Some(edge) = self.children.get(path_byte) {
+                let child_node = edge.child_ref.value(loader)?;
+                let common_prefix_len = common_prefix(&path[1..], &child_node.stem[1..]).len() + 1;
```
limemloh: why are we storing the branching byte in the stem of the child?

plt/plt-block-state/src/persistent/trie.rs
allanbrondum: Done

allanbrondum:
> Found a few things that I fixed here:
> 
> #1699
> 
> Inverted a few conditionals and asserted a property that could fail due to state corruption

Addressed refactor here in this pr. Will have a separate PR to remove the first byte from the stem

PR #1699, opened 2026-09-15
Fix a few edge cases + corruption case

soerenbf:
The only logical change added here is validation of child stems with `validate_child_stem`, which checks for state corruption. 

Other than that, most of the changes aim to increase readability of the code by inverting a couple of conditionals.


allanbrondum:
Addressed on https://github.com/Concordium/concordium-node/pull/1696#issuecomment-5717080712

PR #1700, opened 2026-09-18
Update hashing of trie nodes

allanbrondum:
## Purpose

Change hashing of trie nodes to be similar to smart contract trie

## Changes

## Checklist

      hard-to-understand areas.


plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -899,24 +898,28 @@ impl<const INLINE_KEY_LENGTH: usize, V: Loadable> Loadable for Node<INLINE_KEY_L
...
     ) -> BlockStateResult<Self> {
+        let value = Loadable::load_from_buffer(&mut buffer, loader)?;
+        let stem = <StoreSerialized<TinyVecDeserial<INLINE_KEY_LENGTH>>>::load_from_buffer(
+            &mut buffer,
+            loader,
+        )?
+        .0
+        .0;
+        let children = Loadable::load_from_buffer(&mut buffer, loader)?;
+
```
allanbrondum, line 909: Just reordered to match hashing order

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -951,48 +954,36 @@ impl<const INLINE_KEY_LENGTH: usize, V: Storable> Storable for ChildEdges<INLINE
...
-            child_ref: target_ref,
-        })
-    }
-}
-
-impl<const INLINE_KEY_LENGTH: usize, V: Storable> Storable for Edge<INLINE_KEY_LENGTH, V> {
-    fn store_to_buffer(&self, mut buffer: impl Buffer, storer: &mut impl BlobStoreStore) {
-        self.child_ref.store_to_buffer(&mut buffer, storer);
-    }
-}
```
allanbrondum, line 975: Remove `Edge` implementations to make things simpler

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -951,48 +954,36 @@ impl<const INLINE_KEY_LENGTH: usize, V: Storable> Storable for ChildEdges<INLINE
...
 {
     fn hash(&self, loader: &impl BlobStoreLoad) -> BlockStateResult<Hash> {
-        Ok(hash::hash_of_hashes(
-            StoreSerialized(self.size).hash(loader)?,
-            self.root.hash(loader)?,
-        ))
+        let mut hasher = sha2::Sha256::new();
+        hasher.update(self.size.to_be_bytes());
+        hasher.update(self.root.hash(loader)?);
+        Ok(Hash::new(hasher.finalize().into()))
```
allanbrondum, line 969: This top level hashing is not identical to smart contract trie, we can adjust it later if we want it to be identical to
```
    pub fn hash(&self, loader: &mut impl BackingStoreLoad) -> super::Hash {
        match self {
            PersistentState::Empty => {
                // hash of the node starts with either a 0 or 1 byte. This makes it distinct,
                // but is otherwise an arbitrary choice.
                super::Hash::from(<[u8; 32]>::from(sha2::Sha256::digest(
                    b"empty contract state",
                )))
            }
            PersistentState::Root(root) => root.hash(loader),
        }
    }
```

plt/plt-block-state/src/persistent/trie.rs
```diff
@@ -951,48 +954,36 @@ impl<const INLINE_KEY_LENGTH: usize, V: Storable> Storable for ChildEdges<INLINE
...
     }
 }
 
 impl<const INLINE_KEY_LENGTH: usize, V: Hashable + Loadable> Hashable
     for Node<INLINE_KEY_LENGTH, V>
```
allanbrondum, line 974: Node hashing is identical to smart contract trie except for
1) 8-bit branching instead of 4-bit
2) discriminator byte is included in the stem

limemloh (approved):

PR #1704, opened 2026-09-21
Implement locks persistent state as block state trie

soerenbf:
Changes the persistent locks state implemented in https://github.com/Concordium/concordium-node/pull/1686 to the block state trie implemented in https://github.com/Concordium/concordium-node/pull/1696.

Depends on https://github.com/Concordium/concordium-base/pull/968

## Benchmark comparison

The benchmark median changes (benches implemented in #1703):
```
 ┌─────────────────────────────┬──────────┬──────────┬────────┐
 │ Benchmark                   │ Previous │ Trie     │ Change │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Top-level lock create, 10   │ 6.291 µs │ 1.812 µs │ −71%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Top-level lock create, 100  │ 11.56 µs │ 4.999 µs │ −57%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Top-level lock create, 1000 │ 49.41 µs │ 37.14 µs │ −25%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Scheduler lock create, 10   │ 59.54 µs │ 18.99 µs │ −68%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Scheduler lock create, 100  │ 645.4 µs │ 151.6 µs │ −77%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Scheduler lock create, 1000 │ 9.609 ms │ 1.505 ms │ −84%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Fund absent reference       │ 6.374 µs │ 5.832 µs │ −9%    │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Release draining reference  │ 5.853 µs │ 4.957 µs │ −15%   │
 ├─────────────────────────────┼──────────┼──────────┼────────┤
 │ Send draining reference     │ 6.874 µs │ 5.833 µs │ −15%   │
 └─────────────────────────────┴──────────┴──────────┴────────┘
```
 Cancellation at large sizes regressed slightly, but nothing huge:

 - Many accounts/many tokens, 1000: +3%
 - Many accounts/one token, 1000: +4%
 - One account/many tokens, 1000: +3%
 - One account/many tokens, 10: +10%

Previous implementation of the balance references that are a factor in this was a `BTreeSet` and is now also a `Trie`, meaning that updating this is more efficient, but iterating (as part of cancellation) probably is not 🤷)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
+++ b/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
@@ -8,7 +8,7 @@ use crate::persistent::protocol_level_tokens::p9::TokenIndex;
 use crate::persistent::smart_contract_trie::PersistentState;
 use concordium_base::base::AccountIndex;
 use concordium_base::common::types::TransactionTime;
-use concordium_base::common::{Buffer, Serialize};
+use concordium_base::common::{Buffer, Serialize, from_bytes_complete, to_bytes};
 use concordium_base::hashes::Hash;
 use concordium_base::protocol_level_locks::{LockControllerSimpleV0Capability, LockId};
 use concordium_base::protocol_level_tokens::{CborMemo, RawCbor, TokenId};
@@ -107,33 +107,28 @@ impl Hashable for PersistentLockP11 {
 
 /// Serialize a lock ID for use as a persistent trie key.
 pub(crate) fn lock_id_key(lock_id: &LockId) -> Vec<u8> {
-    concordium_base::common::to_bytes(lock_id)
+    to_bytes(lock_id)
 }
 
 /// Decode a canonical serialized lock ID from a persistent trie key.
 pub(crate) fn lock_id_from_key(key: &[u8]) -> BlockStateResult<LockId> {
-    let lock_id = concordium_base::common::from_bytes_complete(key).map_err(|err| {
+    from_bytes_complete(key).map_err(|err| {
         BlockStateFailure::BlobStoreDecode(format!("Stored lock ID key cannot be decoded: {err}"))
-    })?;
-    Ok(lock_id)
+    })
 }
 
 /// Serialize a persistent lock for use as a persistent trie value.
 pub(crate) fn persistent_lock_value(lock: &PersistentLockP11) -> Vec<u8> {
-    concordium_base::common::to_bytes(&(
-        lock.locked_balances.0.clone(),
-        lock.configuration.0.clone(),
-    ))
+    to_bytes(&(lock.locked_balances.0.clone(), lock.configuration.0.clone()))
 }
 
 /// Decode a persistent lock from a persistent trie value.
 pub(crate) fn persistent_lock_from_value(value: &[u8]) -> BlockStateResult<PersistentLockP11> {
-    let (locked_balances, configuration) = concordium_base::common::from_bytes_complete(value)
-        .map_err(|err| {
-            BlockStateFailure::BlobStoreDecode(format!(
-                "Stored persistent lock value cannot be decoded: {err}"
-            ))
-        })?;
+    let (locked_balances, configuration) = from_bytes_complete(value).map_err(|err| {
+        BlockStateFailure::BlobStoreDecode(format!(
+            "Stored persistent lock value cannot be decoded: {err}"
+        ))
+    })?;
     Ok(PersistentLockP11 {
         locked_balances: StoreSerialized(locked_balances),
         configuration: StoreSerialized(configuration),
--- a/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
+++ b/plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
@@ -1,12 +1,11 @@
 use crate::failure::{BlockStateFailure, BlockStateResult};
-use crate::persistent::blob_reference::hashed_cacheable_reference::HashedCacheableRef;
 use crate::persistent::blob_store::{
     BlobStoreLoad, BlobStoreStore, Loadable, Storable, StoreSerialized,
 };
 use crate::persistent::cacheable::Cacheable;
 use crate::persistent::hash::{self, Hashable};
-use crate::persistent::lfmb_tree::{LfmbTree, LfmbTreeKey};
 use crate::persistent::protocol_level_tokens::p9::TokenIndex;
+use crate::persistent::smart_contract_trie::PersistentState;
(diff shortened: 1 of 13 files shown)


plt/plt-scheduler/benches/block_state_size.rs
soerenbf, line 1: From #1703

plt/plt-scheduler/benches/scheduler.rs
soerenbf, line 1: From #1703

plt/plt-scheduler/Cargo.toml
soerenbf, line 1: From #1703

concordium-consensus/benchmarks/transactions/TransactionsBench.hs
soerenbf, line 1: From #1703

plt/plt-block-state/src/persistent/smart_contract_trie.rs
soerenbf, line 1: From #1686

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -1,116 +1,78 @@
...
-            let Some(lock) = lock.as_ref() else {
-                continue;
-            };
-            let conf = lock.configuration.value(loader)?;
-            lock_id_map.insert(conf.0.lock_id.clone(), lock_index);
-        }
-        Ok(Self { locks, lock_id_map })
+impl TrieKey for LockId {
+    fn to_bytes(&self) -> impl std::borrow::Borrow<[u8]> {
+        to_bytes(self)
```
allanbrondum: The `TrieKey` trait is designed such that you should not need allocating. Serializing will allocate and should not be needed. You can return a `[u8; 3 * 8]` I think? (have a local mutable variable of the type and then write the be_bytes representations of the 3 `u64`)

plt/plt-block-state/src/persistent/protocol_level_locks/p11.rs
```diff
@@ -1,116 +1,78 @@
...
-/// in the API, events or used in state hashing.
-#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize)]
-pub struct LockIndex(pub u64);
-
-impl LfmbTreeKey for LockIndex {
-    fn to_u64(self) -> u64 {
-        self.0
-    }
+/// Inline capacity matching one serialized `u64` key component.
+const TRIE_KEY_INLINE_LENGTH: usize = 8;
```
allanbrondum: When the keys are fixed length, I think we should inline the entire key length: 2 * 8 for balance references and 3 * 8 for lock id.

plt/plt-block-state/src/persistent/smart_contract_trie.rs
```diff
@@ -56,6 +62,26 @@ impl PersistentState {
...
+            BlockStateFailure::Invariant(format!("Error iterating keys in MutableTrie: {err}"))
+        })?;
+
+        Ok(KeyPrefixIterator {
+            loader,
+            trie: trie.clone(),
+            trie_iter,
+        })
+    }
+
```
allanbrondum: I assume this is dead code now? Just so we don't have to review it (there has been subtle bugs interacting with the smart contract trie, so would have to look deeper into the code to see if it correct)

plt/plt-scheduler/benches/block_state_size.rs
```diff
@@ -0,0 +1,423 @@
+//! Scheduler benchmarks isolating block-state size from transaction operation count.
+//!
+//! Each case prepares increasingly complex state outside Divan's timed closure, then executes one
+//! successful operation. Separate dimensions avoid a cross-product: top-level token count measures
+//! one mint, allow-list size measures one insertion, and top-level lock count measures one creation.
```
allanbrondum: I think I understand the purpose of the benchmark, but I think it could be clearer/shorter written, if not written by AI but by human. Also, some parts I do not really understand, like "top-level token count measures one mint, allow-list size measures one insertion, and top-level lock count measures one creation"

plt/plt-scheduler/benches/block_state_size.rs
```diff
@@ -0,0 +1,423 @@
...
+        fixture.sender,
+        fixture.payload,
+    );
+    assert!(
+        matches!(result, Ok(ref result) if matches!(result.outcome, TransactionOutcome::Success(_)))
+    );
+    let _ = divan::black_box((result, fixture.state, fixture.context));
+}
+
+/// Measure one target-token mint as unrelated top-level token count grows.
```
allanbrondum: I have no idea what a sentence like this means :)

allanbrondum: Does it mean "Measure creating a token on a block state with `STATE_SIZES` tokens"?

allanbrondum (commented):
I think the "Benchmark comparison" section is hard to read. When I read the code, I think "Top-level lock create, 10" create a lock on a block state with 10 locks. And "Scheduler lock create, 10" means create 10 locks on an empty block state. But this is not clear at all in the section "Benchmark comparison". Generally, I think AI generated documentation is hard (if possible at all) to understand without context, it takes a lot of time to read it.

allanbrondum (approved):
I think we should make trie key conversion to and from bytes non-allocating, and "inline" the full, fixed key length, else I don't see any blockers.

plt/plt-scheduler/benches/block_state_size.rs
soerenbf: I'll revise and try to pay more attention to these in future, thanks 🙂

plt/plt-block-state/src/persistent/smart_contract_trie.rs
soerenbf: yeah it should be dead, it was added as part of #1686. Will remove

plt/plt-scheduler/benches/block_state_size.rs
soerenbf: mint, not create. but otherwise yes.. changed it

